PR #6, opened 2024-02-06
[OSDEV-494] List upload: List uploads in the "PENDING" status are being displayed on the Contributor Profile page

roman-stolar:
[OSDEV-494](https://opensupplyhub.atlassian.net/browse/OSDEV-494) List upload: List uploads in the "PENDING" status are being displayed on the Contributor Profile page
- Filter facility list data that sent to User Profile by status equal APPROVED


src/django/api/serializers/user/user_profile_serializer.py
```diff
@@ -72,6 +72,7 @@ def get_facility_lists(self, user):
                     source__contributor=contributor,
                     source__is_active=True,
                     source__is_public=True,
+                    status=FacilityList.APPROVED,
```
killev, line 75: @roman-stolar , Frankly saying I doubt that only approved need to be shown. It's not clear from the task. Let's clarify with @mariel-oar

killev (changes requested):
In general it looks good. but I'm not sure this is what really expected to be implemented in the original issue.

killev (approved):
LGTM.

VadimKovalenkoSNF (approved):
Approved

PR #14, opened 2024-02-08
[OSDEV 829] make possibility to push code with less than  1% diff in coverage from main

mazursasha1990:
Increase `minimum-ratio:` to 1

This solution isn't the best as it allows to decrease coverate for 1% what's too much, but this is munimal value


killev (approved):
LGTM, Approved

vladsha-dev (approved):
LGTM

killev (approved):
Approved

PR #26, opened 2024-02-15
[OSDEV-840] - Move handleMergeButtonClick to hooks, handle error clearance for merged facilities

VadimKovalenkoSNF:
Fix issue described in https://opensupplyhub.atlassian.net/browse/OSDEV-840
Fix UX issues after merging facilities in https://opensupplyhub.atlassian.net/browse/OSDEV-219

Common list of improvements:
1. Remove duplicated functionality of logic behind merging facilities, and place it to hooks.
2. Fixed redundant chain of Reject facility calls.
3. Refresh the state of checkboxes after successful merging.
4. Clear out the merge error message when merge modal window is closing.


src/app/src/components/FacilityListItemsConfirmationTableRow.jsx
```diff
@@ -86,24 +87,34 @@ function FacilityListItemsConfirmationTableRow({
     ];
 
     const [matches, setMatches] = useState([]);
+
+    const filterUniqueMatches = matchList => {
```
vladsha-dev, line 91: What do you think about using the useMemo hook to cache the result of a calculation between re-renders?
And move it to hooks.

vladsha-dev (changes requested):
Left one comment.

roman-stolar (approved):
LGTM

src/app/src/components/FacilityListItemsConfirmationTableRow.jsx
VadimKovalenkoSNF: I've checked your suggestion and it didn't work as expected for me, probably because of the usage of the local `matches` state here that should be updated each time the component receives new items. Lets use `useMemo` optimization a bit later after resolving other major tech debt issues.

vladsha-dev: Ok

vladsha-dev (approved):
I have locally verified that the merging functionality works as expected on both the search and list pages.

PR #32, opened 2024-02-20
[OSDEV-752] API. Include token and call info on API settings tab

Innavin369:
[OSDEV-752](https://opensupplyhub.atlassian.net/browse/OSDEV-752)

  User can see in “API” tab:
  
  their API call allowance (API limit)
  
  current call count
  
  their renewal period (monthly or annual)


roman-stolar (approved):
LGTM

vladsha-dev (approved):
LGTM

PR #37, opened 2024-02-22
Add 'Staging' and 'Production' to selection

killev:
Add possibility to select 'Staging' and 'Production' when manually start deploy

### DIFF
--- a/.github/workflows/deploy_to_aws.yml
+++ b/.github/workflows/deploy_to_aws.yml
@@ -19,6 +19,8 @@ on:
         - Development
         - Test
         - Pre-prod
+        - Sandbox
+        - Production
         default: Development
 
 env:
--- a/.github/workflows/deploy_to_aws.yml
+++ b/.github/workflows/deploy_to_aws.yml
@@ -19,7 +19,7 @@ on:
         - Development
         - Test
         - Pre-prod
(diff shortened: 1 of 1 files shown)


vladsha-dev (changes requested):
Looks like such statements should also be updated.
Screenshot 2024-02-22 at 15 59 47

roman-stolar:
> Looks like such statements should also be updated. Screenshot 2024-02-22 at 15 59 47

@vladsha-dev In case that we choose environment from selection (manually) the variable will be taken from **inputs.deploy-env** instead of branches that state could pass through **Development** & **Pre-Prod**. So it should work properly. In future if we want to trigger deployment of staging & production by tags it should be updated.

vladsha-dev:
> > Looks like such statements should also be updated. Screenshot 2024-02-22 at 15 59 47
> 
> @vladsha-dev In case that we choose environment from selection (manually) the variable will be taken from **inputs.deploy-env** instead of branches that state could pass through **Development** & **Pre-Prod**. So it should work properly. In future if we want to trigger deployment of staging & production by tags it should be updated.

Didn't grasp the logic the first time, thanks!

vladsha-dev (approved):
LGTM

PR #38, opened 2024-02-22
Disable the switch in routing to the ContriCleaner

vladsha-dev:
Don use contri-cleaner service for uploading facilities


roman-stolar (approved):
LGTM

killev (approved):

PR #53, opened 2024-02-29
[OSDEV-894] Use ContriCleaner parser in create facility api endpoint

mazursasha1990:
[OSDEV-894](https://opensupplyhub.atlassian.net/browse/OSDEV-894): Use `contricleaner` parser into create facility API endpoint (`facilities_view_set.py`)

All the Contricleaner functionality is ready for implementation into the create facility API endpoint. After these changes, data uploaded by a user in JSON format will be parsed by Contricleaner and returned all necessary fields to create a new facility. In case the data provided by the user are not valid then return a list with error messages.


src/django/contricleaner/lib/serializers/row_serializers/row_clean_field_serializer.py
```diff
@@ -8,7 +8,12 @@ def __init__(self, field: str, new_field: str) -> None:
         self.new_field = new_field
 
     def validate(self, row: dict, current: dict) -> dict:
-        clean_value = clean(row[self.field])
+        field = row.get(self.field)
+
+        if not field:
```
killev, line 13: This change seems to be wrong. If there's no value for clean data, it should return error.

src/django/contricleaner/lib/serializers/row_serializers/row_composite_serializer.py
```diff
@@ -18,11 +18,11 @@
 class RowCompositeSerializer:
     def __init__(self, sector_cache: SectorCacheInterface):
         self.validators = [
+            RowRequiredFieldsSerializer(),
```
killev, line 21: I think this change is excessive.

src/django/contricleaner/lib/serializers/row_serializers/row_country_serializer.py
```diff
@@ -4,8 +4,10 @@
 
 class RowCountrySerializer(RowSerializer):
     def validate(self, row: dict, current: dict) -> dict:
+        country = row.get("country", '')
```
killev, line 7: it's better to move this line under the `try:` block

src/django/contricleaner/lib/serializers/row_serializers/row_clean_field_serializer.py
```diff
@@ -8,7 +8,9 @@ def __init__(self, field: str, new_field: str) -> None:
         self.new_field = new_field
 
     def validate(self, row: dict, current: dict) -> dict:
-        clean_value = clean(row[self.field])
+        field = row.get(self.field, '')
```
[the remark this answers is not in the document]
mazursasha1990: It can't be reverted completely, because `clean` is supposed to receive a value. In the case when a user doesn't add any name or address, the `clean` function will return an error that does not follow the expectations. 
I propose this variant of the solution `clean_value = clean(row.get(self.field, ''))`

src/django/contricleaner/lib/serializers/row_serializers/row_composite_serializer.py
mazursasha1990: Done

src/django/contricleaner/lib/serializers/row_serializers/row_country_serializer.py
mazursasha1990: Done

src/django/contricleaner/tests/test_source_parser_json.py
killev, line 1: I wonder why you need to add this tests as part of this pr?

src/django/countries/tests/test_country_code_parser.py
killev, line 1: Please move this as separate PR

killev (changes requested):
Please split changes into 2 PRs:

1. Changes in countries
2. Changes in ContriCleaner
3. Changes in api/facilities

src/django/countries/tests/test_country_code_parser.py
mazursasha1990: Done

killev (approved):
WOW. Seems this can be merged!

roman-stolar (approved):
LGTM

PR #54, opened 2024-03-01
[OSDEV-902] Implement required fields serializer

roman-stolar:
[OSDEV-902](https://opensupplyhub.atlassian.net/browse/OSDEV-902) Implement required fields serializer
- Implemented header validation structure + required fields validation


killev (changes requested):
As basic architecture, I don't suppose to have header parser at all. Instead, I supposed to create RowRequiredFieldsSerializer with implementation. So that it validates incoming row's keys agains list of retired keys in it

src/django/contricleaner/lib/serializers/row_serializers/row_composite_serializer.py
```diff
@@ -17,6 +20,7 @@ def __init__(self):
             RowSectorSerializer(),
             RowCountrySerializer(),
             RowEmptySerializer(),
+            RowRequiredFieldsSerializer(),
```
killev, line 23: It should go after RowEmptySerializer which is simply add all feeds to the result

vladsha-dev (changes requested):
Left one comment

src/django/contricleaner/lib/serializers/row_serializers/row_required_fields_serializer.py
```diff
@@ -0,0 +1,22 @@
...
+
+class RowRequiredFieldsSerializer(RowSerializer):
+
+    def validate(self, row: dict, current: dict) -> dict:
+        required_fields = ["name",
+                           "address",
+                           "country"]
+
+        for required_field in required_fields:
+            if required_field in row.keys():
```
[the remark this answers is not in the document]
roman-stolar: Resolved

vladsha-dev (approved):
LGTM

src/django/contricleaner/lib/serializers/row_serializers/row_required_fields_serializer.py
```diff
@@ -0,0 +1,20 @@
+from .row_serializer import RowSerializer
+
+
+class RowRequiredFieldsSerializer(RowSerializer):
```
killev, line 4: I think. there's a better implementation could be copied from prev closed PR:

```
class HeaderRequredValidator:
     required_fields = {"name", "address", ""}

     def validate(self, raw_header: List[str]) -> dict:
         diff = self.required_fields.difference(raw_header)

         if len(diff) > 0:
             return {
                 "error": {
                     "message": "{} are missing".format(', '.join(diff)),
                     "type": "Error",
                 },
             }
         return {}
```

killev (changes requested):
Better implementation can be applied, please see comment inline

src/django/contricleaner/lib/serializers/row_serializers/row_required_fields_serializer.py
roman-stolar: Will do

mazursasha1990 (approved):
LGTM

PR #89, opened 2024-03-13
[OSDEV-893] Implement ContriCleaner parser in facility lists processing.

vladsha-dev:
- Implement the ContriCleaner parser for parsing facility lists immediately after list upload.
- Introduce a temporary API endpoint switcher for list uploads to enable switching to the old list upload API endpoint at runtime if the new endpoint impacts production uptime.

**Note:** I haven't added tests for the new list upload API endpoint since tests already exist for the old list upload API endpoint. Once the new endpoint fully replaces the old one, the existing tests should cover the new logic.


VadimKovalenkoSNF (changes requested):
Add comment related to list extension parsing.

mazursasha1990 (changes requested):

src/django/api/views/facility/facility_list_view_set.py
```diff
@@ -225,6 +282,215 @@ def create(self, request):
...
+            contributor=contributor,
+            source_type=Source.LIST,
+            facility_list=new_list)
+
+        is_geocoded = False
+        parsed_items = set()
+        for idx, row in enumerate(rows):
+            item = FacilityListItem.objects.create(
+                    row_index=idx,
+                    raw_data=','.join(row.raw_json.values()),
```
mazursasha1990, line 417: I suppose that we need to consolidate the data that we pass to `raw_data` because in `facility_view_set.py` this field looks like `raw_data=json.dumps(request.data)`

vladsha-dev: I think everything here is correct since `raw_data` is the stringified data of the row, not including the header. I believe the mistake is in the `facility_view_set.py` since the whole object is assigned to the `raw_data` of the facility list item, which seems incorrect because `raw_data` only stands for "_The full, unparsed CSV line as it appeared in the file._" whereas `raw_header` stands for "_The full, unparsed CSV header as it appeared in the file._"

src/django/api/views/facility/facility_list_view_set.py
```diff
@@ -102,6 +121,44 @@ def _extract_header_rows(self, file, request):
         rows = map(clean_row, rows)
         return header, rows
 
+    @staticmethod
+    def __get_serializer(
+            file: Union[InMemoryUploadedFile, TemporaryUploadedFile]
+            ) -> ContriCleanerSerializer:
+        sector_split_pattern = r', |,|\|'
+        ext = file.name[-4:]
+        if ext == 'xlsx':
```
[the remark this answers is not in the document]
vladsha-dev: Thanks for the comment.

1. I have used the `os` Python module, which is handier than the previous solution.
2. Regarding .xls, I agree with you. However, to preserve backward compatibility of the functionality, I wouldn't include such changes in this PR. I think it can be done as a separate task.

mazursasha1990 (approved):
LGTM

VadimKovalenkoSNF (approved):
Approved

PR #103, opened 2024-03-18
Consolidate fields raw_data and raw_header during the creation of FacilityListItem in the facility api endpoint according to the facility list endpoint

mazursasha1990:
Consolidate fields `raw_data` and `raw_header` during the creation of FacilityListItem in the facility api endpoint according to the facility list endpoint.


roman-stolar (approved):
LGTM

PR #116, opened 2024-03-20
[OSDEV-888] Fix an error that occurs when trying to open a facility from the Status Reports page

mazursasha1990:
[OSDEV-888](https://opensupplyhub.atlassian.net/browse/OSDEV-888): Fix an error that occurs when trying to open a facility from the Status Reports page

The error occurred due to activity reports with the status `pending` containing fields with `null` values and these values pass to the `format_date` function as an argument.
Modified the `get_activity_reports` method in the `FacilityIndexDetailsSerializer` to prevent passing a falsy `date` argument into the `format_date` function.

### DIFF
--- a/src/django/api/serializers/facility/facility_index_details_serializer.py
+++ b/src/django/api/serializers/facility/facility_index_details_serializer.py
@@ -190,26 +190,23 @@ class FacilityIndexDetailsSerializer(FacilityIndexSerializer):
     def get_activity_reports(self, facility):
         return [
             {
-                'facility': item['facility'] or None,
-                'reported_by_contributor': (
-                    item['reported_by_contributor'] or None
+                'facility': item.get('facility'),
+                'reported_by_contributor': item.get('reported_by_contributor'),
+                'closure_state': item.get('closure_state'),
+                'approved_at': format_date(item.get('approved_at')),
+                'status_change_reason': item.get('status_change_reason'),
+                'status': item.get('status'),
+                'status_change_by': item.get('status_change_by'),
+                'status_change_date': format_date(
+                    item.get('status_change_date')
                 ),
-                'closure_state': item['closure_state'] or None,
-                'approved_at': format_date(item['approved_at']) or None,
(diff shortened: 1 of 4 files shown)


src/django/api/serializers/facility/utils.py
```diff
@@ -410,6 +410,9 @@ def regroup_claims_for_sector_field(claims: list,
 
 
 def format_date(date: str) -> str:
+    if not date:
```
killev, line 413: I doubt that None passed here is a normal situation. I think we need to dig deeper why it happened here. I expect a problem with logic in some other place

src/django/api/serializers/facility/facility_index_details_serializer.py
```diff
@@ -190,26 +190,29 @@ def get_claim_info(self, facility):
     def get_activity_reports(self, facility):
         return [
             {
-                'facility': item['facility'] or None,
-                'reported_by_contributor': (
-                    item['reported_by_contributor'] or None
+                'facility': item.get('facility'),
```
killev, line 193: What's the difference between item['facility'] and item.get('facility')?

src/django/api/tests/test_facility_index_details_serializer.py
```diff
@@ -348,3 +353,30 @@ def test_get_other_names_embed_mode_active(self):
         expected_other_names = []
         actual_other_names = serializer.get_other_names(facility_index)
         self.assertEqual(expected_other_names, actual_other_names)
+
+    def test_get_activity_reports(self):
+        FacilityActivityReport.objects.create(
```
roman-stolar, line 358: Did test covered a scenario that was reproduced on this bug?

mazursasha1990: Yes, it does.

killev (dismissed):
LGTM, Approved

roman-stolar (approved):
LGTM

killev (approved):
LGTM, Approved

PR #119, opened 2024-03-22
Backup Plan - [OSDEV-966] Facility list upload. The constraints for each DB field aren't validated before saving the data.

vladsha-dev:
_Note:_
This fix is a backup plan. The final solution will be implemented within the ContriCleaner lib and in another pull request.

### DIFF
--- a/src/django/api/views/facility/facility_list_view_set.py
+++ b/src/django/api/views/facility/facility_list_view_set.py
@@ -29,6 +29,7 @@ from django.db import transaction
 from django.db.models import F, Q
 from django.utils import timezone
 from django.contrib.gis.geos import Point
+from django.core.exceptions import ValidationError as DjangoCoreValidationError
 
 from ...aws_batch import submit_jobs, submit_parse_job
 from api.constants import (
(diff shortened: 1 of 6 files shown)


src/django/api/views/facility/facility_list_view_set.py
```diff
@@ -441,6 +448,48 @@ def createlist(self, request):
                     list(row.fields.keys()),
                     list(row.fields.values())
                 )
+
+                try:
+                    # TODO: Make this validation inside the ContriCleaner lib.
+                    item.full_clean(exclude=(
```
killev, line 454: @vladsha-dev why not to make this as part of this PR?

killev (changes requested):
I think validation needs to be moved as part of contri cleaner as part of this PR

src/django/api/views/facility/facility_list_view_set.py
vladsha-dev: Will do as part of this PR.

vladsha-dev: I don't think we need to add it to the ContriCleaner library for now, as this check wasn't performed for the API requests previously. However, I believe we need to address it for list uploads, as it was already functioning for them. Afterward, I will create a follow-up Jira ticket to consolidate constraints and validate latitude and longitude. The intent of the OSDEV-966 Jira task was to ensure backward compatibility, not to introduce new functionality.

vladsha-dev: Hi @killev,
After discussing with Mariel and Vanessa, we agreed to preserve the validation as it currently functions on production for list uploads. Since the goal of the OSDEV-966 ticket is to revert to the current production state, particularly for list uploads, I propose creating a separate ticket for your idea. While your suggestion may streamline API uploads and list uploads validation, it would necessitate rewriting and reestimating the OSDEV-966 ticket.
Does it sound good to you?

vladsha-dev: Will do.

vladsha-dev:
Closed in favor of https://github.com/opensupplyhub/open-supply-hub/pull/126

PR #130, opened 2024-03-28
Hotfix [OSDEV-979] Upload: List are receiving an error at the parsing phase during upload

vladsha-dev:
Fix [[OSDEV-979](https://opensupplyhub.atlassian.net/browse/OSDEV-979)] Upload: List are receiving an error at the parsing phase during upload

### DIFF
--- a/src/django/api/views/facility/facility_list_view_set.py
+++ b/src/django/api/views/facility/facility_list_view_set.py
@@ -29,6 +29,7 @@ from django.db import transaction
 from django.db.models import F, Q
 from django.utils import timezone
 from django.contrib.gis.geos import Point
+from django.core.exceptions import ValidationError as DjangoCoreValidationError
 
 from ...aws_batch import submit_jobs, submit_parse_job
 from api.constants import (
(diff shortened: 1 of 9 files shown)


src/django/api/views/facility/facility_list_view_set.py
```diff
@@ -62,7 +63,7 @@
     FacilityListItemsQueryParamsSerializer,
 )
 from api.extended_fields import (
-    create_extendedfields_for_listitem
+    create_extendedfields_for_single_item
```
roman-stolar, line 66: **Non blocking question:**
Why you replace the method to single item as it list upload?

src/django/api/views/facility/facility_list_view_set.py
```diff
@@ -414,62 +431,107 @@ def createlist(self, request):
...
+            # Created a partially filled FacilityListItem for the
+            # create_extendedfields_for_listitem method, as this method
+            # necessitates a row already saved in the database.
             item = FacilityListItem.objects.create(
                     row_index=idx,
                     raw_data=','.join(row.raw_json.values()),
                     raw_json=row.raw_json,
                     raw_header=header_str,
-                    sector=row.sector,
+                    sector=[],
```
roman-stolar, line 442: **Non blocking question:**
Why you change it to empty array when FacilityListItem creating?

roman-stolar (approved):
In general LGTM, just left question for understanding.

killev (approved):
Approve this changes only because it's temporary solution and will be tested thoroughly

src/django/api/views/facility/facility_list_view_set.py
vladsha-dev: It can't be empty, but I need to assign some value to it since there are no default values for it.

src/django/api/views/facility/facility_list_view_set.py
vladsha-dev: It won't work with `create_extendedfields_for_listitem` since the data from the ContriCleaner has already been adapted for `create_extendedfields_for_single_item`, and another advantage is that the ContriCleaner did the same thing as `create_extendedfields_for_listitem` in terms of assigning `processing_type` and `facility_type`.

PR #132, opened 2024-03-28
Hotfix [OSDEV-979] Upload: List are receiving an error at the parsing phase during upload

vladsha-dev:
Hotfix [[OSDEV-979](https://opensupplyhub.atlassian.net/browse/OSDEV-979)] Upload: List are receiving an error at the parsing phase during upload

- Return to the old method of header validation since the current implementation cannot properly function due to the focus of ContriCleaner on API uploads. Each row returned from the ContriCleaner library doesn't guarantee that it consists of all the header fields. For example, the first row may have only two required fields, but the next row may have all the required fields. It depends on whether they are filled in the table or not. Therefore, we cannot determine whether the header contains all the necessary required fields based solely on the first row.
- Return to the old method of creating nonstandard fields because each `raw_json` object can contain a different number of fields, as not all fields can be filled for each row in the table. The previous solution did not result in the creation of all uploaded extended fields via the table.

### DIFF
--- a/src/django/api/views/facility/facility_list_view_set.py
+++ b/src/django/api/views/facility/facility_list_view_set.py
@@ -2,9 +2,8 @@ import csv
 import operator
 import os
 import traceback
-import re
 from functools import reduce
-from typing import Union, List
+from typing import Union
 
 from api.helpers.helpers import (
     get_raw_json,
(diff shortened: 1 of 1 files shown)


roman-stolar (approved):

killev (approved):
Approve as temporary solution

PR #136, opened 2024-04-02
[OSDEV-990] Create a ContriCleaner facade class for interacting with the client code.

vladsha-dev:
[[OSDEV-990](https://opensupplyhub.atlassian.net/browse/OSDEV-990)] Create a ContriCleaner facade class for interacting with the client code.

Implement a ContriCleaner facade class to simplify interaction with client code. With this change, the client code only needs to instantiate the ContriCleaner class, pass the input data, and then call the `process_data` method without the need to define strategies or other details. This abstraction helps streamline the process and encapsulate complexity.


killev (changes requested):
Vlad, could you please split this PR into several smaller so that I see which refactoring you do on every step?

vladsha-dev:
> Vlad, could you please split this PR into several smaller so that I see which refactoring you do on every step?

No, I can't. I have already done it as small as possible.

roman-stolar (approved):
In general LGTM.

src/django/contricleaner/tests/sector_cache_mock.py
```diff
@@ -1,4 +1,6 @@
-from contricleaner.lib.sector_cache_interface import SectorCacheInterface
+from contricleaner.lib.client_abstractions.sector_cache_interface import (
+    SectorCacheInterface
+)
 
 
 class MockSectorCache(SectorCacheInterface):
```
mazursasha1990, line 6: I suppose it's suitable to rename the class for more explicit after renaming the file name.

src/django/api/views/facility/facilities_view_set.py
```diff
@@ -583,11 +581,9 @@ def create(self, request):
 
         log.info(f'[API Upload] Uploading data: {request.data}')
         log.info('[API Upload] Started CC Parse process!')
-        split_pattern = r', |,|\|'
```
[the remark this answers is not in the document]
vladsha-dev: I don't believe the client code (Django) should handle the `split_pattern`, as it primarily pertains to ContriCleaner's functionality. I don't see it as a value that changes dynamically based on user input or other factors. Therefore, let's move it to the ContriCleaner.

src/django/contricleaner/tests/sector_cache_mock.py
vladsha-dev: Will do.
Thanks👍

vladsha-dev: Done

mazursasha1990 (approved):
LGTM

PR #139, opened 2024-04-03
[OSDEV-984] Facility list upload. Header validation is failing, even though all the required columns and data are filled.

roman-stolar:
[OSDEV-984](https://opensupplyhub.atlassian.net/browse/OSDEV-984) Facility list upload. Header validation is failing, even though all the required columns and data are filled.
- created basic structure for early header validation
- added composite chain for early validation


vladsha-dev (changes requested):
Please update the release notes file.

src/django/contricleaner/lib/validators/early_validators/early_header_validator.py
killev, line 1: Lets name it pre_ (not early_)

src/django/contricleaner/lib/handlers/handler.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+from typing import List
+from contricleaner.lib.dto.list_dto import ListDTO
+
+
+class Handler(ABC):
```
killev, line 6: It's very common name for class. Could you please use more specific (say RowHandler, or similar)
?

roman-stolar: CompositeChainHandler

vladsha-dev: In this case, I think we can name it `ListRowHandler`

src/django/contricleaner/lib/validators/early_validators/early_header_validator.py
vladsha-dev: +

src/django/contricleaner/lib/handlers/handler.py
roman-stolar: Yep, sounds good

src/django/contricleaner/lib/dto/list_dto.py
```diff
@@ -0,0 +1,7 @@
+from typing import List, NamedTuple
+from contricleaner.lib.dto.row_dto import RowDTO
+
+
+class ListDTO(NamedTuple):
+    rows: List[RowDTO]
```
vladsha-dev, line 6: Let's set the default value to an empty array `[]`. It will be useful in cases where you need to check whether it contains any values or not.

src/django/contricleaner/lib/handlers/handler.py
vladsha-dev: It can't be CompositeHandler because it doesn't relate to the Composite pattern.

src/django/contricleaner/lib/handlers/handler.py
```diff
@@ -0,0 +1,14 @@
...
+
+    @abstractmethod
+    def setNext(self, next):
```
vladsha-dev: 1. Please use Pythonic style for function naming: `set_next`.
2. Please add a type hint for the `next` parameter.
3. Please implement this function directly in this class. Though this class will be more like an abstract class than an interface, it is acceptable since child classes will not need to implement the same code repeatedly.

src/django/contricleaner/lib/handlers/early_validation_handler.py
```diff
@@ -0,0 +1,28 @@
...
+from contricleaner.lib.validators.early_validators.\
+    early_header_validator import EarlyHeaderValidator
+from contricleaner.lib.validators.early_validators.\
+    composite_early_validator import CompositeEarlyValidator
+
+
+class EarlyValidationHandler(Handler):
+    __next: Handler
+
+    def setNext(self, next: Handler):
```
vladsha-dev: This function can be deleted from this class since it must be implemented within the parent class.

vladsha-dev (changes requested):
Left comments

src/django/contricleaner/lib/dto/list_dto.py
vladsha-dev: the same for errors

killev (approved):
LGTM, Approved

vladsha-dev (approved):
LGTM

PR #144, opened 2024-04-05
[OSDEV-1000] Move facility processing received from api to a separate class

mazursasha1990:
[OSDEV-1000](https://opensupplyhub.atlassian.net/browse/OSDEV-1000) -  Move facility processing received from API to a separate class `ProcessingFacility`.

The functionality of processing a new facility, which was previously in `facilities_view_set.py`, has been moved to `processing_facility.py.` 
This marks the initial step in implementing a single class responsible for managing the processing of new facilities from both API requests and list uploads.

### DIFF
--- a/src/django/api/facility_actions/create_facility.py
+++ b/src/django/api/facility_actions/create_facility.py
@@ -1,0 +1,228 @@
+import asyncio
+import json
+import logging
+import traceback
+
+from api.constants import ErrorMessages, ProcessingAction
+from api.geocoding import geocode_address
+from api.kafka_producer import produce_message_match_process
+from api.models.facility.facility_list_item import FacilityListItem
+from api.models.facility.facility_list_item_temp import FacilityListItemTemp
+from api.processing import handle_external_match_process_result
+from api.views.fields.create_nonstandard_fields import (
+    create_nonstandard_fields,
+)
+from rest_framework import status
+from rest_framework.response import Response
+
+from django.contrib.gis.geos import Point
+from django.core import exceptions as core_exceptions
+from django.utils import timezone
+
+# initialize logger
+logging.basicConfig(
+    format='%(asctime)s - %(levelname)s - %(message)s', level=logging.INFO
+)
+log = logging.getLogger(__name__)
+
+
+class CreateFacility:
+
+    def createApi(request, row, source, should_create):
+
+        if row.errors:
+            log.error(f'[API Upload] CC Parsing Errors: {row.errors}')
+            return Response(
+                {
+                    "message": "The provided data could not be parsed",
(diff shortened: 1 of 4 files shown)


vladsha-dev (changes requested):
It seems like there's no consolidation between the lists and the API when it comes to saving facilities. It's simply a matter of moving the facility saving process from the API to a separate function.

mazursasha1990:
> It seems like there's no consolidation between the lists and the API when it comes to saving facilities. It's simply a matter of moving the facility saving process from the API to a separate function.

You are right. I've decided on the following approach for consolidation:

- Move facility creation received from the API to a separate class called `CreateFacility`.
- Move facility creation received from list uploads to another separate class, also named `CreateFacility`.
- Consolidate these two functions as much as possible.

I plan to implement all of these steps in different PRs.

killev (approved):
LGTM, Approved

src/django/api/facility_actions/create_facility.py
```diff
@@ -0,0 +1,228 @@
...
+        result = {
+            'matches': [],
+            'item_id': item.id,
+            'geocoded_geometry': None,
+            'geocoded_address': None,
+            'status': item.status,
+        }
+
+        try:
+            (item, row.fields)
```
vladsha-dev, line 87: The function call is missing here.

vladsha-dev (changes requested):
1. Left one comment.
2. Please fill in the Release notes file.

vladsha-dev (approved):
LGTM

src/django/api/views/facility/facilities_view_set.py
```diff
@@ -604,194 +583,19 @@ def create(self, request):
...
+        processed_data = contri_cleaner.process_data()
+        rows = processed_data.rows
+        row = rows[0]
 
-        if (should_create
-                and result['status'] not in errors_status):
-            return Response(result, status=status.HTTP_201_CREATED)
-        else:
-            return Response(result, status=status.HTTP_200_OK)
+        return CreateFacility.createApi(request, row, source, should_create)
```
roman-stolar, line 598: @killev @mazursasha1990 I'm not sure that **CreateFacility** is a good name for it. As there is not always new facility creation happening, there could be "matched", "pending (potential match)". I suggest name it "ProcessingFacility" or something similar.

vladsha-dev: Good idea

mazursasha1990: You are right. 
Thanks!

roman-stolar (approved):

PR #154, opened 2024-04-09
[OSDEV-974] Reporting. Contributor type by %.

Innavin369:
[OSDEV-974](https://opensupplyhub.atlassian.net/browse/OSDEV-974)  Reporting. Contributor type by %.

* Admin see in the report data for the percent of data contributors on the platform by type (this should be in percent format with two decimal places shown), only accounts that have contributed data, the data should be ordered by most recent to oldest month and display mid-month values.


roman-stolar (approved):
LGTM

doc/release/RELEASE-NOTES.md
```diff
@@ -30,6 +30,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 * [OSDEV-660](https://opensupplyhub.atlassian.net/browse/OSDEV-660) - Remove punctuation issues with duplicated commas and double quotes while facility list uploading.
 
 ### What's new
+* [OSDEV-974](https://opensupplyhub.atlassian.net/browse/OSDEV-974) Reporting. Contributor type by %. Admin see in the report data for the percent of data contributors on the platform by type (this should be in percent format with two decimal places shown), only accounts that have contributed data, the data should be ordered by most recent to oldest month and display mid-month values.
```
VadimKovalenkoSNF, line 33: Fix typo, it should be `Admin sees`

src/django/api/reports/contributor_type_by_percent.sql
```diff
@@ -0,0 +1,48 @@
+SELECT
```
VadimKovalenkoSNF, line 1: I think this query can be optimized, but need other developer to take a look.

VadimKovalenkoSNF (changes requested):
Left few comments

src/django/api/reports/contributor_type_by_percent.sql
VadimKovalenkoSNF: Looks better now, thanks

roman-stolar (approved):
Approved

PR #157, opened 2024-04-10
[OSDEV-986] Fix the population of the custom data points uploaded via lists.

vladsha-dev:
[[OSDEV-986](https://opensupplyhub.atlassian.net/browse/OSDEV-986)] Fix the population of the custom data points uploaded via lists.

Correct the population of the custom data points that are uploaded via lists. Ensure that the full list header is saved in the database, and that the raw data for each facility list item is saved as a string of strings, with each value separated by a comma. This way, it helps maintain backward compatibility with the functionality responsible for displaying custom data points on the embedded maps. Also, revert to the previous default logic, which saves the sector as `Unspecified` when sector, sector_product_type, or product_type have empty values.

_Note:_ I successfully tested the list and API uploads locally with different lists and request bodies.


src/django/api/views/facility/facility_list_view_set.py
```diff
@@ -390,7 +390,8 @@ def createlist(self, request):
         for idx, row in enumerate(rows):
             item = FacilityListItem.objects.create(
                     row_index=idx,
-                    raw_data=','.join(row.raw_json.values()),
+                    raw_data=','.join(
+                        f'"{value}"' for value in row.raw_json.values()),
```
roman-stolar, line 394: could we just use **str(value)**?

vladsha-dev: I thought about it, but no. To maintain backward compatibility, it's necessary to enclose the values in single quotes first, and then in double quotes. This will generate the following string: '"Name value","Street, value", "Country value"'. While generating the contributor fields, the code csv.reader(csv_data.split('\n'), delimiter=',') splits the raw_data based on commas between double quotes. This helps avoid a situation where the address is split into separate values.

roman-stolar (approved):
LGTM

src/django/api/views/facility/facility_list_view_set.py
```diff
@@ -402,56 +403,54 @@ def createlist(self, request):
...
+                            and FileHeaderField.LNG in row.fields.keys()):
+                        # TODO: Move floating to the ContriCleaner library.
+                        lat = float(row.fields[FileHeaderField.LAT])
+                        lng = float(row.fields[FileHeaderField.LNG])
+                        item.geocoded_point = Point(lng, lat)
+                        is_geocoded = True
+
+                    create_extendedfields_for_single_item(item, row.fields)
+                except Exception as e:
+                    log.error(
```
VadimKovalenkoSNF, line 439: Add Rollbar logging here as well.

VadimKovalenkoSNF (approved):
Approved, but left optional comment.

src/django/api/views/facility/facility_list_view_set.py
vladsha-dev: Done

PR #182, opened 2024-04-22
[OSDEV-931] Facility Claims. Use account email for contact info on claim request.  

vladsha-dev:
[OSDEV-931](https://opensupplyhub.atlassian.net/browse/OSDEV-931)

The following features have been implemented:
- Made the Email field in the claim form uneditable, setting the claimer's email as the default value for this field.
- Removed the _Preferred method of contact_ field from both the claim form and the claim details page in the admin dashboard.
- Since `email` and `preferred_contact_method` are no longer necessary for the claim form, they have been removed from the `FacilityClaim` model and the respective history table. Additionally, the old `index_approved_claim` function has been replaced with a similar one that does not index the `preferred_contact_method` field.


src/app/src/__tests__/utils.tests.js
```diff
@@ -1445,15 +1445,10 @@ it('creates a URL for POSTing the claim a facility form', () => {
 
 it('checks whether the claim a facility form is valid', () => {
     const validForm = {
-        email: 'email@example.com',
```
VadimKovalenkoSNF, line 1448: Why did you remove 'email' field from test as it's still populated in the request body?

VadimKovalenkoSNF:
Question: There are multiple files in `src/django/sqls` that contain `preferred_contact_method`. Should they be updated as well?

vladsha-dev:
> Question: There are multiple files in `src/django/sqls` that contain `preferred_contact_method`. Should they be updated as well?

Hi @VadimKovalenkoSNF,
Thank you for the question. To maintain the ability to set up the database from its initial state to the latest migration, we store all SQL files for schema updates in the `src/django/sqls` folder. If we need to modify any SQL introduced as SQL functions in the database, we create a new SQL file, such as `src/django/sqls/0143_index_approved_claim.sql `and include it in the DB migration. This ensures that the code from the latest SQL file is applied in the database via one of the latest DB migrations, even though all previous files still exist.

src/django/api/mail.py
```diff
@@ -48,20 +48,19 @@ def send_claim_facility_confirmation_email(request, facility_claim):
         'facility_country': facility_country,
         'facility_url': make_facility_url(request, facility_claim.facility),
         'contact_person': facility_claim.contact_person,
-        'email': facility_claim.email,
+        'email': facility_claim.contributor.admin.email,
```
VadimKovalenkoSNF, line 51: A bit unsafe construction. I would add optional chainning here because property has a few nesting levels.

VadimKovalenkoSNF (changes requested):
Add minor comment.

src/django/api/mail.py
vladsha-dev: I don't think we need to do that because of the database constraints. It's impossible to have a facility claim without a contributor, a contributor without an admin, or an admin without an email.

vladsha-dev: @VadimKovalenkoSNF

VadimKovalenkoSNF (approved):
Approved

roman-stolar (dismissed):
LGTM

roman-stolar (approved):

VadimKovalenkoSNF (approved):
Approved

PR #188, opened 2024-04-30
Dedupe Hub Fix: handle condition when no message received on kafka topic

roman-stolar:
Dedupe Hub Fix: handle condition when no message received on kafka topic
- Those changes with condition fix possible stuck when no message received or DB is empty.


vladsha-dev (approved):
LGTM

PR #192, opened 2024-05-06
[OSDEV-744] API. When user want to confirm/reject potential_match it didn't found a match through 'id'

roman-stolar:
[OSDEV-744](https://opensupplyhub.atlassian.net/browse/OSDEV-744) API. When user want to confirm/reject potential_match it didn't found a match through 'id'
- Got match id from live/origin table when create equal true for Potential Match confirm url


src/django/api/processing.py
```diff
@@ -639,13 +639,21 @@ def handle_external_match_process_result(id, result, request, should_create):
...
         # No Match and Geocoder Returned No Results
         return get_error_match_result(f_l_item.id, result)
-    if (queryset_f_m.count() == 1
-            and queryset_f_m[0].status == FacilityMatchTemp.AUTOMATIC):
+
+    automatic_matches = [
+        match for match in queryset_f_m
+        if match.status == match_object_type.AUTOMATIC
+    ]
+    if (len(automatic_matches) == 1):
```
VadimKovalenkoSNF, line 656: Question: If there is at least 1 automatic match here, shall we operate with new facility here at all? I mean, why do we need `# New Facility` block here?

VadimKovalenkoSNF (commented):
Left a question

src/django/api/processing.py
roman-stolar: @VadimKovalenkoSNF Yes, Automatic Match status apply for both situations **NEW_FACILITY** & **MATCHED**.

VadimKovalenkoSNF (commented):
Approved

VadimKovalenkoSNF (approved):

mazursasha1990 (approved):
Approve

PR #205, opened 2024-05-15
Delete the 'public' scheme before database restore

killev:
* if newer database has changes with new relations/constraints, restoring the database from dumps leads to conflicts. So removing schema allows not to have conflicting data and restoring database from empty state


VadimKovalenkoSNF (approved):
Approved

PR #207, opened 2024-05-15
Ignore errors on restore

killev:
* ignore results of pg_restore as it always fails even when db restored correctly so that pipelines proceeds with further steps


roman-stolar (approved):

VadimKovalenkoSNF (approved):
Approved.

PR #212, opened 2024-05-20
[OSDEV-788] API user. Incorrect response upon attempting to check for a match.

roman-stolar:
[OSDEV-788](https://opensupplyhub.atlassian.net/browse/OSDEV-788) API user. Incorrect response upon attempting to check for a match.
- Re-written logic for New_Facility/Automatic_Match when we collect & save data for FacilityListItemTemp/FacilityMatchTemp.


roman-stolar:
Working on possible tests

roman-stolar:
Team please skip Dedupe Hub App | Unittest test suite - Code coverage report as it didn't showed properly in case of missed init.py files all over the project.

src/dedupe-hub/api/tests/test_cumulative_matcher.py
```diff
@@ -0,0 +1,108 @@
...
+            {
+                "facility_id": "CN20241096SFEBA",
+                "confidence": 1.0,
+                "facility_list_item_id": "1",
+                "status": "AUTOMATIC",
+                "results": {}
+            },
+        ]
+        result_two = c_m.clean_matched(self.messy, processed_matches_two)
+        self.assertEqual(result_two, {
```
VadimKovalenkoSNF, line 102: Use `assertDictEqual` here instead.

VadimKovalenkoSNF (changes requested):
Left minor comment.

src/dedupe-hub/api/tests/test_cumulative_matcher.py
roman-stolar: Done

VadimKovalenkoSNF (approved):
Looks good overall, but need other developers to verify.

roman-stolar:
> But pls, increase codecov issue for DD.

As I mention it an global issue that init.py files missed, so the DD coverage result are wrong.

vladsha-dev:
> > But pls, increase codecov issue for DD.
> 
> As I mention it an global issue that init.py files missed, so the DD coverage result are wrong.

I suspect the code coverage has decreased because some files were not covered by tests previously, but now they are. It looks like these files weren't fully covered while writing the new tests, which is why it says that the code coverage decreased. So, it requires writing more tests for these specific files, or you can add `__init__.py` files in a separate PR but make this PR dependent on the first one. This way, we will not have this problem in the future with the DD hub, at least.

roman-stolar:
> I suspect the code coverage has decreased because some files were not covered by tests previously, but now they are. It looks like these files weren't fully covered while writing the new tests, which is why it says that the code coverage decreased. So, it requires writing more tests for these specific files, or you can add **init**.py files in a separate PR but make this PR dependent on the first one. This way, we will not have this problem in the future with the DD hub, at least.

Yes, I will create seperate PR with init.py files for DD and ask to merge them first.
cc @vladsha-dev @VadimKovalenkoSNF

roman-stolar:
@VadimKovalenkoSNF @vladsha-dev Please review & approve https://github.com/opensupplyhub/open-supply-hub/pull/216

VadimKovalenkoSNF (dismissed):
Approved

vladsha-dev (approved):
LGTM

VadimKovalenkoSNF (approved):
Lgtm

PR #243, opened 2024-06-13
Wait till AWS ECS task ends

killev:
An additional loop was added to the `run_cli_task` script that repeatedly checks the status of an AWS ECS task, waiting for it to stop.

The current solution should prevent the `Waiter TasksStopped failed: Max attempts exceeded` error that occurred during the execution of the `post_deployment` task.

### DIFF
--- a/deployment/run_cli_task
+++ b/deployment/run_cli_task
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-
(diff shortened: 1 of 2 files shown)


VadimKovalenkoSNF (approved):
Approved

roman-stolar (approved):

PR #258, opened 2024-06-25
[OSDEV-1098] Add migration for the report "Contributor type by %"

Innavin369:
[OSDEV-1098](https://opensupplyhub.atlassian.net/browse/OSDEV-1098) Reporting. A columns values in the report "Contributor type by %" are not cumulative.

* Added migration for dropping unnecessary calc_column function and creation add_percent_to_number function for  the "Contributor type by %" report.


src/django/sqls/drop_calc_column_func.sql
```diff
@@ -0,0 +1 @@
+DROP FUNCTION IF EXISTS calc_column(numeric, numeric);
```
vladsha-dev, line 1: There is no such function. Please use the actual types while deleting this function.

src/django/sqls/0150_revert_add_percent_to_number.sql
```diff
@@ -0,0 +1 @@
+DROP FUNCTION IF EXISTS add_percent_to_number(int8, int8);
```
vladsha-dev, line 1: If you want to delete this function, you should use the same parameter types that were set when declaring the function.

vladsha-dev (changes requested):

src/django/sqls/0150_revert_add_percent_to_number.sql
Innavin369: changed

vladsha-dev (dismissed):
LGTM

roman-stolar (approved):
LGTM

vladsha-dev (approved):
LGTM

roman-stolar (approved):
LGTM

PR #220, opened 2024-05-28
[Osdev 1004] Introduce ingress and egress rules for the Opensearch

VadimKovalenkoSNF:
Fix [OSDEV-1004](https://opensupplyhub.atlassian.net/browse/OSDEV-1004)

- Introduce ingress and egress rules for the Opensearch and Logstash
- Parameterized database credentials for the logstash configs input.
- Parameterized OpenSearch domain for the logstash configs output.
- Specifies the ARN of an IAM role to be used as the master user for the OpenSearch domain.
- Set EFS access point permissions for logstash:root user.

PR #267, opened 2024-06-28
[OSDEV-1005] Propagate facility deletions to OpenSearch data source

vladsha-dev:
[[OSDEV-1005](https://opensupplyhub.atlassian.net/browse/OSDEV-1005)] Propagate facility deletions to OpenSearch data source

- Implement the propagation of production location deletions from the PostgreSQL database to the OpenSearch cluster. After this fix, the locations that were deleted will be excluded from the response of the `v1/production-location` GET API endpoint.
- Enable deployment of the Logstash and OpenSearch infra to AWS environments.
- Disconnect location deletion propagation to the OpenSearch cluster while the Django tests are running, as it is outside the scope of Django unit testing.

[OSDEV-1005]: https://opensupplyhub.atlassian.net/browse/OSDEV-1005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployment/terraform/firewall.tf
```diff
@@ -425,3 +435,13 @@ resource "aws_security_group_rule" "opensearch_ingress" {
   security_group_id        = aws_security_group.opensearch.id
   source_security_group_id = aws_security_group.app_logstash.id
 }
+
+resource "aws_security_group_rule" "app_opensearch_ingress" {
```
VadimKovalenkoSNF, line 439: What problem this ingress rule solved? Afair, there were no problems with incoming traffic to the OpenSearch previously in https://github.com/opensupplyhub/open-supply-hub/pull/220.

src/django/api/signals.py
```diff
@@ -0,0 +1,44 @@
...
+from django.db.models.signals import post_delete
+from django.dispatch import receiver
+from opensearchpy.exceptions import ConnectionError
+
+from api.models.facility.facility import Facility
+from api.services.opensearch import OpenSearchServiceConnection
+from oar.rollbar import report_error_to_rollbar
+
+
+# Initialize logger.
```
VadimKovalenkoSNF, line 13: This comment can be removed

src/django/api/signals.py
```diff
@@ -0,0 +1,44 @@
...
+log = logging.getLogger(__name__)
+
+
+@receiver(post_delete, sender=Facility)
+def location_post_delete_handler_for_opensearch(sender, **kwargs):
+    opensearch = OpenSearchServiceConnection()
+    location_instance = kwargs.get('instance')
+    try:
+        response = opensearch.client.delete(
+            index='production-locations',
```
VadimKovalenkoSNF, line 23: In the future I'd like to have these index values listed in a separate file as constants.

VadimKovalenkoSNF (changes requested):
Overall looks good for me, but left few notes before approval.

src/django/api/signals.py
vladsha-dev: Done

deployment/terraform/firewall.tf
vladsha-dev: It is necessary to enable a network connection from Django to OpenSearch. Previously, ingress traffic from Django to OpenSearch was blocked.

VadimKovalenkoSNF (approved):
LGTM

PR #271, opened 2024-07-02
[OSDEV-1036] Claims. Add a sortable "claim decision" column to claims admin page. 

roman-stolar:
[OSDEV-1036](https://opensupplyhub.atlassian.net/browse/OSDEV-1036) **Claims. Add a sortable "claim decision" column to claims admin page.**
- added new data point 'claim_decision' to response & create new column on page
- update styling & location of the field
- update sort logic for null value

[OSDEV-1036]: https://opensupplyhub.atlassian.net/browse/OSDEV-1036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/react/src/components/DashboardClaimsListTable.jsx
+++ b/src/react/src/components/DashboardClaimsListTable.jsx
@@ -165,6 +165,9 @@ function DashboardClaimsListTable({
                                 <TableCell padding="dense">
                                     {claim.status}
                                 </TableCell>
+                                <TableCell padding="dense">
+                                    {claim.last_decision}
+                                </TableCell>
                             </TableRow>
                         ))}
                     </TableBody>
--- a/src/react/src/components/DashboardClaimsListTable.jsx
+++ b/src/react/src/components/DashboardClaimsListTable.jsx
@@ -160,13 +160,17 @@ function DashboardClaimsListTable({
                                     {moment(claim.created_at).format('LL')}
                                 </TableCell>
                                 <TableCell padding="dense">
-                                    {moment(claim.updated_at).format('LL')}
+                                    {claim.claim_decision !== null
+                                        ? moment(claim.claim_decision).format(
+                                              'LL',
+                                          )
+                                        : 'N/A'}
                                 </TableCell>
                                 <TableCell padding="dense">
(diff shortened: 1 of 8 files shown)


src/react/src/components/DashboardClaimsListTable.jsx
```diff
@@ -160,11 +160,18 @@ function DashboardClaimsListTable({
                                     {moment(claim.created_at).format('LL')}
                                 </TableCell>
                                 <TableCell padding="dense">
-                                    {moment(claim.updated_at).format('LL')}
+                                    {claim.claim_decision !== null
```
VadimKovalenkoSNF, line 163: Can we use just `claim.claim_decision` instead of explicit null check?

src/react/src/util/util.js
```diff
@@ -1155,10 +1155,10 @@ function descendingComparator(a, b, orderBy) {
         bValue = bValue.toLowerCase();
     }
 
-    if (bValue < aValue) {
+    if (aValue === null || bValue < aValue) {
```
VadimKovalenkoSNF, line 1158: Why this null check is needed here? Is it fix some edge case?

VadimKovalenkoSNF (commented):
Ask few questions.

src/react/src/util/util.js
roman-stolar: @VadimKovalenkoSNF  We have requirements that empty/NULL values should be on top if we sort from recent to oldest & on bottom if we sort from oldest to recent. This case handled here.

VadimKovalenkoSNF: Got it, thank you!

src/react/src/components/DashboardClaimsListTable.jsx
VadimKovalenkoSNF: Ack.

VadimKovalenkoSNF (approved):
Approved

vladsha-dev (approved):
LGTM

PR #272, opened 2024-07-02
[OSDEV-705] Upload Error. "Could not convert string to float" for empty coordinate cells.

mazursasha1990:
[OSDEV-705](https://opensupplyhub.atlassian.net/browse/OSDEV-705) Upload Error. "Could not convert string to float" for empty coordinate cells.

Fixed the error “could not convert string to float” that occurred when a list contained columns for “lat” and “lng” and only some of the rows in these columns had data. As a result, rows are processed regardless of whether the values for “lat” and “lng” are present and valid, invalid, or empty.
Created an additional `RowCoordinatesSerializer` in the ContriCleaner to handle coordinate values ("lat" and "lng"). Moved the conversion of "lat" and "lng" into float point numbers from `FacilityListViewSet` to this serializer. 

[OSDEV-705]: https://opensupplyhub.atlassian.net/browse/OSDEV-705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


roman-stolar (dismissed):
LGTM

src/django/contricleaner/lib/serializers/row_serializers/row_coordinates_serializer.py
```diff
@@ -0,0 +1,34 @@
...
+        if not lat_str or not lng_str:
+            current['lat'] = None
+            current['lng'] = None
+            return current
+
+        try:
+            lat = float(lat_str)
+            lng = float(lng_str)
+
+            if not (-90 <= lat <= 90 and -180 <= lng <= 180):
```
VadimKovalenkoSNF, line 23: Please, assign `90` and `180` to constants. It looks like magic numbers for me.

VadimKovalenkoSNF (changes requested):
Looks good, asked for minor change.

src/django/contricleaner/lib/serializers/row_serializers/row_coordinates_serializer.py
roman-stolar: oh ye, good catch.

roman-stolar (approved):
LGTM

VadimKovalenkoSNF (approved):

PR #275, opened 2024-07-04
Configure error logging through django settings

vladsha-dev:
- Introduced a general format for all Python logs by updating the Django `LOGGING` constant.
- Disabled propagation for the `django` logger to the `root` logger to avoid log duplication.
- Removed unnecessary calls to the `basicConfig` method since only the configuration defined in the `LOGGING` constant in the settings.py file is considered valid by the current Django app.

Please refer to [the official documentation](https://docs.djangoproject.com/en/3.2/topics/logging/) on configuring Python logging in a Django app.


mazursasha1990 (approved):
LGTM

roman-stolar (approved):
LGTM

PR #286, opened 2024-07-12
[OSDEV-1006] Create new "production-locations" endpoint

VadimKovalenkoSNF:
Fix [OSDEV-1006](https://opensupplyhub.atlassian.net/browse/OSDEV-1006)
Boilerplates for production-locations endpoint.

[OSDEV-1006]: https://opensupplyhub.atlassian.net/browse/OSDEV-1006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

1. Introduce OpenSeachConnection class as Singleton to handle connetion to the OpenSearch.
2. Introduce OpenSearchService that implements search methods.
3. Introduce SearchInterface that defines abstarct search methods that implemented by OpenSearchService.
4. Introduce OpenSearchQueryBuilder as a part of Builder pattern implementation.
5. Introduce ProductionLocationsSerializer using Strategy pattern.
6. Handle client and server errors.
7. Move index name to list of reusable constants.
8. Add unit tests for v1 utils and OpenSearch query builder.

🙈  This PR won't contain integration tests because they will be implemented in a separate ticket: https://opensupplyhub.atlassian.net/browse/OSDEV-1153


src/django/api/services/opensearch.py
```diff
@@ -0,0 +1,21 @@
...
+
+logger = logging.getLogger(__name__)
+
+client = OpenSearch(
+    hosts=[{'host': host, 'port': port}],
+    use_ssl=False,
+    verify_certs=False
+)
+
+def search_production_locations(query_body):
```
vladsha-dev, line 15: Search functionality should be in a separate module to divide responsibilities, as there can be other special operations like deletions, reindexing, etc. I recommend creating an `opensearch` [package](https://realpython.com/python-modules-packages/#python-packages)(regular folder `services/opensearch`) and including `opensearch.py` and `search.py` (create one) in this package.

src/django/oar/urls.py
```diff
@@ -134,4 +134,8 @@
     path('api/geocoder/', views.get_geocoding, name='get_geocoding'),
 ] + static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
 
-urlpatterns = public_apis + internal_apis + info_apis
+opensearch_apis = [
```
vladsha-dev, line 137: While naming this variable, I suggest focusing on `v1` rather than on OpenSearch.

src/django/api/services/opensearch.py
vladsha-dev, line 1: 1. This module should have only one responsibility, which is to expose a connection to the OpenSearch cluster. You can fully copy the content of the opensearch.py module that I have implemented in this [PR](https://github.com/opensupplyhub/open-supply-hub/pull/267/files#diff-dfcb74aa55ba936b3f4401fd8728021a509a709f5600852d388cb6e80e7c7fe6). To make it work, you need to introduce environment variables. After implementing it, you can just do:
`opensearch = OpenSearchServiceConnection()`
`client = opensearch.client`

vladsha-dev (changes requested):
Hi @VadimKovalenkoSNF,
Left several comments. Please consider them.

src/django/api/services/opensearch.py
VadimKovalenkoSNF: I will do this once finished with preparations before deploy to Dev for testing.

VadimKovalenkoSNF:
Rebased on top of https://github.com/opensupplyhub/open-supply-hub/pull/267

src/django/api/services/__init__.py
vladsha-dev, line 1: What is the purpose of having this `__init__.py` file?

src/django/api/views/__init__.py
```diff
@@ -71,3 +71,5 @@
 from .product_types import product_types
 from .sectors import sectors
 from .claim_statuses import claim_statuses
+
+from .v1.production_locations.production_locations import ProductionLocations
```
vladsha-dev, line 75: No need to import it here.
As we agreed, let's use only absolute imports.

src/django/api/views/v1/opensearch_query_builder/__init__.py
vladsha-dev, line 1: What is the purpose of having this __init__.py file?
I don’t think it is necessary since we are using absolute imports wherever possible.

vladsha-dev (changes requested):

src/django/oar/urls.py
```diff
@@ -135,4 +135,9 @@
     path('api/geocoder/', views.get_geocoding, name='get_geocoding'),
 ] + static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
 
-urlpatterns = public_apis + internal_apis + info_apis
+api_v1 = [
+     path('api/v1/production-locations',
+          views.ProductionLocations.as_view(), name='production_locations')
```
vladsha-dev, line 140: Please, use absolute import for the `ProductionLocations` view. It will make the location of the class more obvious.

src/django/api/views/v1/opensearch_query_builder/opensearch_query_builder_interface.py
```diff
@@ -0,0 +1,40 @@
+from abc import abstractmethod
+
+
+class OpenSearchQueryBuilderInterface:
```
roman-stolar, line 4: Will we reused it in the future?

VadimKovalenkoSNF: Most likely, no.

src/django/api/views/v1/opensearch_query_builder/__init__.py
VadimKovalenkoSNF: My intention was to make `opensearch_query_builder` to act as a module, so it's classes can be imported like: `from api.views.v1.opensearch_query_builder import OpenSearchQueryDirector, OpenSearchQueryBuilder`. This looks like absolute import for me. Otherwise, could you give me an example of the good usage of absolute imports in hte codebase, so I can refer to it?

src/django/api/views/v1/opensearch_query_builder/opensearch_query_builder_interface.py
vladsha-dev: > Will we reused it in the future?

It can be used to create a new query builder for other GET endpoints.

VadimKovalenkoSNF: > It can be used to create a new query builder for other GET endpoints.

Thats possible, but for now I have only once common `OpenSearchQueryBuilder`. I introduced`OpenSearchQueryBuilderInterface` in this PR just for sake of clean architecture and for delelopers who want to quickly check which methods are implemented in query builder.

vladsha-dev: Agreed, good idea that fully aligns with the Builder pattern

src/django/api/views/v1/opensearch_query_builder/__init__.py
vladsha-dev: It should be:
```
from api.views.v1.opensearch_query_builder.opensearch_query_builder import OpenSearchQueryBuilder
from api.views.v1.opensearch_query_builder.opensearch_query_director import OpenSearchQueryDirector
```

roman-stolar:
@VadimKovalenkoSNF Will approve when all comments will be resolved.

src/django/oar/urls.py
vladsha-dev: Yes, as I recall from the discussion with Peter, we are moving away from this approach.

src/django/api/services/search.py
```diff
@@ -0,0 +1,24 @@
...
+from .opensearch import OpenSearchServiceConnection
+from .search_interface import SearchInterface
+from opensearchpy.exceptions import OpenSearchException
+
+logger = logging.getLogger(__name__)
+
+
+class OpenSearchService(SearchInterface):
+    def __init__(self):
+        self.client = OpenSearchServiceConnection().client
```
vladsha-dev, line 11: I think you can clearly indicate that is private by using two underscores, since it is used only internally.
E.g. `self.__client = OpenSearchServiceConnection().client`

vladsha-dev (changes requested):

vladsha-dev (changes requested):
I noticed that 400 and 500 errors are documented for the v`1/production-locations` GET endpoint. I think throwing them should also be implemented.

1. For instance, handling an OpenSearchException was suppressed. But as I see it, that is a 500 error and can be handled in the `ProductionLocation` view.
2. Additionally, the query parameter validation needs to be implemented according to the API spec.
Screenshot 2024-07-31 at 15 07 34

src/django/api/views/v1/opensearch_query_builder/opensearch_query_director.py
```diff
@@ -0,0 +1,63 @@
...
+            elif query_type == 'geo_distance':
+                lat = query_params.get(f'{field}[lat]')
+                lon = query_params.get(f'{field}[lon]')
+                distance = query_params.get('distance', '10km')
+                if lat and lon:
+                    self.__builder.add_geo_distance(
+                        field,
+                        float(lat),
+                        float(lon),
+                        distance)
```
protsack-stephan, line 45: This part of the code seems fairly complicated. Is there any way we can refactor this to avoid nested if statements or move it to a separate method?

src/django/api/views/v1/opensearch_query_builder/opensearch_query_builder.py
```diff
@@ -0,0 +1,147 @@
...
+        max_value = int(max_value) if max_value else None
+
+        range_query = {}
+        if min_value is not None:
+            range_query['gte'] = min_value
+        if max_value is not None:
+            range_query['lte'] = max_value
+
+        if range_query:
+            if field == 'number_of_workers':
```
protsack-stephan, line 71: There are multiple parts of the code where we look for a specific field name to decide how to build a search query. Is there any way we can do something similar to what we do in `OpenSearchQueryDirector`, by abstracting the code from using specific field names and defining some kind of mapping to refer to?

protsack-stephan (approved):
Awesome work! Here are a couple of comments from me:

P.S. This is just my opinion in this code review. It's your choice whether to act on these comments or ignore them.

src/django/api/views/v1/opensearch_query_builder/opensearch_query_builder.py
VadimKovalenkoSNF: This is a good point, but I decided to not use field-name-method mapping here because of decreasing of the code readability. Some builder methods in `OpenSearchQueryBuilder` accept `filed` name so it is ok **for me** that these particular methods will act depending on `field` value they receive. In the opposite, I need to create and maintain separate `_add_number_of_workers_range` which I'd like to avoid (at least for now). I agree that this is leaky abstraction, and `OpenSearchQueryBuilder` shouldn't suppose to know about field names, but I'd like to stick with strict OpenSearch queries naming (eg `add_range` , `add_terms`, `add_geo_distance` etc.) to preserve single responsibility principle here and expand those methods if necessary.

src/django/api/views/v1/opensearch_query_builder/opensearch_query_director.py
VadimKovalenkoSNF: This has been refactored, thank you for noticing.

src/django/api/services/search.py
```diff
@@ -0,0 +1,24 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class OpenSearchService(SearchInterface):
+    def __init__(self):
+        self.__client = OpenSearchServiceConnection().client
+
+    def search_production_locations(self, query_body):
+        index_name = 'production-locations'
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: This is a VERY good point. I've added PRODUCTION_LOCATIONS_INDEX into an `index_names.py` const file and applied it where it was possible, pls check.

src/django/api/services/search.py
VadimKovalenkoSNF: Done, pls check

src/django/api/views/__init__.py
VadimKovalenkoSNF: Done.

src/django/api/views/v1/opensearch_query_builder/opensearch_query_director.py
```diff
@@ -0,0 +1,62 @@
+class OpenSearchQueryDirector:
+    def __init__(self, builder):
+        self.builder = builder
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Done.

src/django/api/views/v1/opensearch_query_builder/opensearch_query_director.py
```diff
@@ -0,0 +1,62 @@
...
+        self._opensearch_template_fields = {
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Done.

src/django/api/views/v1/production_locations/production_locations.py
```diff
@@ -0,0 +1,24 @@
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from api.services.search import OpenSearchService
+from api.views.v1.opensearch_query_builder \
+    import OpenSearchQueryDirector, OpenSearchQueryBuilder
+
+
+class ProductionLocations(APIView):
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: 1. `ViewSet` is now using instead of `APIView` for ProductionLocations class.
2. I created handlers for client and server errors. For 500 error this is specifically an error about losing connection to the OpenSearch. For client error:  error triggers for improper setting of `number_of_workers`, `percent_female_workers` and `coordinates`. If we define rules for other parameters, we can expand param serializer class (aka `ProductionLocationsSerializer`) that I've introduced in this PR.

src/django/api/views/v1/opensearch_query_builder/opensearch_query_builder.py
protsack-stephan: Sounds fair.

src/django/api/views/v1/production_locations.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.opensearch_service = OpenSearchService()
+        self.opensearch_query_builder = OpenSearchQueryBuilder()
+
+    def list(self, request):
+        opensearch_query_director = OpenSearchQueryDirector(
+            self.opensearch_query_builder
+        )
+        query_params = request.GET
```
protsack-stephan, line 29: I see that this variable is used twice, but it looks redundant; simply passing `request.GET` should be sufficient. Alternatively, if we are to keep this variable, we should probably rename it to something like `request_method`.

One more thing: is there a property like `method` in Django's `request` object, something like `request.method`? If there is, we should be able to avoid using this variable at all.

src/django/api/services/search.py
```diff
@@ -0,0 +1,32 @@
...
+    ):
+        self.message = message
+        super().__init__(self.message)
+
+
+class OpenSearchService(SearchInterface):
+    def __init__(self):
+        self.__client = OpenSearchServiceConnection().client
+
+    def search_production_locations(self, index_name, query_body):
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Good point. Fixed.

src/django/api/views/v1/production_locations.py
VadimKovalenkoSNF: I refactored this part of code, pls re-check.

src/django/api/views/v1/opensearch_query_builder/opensearch_query_director.py
```diff
@@ -0,0 +1,77 @@
...
+            elif query_type == 'terms':
+                values = query_params.getlist(field)
+                self.__add_terms_query(field, values)
+            elif query_type == 'range':
+                self.__add_range_query(field, query_params)
+            elif query_type == 'geo_distance':
+                lat = query_params.get(f'{field}[lat]')
+                lon = query_params.get(f'{field}[lon]')
+                distance = query_params.get('distance', '10km')
+                self.__add_geo_distance_query(field, lat, lon, distance)
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Agree, this variant is more readable. Fixed.

src/django/api/serializers/v1/opensearch_common_validators/coordinates_validator.py
```diff
@@ -0,0 +1,30 @@
...
+    def validate_opensearch_params(self, data, errors):
+        lat = data.get('coordinates_lat')
+        lon = data.get('coordinates_lon')
+
+        if ((lat is not None and lon is None) or
+                (lat is None and lon is not None)):
+            errors.append({
+                "field": "coordinates",
+                "message": "Both latitude and longitude must be provided."
+            })
```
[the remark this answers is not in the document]
protsack-stephan: Yes, you understood me correctly.

src/django/api/views/v1/production_locations.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class ProductionLocations(ViewSet):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.opensearch_service = OpenSearchService()
+        self.opensearch_query_builder = OpenSearchQueryBuilder()
+
+    def list(self, request):
+        if request.method == 'GET':
```
protsack-stephan, line 26: Can we use [view decorators](https://docs.djangoproject.com/en/5.0/topics/http/decorators/#django.views.decorators.http.require_http_methods) instead of checking for the `GET` request?

protsack-stephan: Alternatively, we can do something do a negative check here:

```python
if request.method != request.GET:
  throw MethodNotAllowed()
```

protsack-stephan (approved):
Looks good, last set of small comments from me. Awesome work!

roman-stolar (dismissed):
LGTM, Thank you!

src/django/api/serializers/v1/opensearch_error_list_builder.py
```diff
@@ -0,0 +1,11 @@
+class OpenSearchErrorListBuilder:
+    def __init__(self, validators):
+        self.validators = validators
+
+    def build_error_list(self, data):
+        errors = []
+        for validator in self.validators:
+            validator_errors = validator.validate(data)
+            if validator_errors:
+                errors.extend(validator_errors)
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Done.

src/django/api/views/v1/production_locations.py
VadimKovalenkoSNF: Let's stick with `raise MethodNotAllowed()` here because forementioned decorator is generally used with function-based views, not class-based views. I've updated the code.

vladsha-dev: @VadimKovalenkoSNF 
Do we really need this check? As I understand it, the `def list` method is considered by Django as the GET method be default.

In the case of custom methods, we need to use decorators.

VadimKovalenkoSNF: Yes, you are right. I've double checked it, and it's redundant. Code has been updated.

vladsha-dev (approved):
LGTM

roman-stolar (approved):
LGTM

PR #294, opened 2024-07-18
[OSDEV-1145] Error message appearing as red dot with no context

Innavin369:
[OSDEV-1145] (https://opensupplyhub.atlassian.net/browse/OSDEV-1145) Error message appearing as a red dot with no context.

* Removed const variable with components and error messages, tests
* Added function to parse tag
* Added function to check error string contains a tag

The task which was implemented [OSDEV-1105](https://opensupplyhub.atlassian.net/browse/OSDEV-1105)

[OSDEV-1145]: https://opensupplyhub.atlassian.net/browse/OSDEV-1145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OSDEV-1105]: https://opensupplyhub.atlassian.net/browse/OSDEV-1105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/react/src/components/ContributeForm.jsx
```diff
@@ -94,6 +94,30 @@ class ContributeForm extends Component {
 
     handleUploadList = () => this.props.uploadList(this.fileInput);
 
+    containsTag = (content, tagName) => {
+        const tagRegex = new RegExp(`<${tagName}>.*?<\\/${tagName}>`, 'i');
+        return tagRegex.test(content);
+    };
+
+    parseContent = content => {
+        if (!this.containsTag(content, 'strong')) {
```
VadimKovalenkoSNF, line 103: Why you need to check tags here? Is this come from the backend?

VadimKovalenkoSNF (changes requested):
Left concern about parsing HTML in React component.

src/react/src/components/ContributeForm.jsx
Innavin369: No. it's not. I check if a string contains tags not to proceed with function logic if it's no sens. Moved logic to separate function.

VadimKovalenkoSNF: Why string can contain tags?

src/react/src/util/constants.jsx
```diff
@@ -1272,18 +1272,3 @@ export const facilityClaimStepsNames = Object.freeze({
     SUPPORT_DOC: 'Support Documentation',
     ADDITIONAL_DATA: 'Additional Data',
 });
-
-export const componentsWithErrorMessage = Object.freeze({
```
vladsha-dev, line 1276: Why did you decide to delete `componentsWithErrorMessage` ?

vladsha-dev (changes requested):

src/react/src/components/ContributeForm.jsx
Innavin369: I added the tag to style List Name bold. It's requirements from the client.

src/react/src/util/constants.jsx
Innavin369: I added this componentsWithErrorMessage in the previous PR. But the rest of errors messages(from BE) stayed as string and that became the reason of the bug.

src/react/src/components/ContributeForm.jsx
VadimKovalenkoSNF: Where can I see this requirement? OSDEV-1145 doesn't contain any links to origin ticket.

Innavin369: My bad https://opensupplyhub.atlassian.net/browse/OSDEV-1105. Added to PR description

VadimKovalenkoSNF: Ok, I see it now. Still, attempting to parse HTML using regular expression [considering as antipattern](https://stackoverflow.com/a/1732454). Please, find another solution that solve this issue and keep the codebase consistent at the same time.

src/react/src/util/constants.jsx
vladsha-dev: Then, you can make each error message just a string.

vladsha-dev: Then you can make each error message just a string.

PR #296, opened 2024-07-19
Disable using legacy template endpoint for OpenSearch

vladsha-dev:
- This PR removes the unintended behaviour where OpenSearch applies [dynamic mapping](https://opensearch.org/docs/latest/field-types/#dynamic-mapping) instead of [explicit mapping](https://opensearch.org/docs/latest/field-types/#explicit-mapping) defined in the Logstash app. For instance, this is extremely important for the `coordinates` field, which, by dynamic mapping, is recognized as a simple object with `lat` and `lon` properties. However, it should be of [the geo_point type](https://opensearch.org/docs/latest/field-types/supported-field-types/geo-point/), which is manually set via the `production_locations.json` index template to enable [Geopolygon queries](https://opensearch.org/docs/latest/query-dsl/geo-and-xy/geopolygon/) for the search performed by drawing the boundary on the map.
Screenshot 2024-07-19 at 19 23 50

**Tech Details:**
To fix that bug, the legacy `_template` API endpoint was disabled via the configuration file in favor of the new `_index_template` API endpoint, since [the composable index template](https://opensearch.org/docs/latest/im-plugin/index-templates/#composable-index-templates) is used for OpenSearch. The `legacy_template` was set to `false` to start using the defined composable index template in the `production_locations.json` file. See [the documentation](https://opensearch.org/docs/latest/tools/logstash/ship-to-opensearch/#optional-parameters) about the `legacy_template` configuration parameter. This change is necessary to avoid omitting the `production_locations.json` index template for the `production-locations` index defined in the Logstash app and to enforce the OpenSearch cluster to use the [explicit mapping](https://opensearch.org/docs/latest/field-types/#explicit-mapping) for the `production-locations` index.


VadimKovalenkoSNF (approved):
Approved

roman-stolar (approved):
LGTM

PR #310, opened 2024-08-07
[OSDEV-982] Search, API. Add OS ID query parameter to v1/production-locations.

VadimKovalenkoSNF:
Fix [OSDEV-982](https://opensupplyhub.atlassian.net/browse/OSDEV-982). Add os_id query an path param

[OSDEV-982]: https://opensupplyhub.atlassian.net/browse/OSDEV-982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/views/v1/opensearch_query_builder/opensearch_query_builder.py
```diff
@@ -41,8 +41,13 @@ def add_multi_match(self, query):
...
             return self.query_body
-        terms_field = f'{field}.alpha_2' if field == 'country' \
-            else f'{field}.keyword'
+
+        if (field == 'country'):
+            terms_field = f'{field}.alpha_2'
+        elif (field == 'os_id'):
+            terms_field = field
+        else:
+            terms_field = f'{field}.keyword'
```
protsack-stephan, line 50: Please note that this set of if statements is starting to grow. If it continues to expand, it might become problematic to maintain. We need to consider a more extendable solution in the long run. There's nothing to do for now, but it's something we need to keep in mind the next time we try to extend this code.

protsack-stephan (approved):
LGTM! One small comment from me.

vladsha-dev (approved):
LGTM

vladsha-dev (changes requested):
Could you please add validation for the `os_id` query parameter to ensure it is a string?

src/django/api/views/v1/opensearch_query_builder/opensearch_query_builder.py
roman-stolar: @VadimKovalenkoSNF There is two useful solution:
- Using **match -> case**
- Pre-define options for each case
```
build_options = {
  'country': build_country
  'os_id': build_os_id
}

action = build_options.get(field)
```

VadimKovalenkoSNF:
> Could you please add validation for the os_id query parameter to ensure it is a string?

@vladsha-dev , do we really need any validation here? OS contains from letters and numbers, 15 characters length. If we need to apply validation here, there should be some kind of regular expression. Is this what you expected?

roman-stolar:
> > Could you please add validation for the os_id query parameter to ensure it is a string?
> 
> @vladsha-dev , do we really need any validation here? OS contains from letters and numbers, 15 characters length. If we need to apply validation here, there should be some kind of regular expression. Is this what you expected?

@VadimKovalenkoSNF I suppose it will be enough to just check that params contain letter & not only numbers.

roman-stolar:
But I agree that it's more options things that just good to have.

VadimKovalenkoSNF:
I've found this function `validate_os_id` but not sure whether we need it in the OpenSearch right now.

vladsha-dev (dismissed):
LGTM

roman-stolar (approved):
LGTM

doc/release/RELEASE-NOTES.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Code/API changes
 * [OSDEV-1006](https://opensupplyhub.atlassian.net/browse/OSDEV-1006) - Create new "api/v1/production-locations" endpoint.
+* [OSDEV-982](https://opensupplyhub.atlassian.net/browse/OSDEV-982) - Search, API. Add OS ID query parameter to v1/production-locations.
```
vladsha-dev, line 22: The `api/v1/production-locations/{os_id}` endpoint has also been implemented. Please update the RELEASE-NOTES.md file accordingly.

vladsha-dev (changes requested):
Minor comment

vladsha-dev (approved):
LGTM

roman-stolar (approved):

PR #348, opened 2024-09-06
[OSDEV-1177] Templates for generating OpenSearch index cleaning scripts

roninzp:
[[OSDEV-1177](https://opensupplyhub.atlassian.net/browse/OSDEV-1177)] 

Script templates for running on a bastion host so that changes can be made in one place and not in each pipeline separately.
Stop/start Logstash and clearing OpenSearch indexes moved to separate jobs of pipelines.
Stop/start Logstash and clearing OpenSearch indexes now runs on ubuntu-latest runner.

[OSDEV-1177]: https://opensupplyhub.atlassian.net/browse/OSDEV-1177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/db_apply_anonimized.yml
```diff
@@ -65,27 +65,24 @@ jobs:
           echo "EFS_AP_ID=$EFS_AP_ID" >> $GITHUB_OUTPUT
           echo "OPENSEARCH_DOMAIN=$OPENSEARCH_DOMAIN" >> $GITHUB_OUTPUT
       - name: Clear the OpenSearch indexes for ${{ vars.ENV_NAME }}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: "eu-west-1"
```
vladsha-dev, line 71: Do we need the AWS_DEFAULT_REGION here?

.github/workflows/deploy_to_aws.yml
```diff
@@ -304,27 +304,24 @@ jobs:
           echo "OPENSEARCH_DOMAIN=$OPENSEARCH_DOMAIN" >> $GITHUB_OUTPUT
       - name: Clear the OpenSearch indexes for ${{ vars.ENV_NAME }}
         if: ${{ (vars.ENV_NAME != 'Production' && vars.ENV_NAME != 'Staging' && vars.ENV_NAME) && inputs.restore-db == true}}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: "eu-west-1"
```
vladsha-dev, line 310: Do we need the AWS_DEFAULT_REGION here?

src/clear_opensearch/clear_opensearch.sh.tpl
vladsha-dev, line 1: Please move the `clear_opensearch` folder to the ./deployment folder, as we typically save application code in the ./src folder.

src/clear_opensearch/run.sh.tpl
vladsha-dev, line 1: Please move the clear_opensearch folder to the ./deployment folder, as we typically save application code in the ./src folder.

vladsha-dev (changes requested):
1. I left the comments.
2. Please update the `RELEASE-NOTES.md` file to reflect the improvements you have made.

deployment/terraform/clear_opensearch/clear_opensearch.sh.tpl
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+echo -e "\nDelete OpenSearch index\n"
```
vladsha-dev, line 2: Let's make it plural — **indexes**. In the near future, we will have a few more indexes, so it won't just be one.

deployment/terraform/clear_opensearch/run.sh.tpl
```diff
@@ -0,0 +1,9 @@
+chmod 700 /root/.ssh
+chmod 600 /root/.ssh/id_rsa
+chmod 600 /root/.ssh/config
+chown -R root:root /root/.ssh
+chmod +x /script/clear_opensearch.sh
+echo -e "\nCopy script to Bastion host\n"
+scp /script/clear_opensearch.sh ec2-user@$BASTION_IP:
+echo -e "\nRun cleaning OpenSearch index\n"
```
vladsha-dev, line 8: Let's make it plural — **indexes**. In the near future, we will have a few more indexes, so it won't just be one.

.github/workflows/deploy_to_aws.yml
```diff
@@ -31,6 +31,11 @@ on:
         required: false
         type: boolean
         default: false
+      clear-opensearch:
+        description: 'Clear OpenSearch index'
```
vladsha-dev, line 35: Let's make it plural — **indexes**. In the near future, we will have a few more indexes, so it won't just be one.

.github/workflows/deploy_to_aws.yml
```diff
@@ -262,73 +267,28 @@ jobs:
...
-              -v ./script:/script \
-              -v ./ssh:/root/.ssh \
-              kroniak/ssh-client bash /script/run.sh
+              --service=OpenSupplyHub${{vars.ENV_NAME}}AppLogstash
+
+  restore_database:
+    needs: stop_logstash
+    runs-on: self-hosted
+    environment: ${{ inputs.deploy-env || (github.ref_name == 'main' && 'Development') || (startsWith(github.ref_name, 'releases/') && 'Pre-prod') || (startsWith(github.ref_name, 'production-') && 'Production') || (startsWith(github.ref_name, 'sandbox-') && 'Staging') || 'None' }}
+    if:  ${{ inputs.deploy-plan-only == false }}
```
vladsha-dev, line 283: Please, add a check for inputs.restore-db == true to ensure this step doesn't start if it's false.

.github/workflows/deploy_to_aws.yml
```diff
@@ -248,11 +253,11 @@ jobs:
...
           AWS_DEFAULT_REGION: "eu-west-1"
 
-  restore_database:
+  stop_logstash:
     needs: create_kafka_topic
-    runs-on: self-hosted
+    runs-on: ubuntu-latest
     environment: ${{ inputs.deploy-env || (github.ref_name == 'main' && 'Development') || (startsWith(github.ref_name, 'releases/') && 'Pre-prod') || (startsWith(github.ref_name, 'production-') && 'Production') || (startsWith(github.ref_name, 'sandbox-') && 'Staging') || 'None' }}
-    if:  ${{ inputs.deploy-plan-only == false }}
+    if: ${{ inputs.deploy-plan-only == false }}
```
vladsha-dev, line 260: Please, add a check to ensure it only starts if `restore-db` or `clear-opensearch` are `true`. Without this, we will stop Logstash even when we don't need to clear OpenSearch, which is unnecessary.

.github/workflows/deploy_to_aws.yml
```diff
@@ -412,3 +362,83 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: "eu-west-1"
 
+  clear_opensearch:
```
vladsha-dev, line 365: Please, add a check to ensure it only starts if restore-db or clear-opensearch are true. Without this, we will clear OpenSearch even when we don't need to clear OpenSearch.

.github/workflows/deploy_to_aws.yml
```diff
@@ -412,3 +362,83 @@ jobs:
...
+          echo "" >> ssh/id_rsa
+          echo -n ${{ vars.BASTION_IP }} > script/.env
+          envsubst < clear_opensearch.sh.tpl > script/clear_opensearch.sh
+          envsubst < run.sh.tpl > script/run.sh
+          docker run --rm  \
+              -v ./script:/script \
+              -v ./ssh:/root/.ssh \
+              kroniak/ssh-client bash /script/run.sh
+
+  start_logstash:
```
vladsha-dev, line 424: Please, add a check to ensure it only starts if restore-db or clear-opensearch are true. Without this, we will start Logstash even when we don't need to clear OpenSearch.

vladsha-dev (changes requested):

vladsha-dev (changes requested):
@roninzp, please update the PR description to reflect all the main changes you made.

vladsha-dev (changes requested):
Since we use our custom `[Release] Deploy` to trigger deploying changes to Staging and Production by creating a tag, which in turn triggers Deploy to AWS, you also should update the `[Release] Deploy` workflow. Add the same `clear-opensearch` flag, which should then be passed to the `Deploy to AWS` workflow as an input (**probably**) to trigger clearing OpenSearch if needed.

vladsha-dev (changes requested):
@roninzp 
1. Do we need to use an exact IP for production?
Screenshot 2024-09-12 at 19 21 03

I noticed that for Staging, you used `bastion.staging.opensupplyhub.org` without any numbers.

2. Have you checked the AWS credentials for Staging and Production? Will it be possible to perform a curl request from EC2 to OpenSearch? Have you configured this?

deployment/terraform/clear_opensearch/run.sh.tpl
```diff
@@ -0,0 +1,9 @@
+chmod 700 /root/.ssh
```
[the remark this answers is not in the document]
dmovd: no, because we run it by 
```
 bash /script/run.sh
```
we don't need to run an additional shell

.github/workflows/deploy_to_aws.yml
dmovd: We discussed that at the post-deployment stage there may be migrations affecting the data schema and therefore Logstash should not be running at this time. Will there be a situation where the database does not need to be restored, but migrations will still occur?

.github/workflows/deploy_to_aws.yml
dmovd: We discussed that at the post-deployment stage there may be migrations affecting the data schema and therefore Logstash should not be running at this time. Will there be a situation where the database does not need to be restored, but migrations will still occur?

.github/workflows/deploy_to_aws.yml
dmovd: We discussed that at the post-deployment stage there may be migrations affecting the data schema and therefore Logstash should not be running at this time. Will there be a situation where the database does not need to be restored, but migrations will still occur?

.github/workflows/deploy_to_aws.yml
dmovd: We discussed that at the post-deployment stage there may be migrations affecting the data schema and therefore Logstash should not be running at this time. Will there be a situation where the database does not need to be restored, but migrations will still occur?

deployment/terraform/clear_opensearch/clear_opensearch.sh.tpl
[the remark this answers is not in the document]
roninzp: done

doc/release/RELEASE-NOTES.md
```diff
@@ -22,7 +22,13 @@ To make this possible, the `sync_production_locations.sql` script, which generat
 Additionally, a `historical_os_id` filter was added to the `sync_production_locations.conf`, ensuring that the `historical_os_id` is included in the index document only when the `historical_os_id_value` is not empty.
 
 ### Architecture/Environment changes
-* *Describe architecture/environment changes here.*
+* [OSDEV-1153](https://opensupplyhub.atlassian.net/browse/OSDEV-1153) - Created integration tests for the OpenSearch and for new `/api/v1/production-locations/` API endpoint.
```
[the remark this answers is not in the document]
roninzp: fixed

doc/release/RELEASE-NOTES.md
```diff
@@ -22,7 +22,13 @@ To make this possible, the `sync_production_locations.sql` script, which generat
...
+* [OSDEV-1177](https://opensupplyhub.atlassian.net/browse/OSDEV-1177)
+  - Improved OpenSearch index cleanup step in the pipeline to use script templates so that changes can be made in one place rather than in each pipeline separately
+  - Stop/start Logstash and clearing OpenSearch indexes moved to separate jobs of pipeline.
+  - Stop/start Logstash and clearing OpenSearch indexes now runs on ubuntu-latest runner.
+
+
```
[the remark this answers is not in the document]
roninzp: fixed

.github/workflows/deploy_to_aws.yml
roninzp: done

.github/workflows/deploy_to_aws.yml
vladsha-dev: It is the `restore_database` job, which should only be started if `inputs.restore-db` equals `true`. All the steps inside it are triggered only if this condition is true. See the screenshots. I think you can even remove the checks for `inputs.restore-db == true` within the steps if this check is applied at the `restore_database` job level.
1. Screenshot 2024-09-13 at 11 07 19
--------------------
2. Screenshot 2024-09-13 at 11 07 43

.github/workflows/deploy_to_aws.yml
vladsha-dev: There are two cases to consider:

1. If a developer knows that OpenSearch needs to be reindexed due to changes from DB migrations or OpenSearch index mapping modifications, they should set the "clear opensearch" flag via the `Deploy to AWS` or `Release [Deploy]` workflows. Only then should this job be activated.
2. If the "restore db" flag is set in the `Deploy to AWS` workflow, it means that Logstash and OpenSearch should be cleared.

In all other cases, this job should not be activated.

.github/workflows/deploy_to_aws.yml
vladsha-dev: There are two cases to consider:

1. If a developer **knows** that OpenSearch needs to be reindexed due to changes from DB migrations or OpenSearch index mapping modifications, they should set the "clear opensearch" flag via the `Deploy to AWS` or `Release [Deploy]` workflows. Only then should this job be activated.
2. If the "restore db" flag is set in the `Deploy to AWS` workflow, it means that Logstash and OpenSearch should be cleared.

In all other cases, this job should not be activated.

roninzp:
> Do we need to use an exact IP for production?
> I noticed that for Staging, you used bastion.staging.opensupplyhub.org without any numbers.

We can change this variable when DNS record will be fixed

> Have you checked the AWS credentials for Staging and Production? Will it be possible to perform a curl request from EC2 to OpenSearch? Have you confirmed this?

I added terraform-ci IAM user as master user to OpenSearch. Since I don't have the keys for this user, I can't check the request in OpenSearch.

.github/workflows/deploy_to_aws.yml
roninzp: yes. 
we must specify the region to execute requests via aws cli

.github/workflows/deploy_to_aws.yml
vladsha-dev: @roninzp it looks like it hasn't been done.
I meant to add the check `inputs.restore-db == true` at the `restore_database` job level and remove this check from each step under this job.

vladsha-dev: Why can't this check be done on the job level?

.github/workflows/deploy_to_aws.yml
vladsha-dev: Why can't this check be done on the job level?

.github/workflows/deploy_to_aws.yml
vladsha-dev: @roninzp 
Why can't this check be done on the job level? To avoid starting this job.

.github/workflows/deploy_to_aws.yml
vladsha-dev: @roninzp 
Why can't this check be done on the job level? To avoid starting this job. Will it block the next jobs if the check is at the job level?

vladsha-dev:
@roninzp 
`We can change this variable when DNS record will be fixed`
How to fix it? What is wrong with it?

.github/workflows/deploy_to_aws.yml
roninzp: if we add this check on job level, all followed jobs be skipped every time when triger not be set to true

.github/workflows/deploy_to_aws.yml
roninzp: if we add this check on job level, all followed jobs be skipped every time when triger not be set to true

.github/workflows/deploy_to_aws.yml
roninzp: if we add this check on job level, all followed jobs be skipped every time when triger not be set to true

.github/workflows/deploy_to_aws.yml
roninzp: if we add this check on job level, all followed jobs be skipped every time when triger not be set to true

roninzp:
> How can we fix it? Could you please let me know what is wrong?

DNS record will be updated next time when terrafor code be runned

vladsha-dev:
@roninzp 

> DNS record will be updated next time when terraform code be runned

Does this mean that for the upcoming deployment to production, it will work with 34.242.166.51, but after the release, we will need to update it to `bastion.production.opensupplyhub.org`?

.github/workflows/deploy_to_aws.yml
vladsha-dev: Got it, thank you!

.github/workflows/deploy_to_aws.yml
vladsha-dev: Got it, thank you!

.github/workflows/deploy_to_aws.yml
vladsha-dev: Got it, thank you!

.github/workflows/deploy_to_aws.yml
vladsha-dev: Got it, thank you!

vladsha-dev (approved):
LGTM, great job! Thank you for your answers!

@roninzp, please ask the other developers to review it as well

VadimKovalenkoSNF (approved):
Approved.

PR #356, opened 2024-09-20
[OSDEV-1039] make list parsing asynchronous

vladsha-dev:
[OSDEV-1039](https://opensupplyhub.atlassian.net/browse/OSDEV-1039) - Made the list parsing asynchronous and increased the list upload limit to 10,000 facilities per list to reduce manual work for moderators when they split large lists into smaller ones. The following architectural and code changes have been made:

1. Renamed the previously copied `api/facility-lists/createlist` POST endpoint to the `api/facility-lists` POST endpoint. Deleted the old implementation of the `api/facility-lists` POST endpoint along with the `use_old_upload_list_endpoint` switcher that was necessary to toggle between the old and new list upload endpoints.
2. Removed the triggering of ContriCleaner from the `api/facility-lists` POST endpoint and moved it to the async parse AWS batch job to reduce the load on the endpoint. Introduced a `parsing_errors` field for the FacilityList model to collect list-level and internal errors logged during the background parsing of the list.
3. Established a connection between the EC2 instance within the AWS batch job and the S3 bucket where all the uploaded list files are saved. This is necessary because the parse job retrieves a particular list from the S3 bucket via Django.
4. Deleted redundant code from the previous implementation of the list item parsing.
5. Adjusted Django, ContriCleaner, and integration tests. Regarding integration tests, the `facility_list_items.json` fixture was converted to concrete CSV lists, which were connected to the `facility_lists.json` fixture to upload them to the DB while creating the test DB for the integration tests. This is necessary because the parsing function that triggers ContriCleaner can only work with real files, not facility list items as it did previously.
6. Refactored the ContributeForm component in the front-end app.
7. The list page was adjusted to work with asynchronous parsing, and a new dialog window was added to notify users about the list parsing process, indicating that they need to wait.
8. Introduced a UI to display list parsing errors on the list page after the page refresh.

[OSDEV-1039]: https://opensupplyhub.atlassian.net/browse/OSDEV-1039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployment/terraform/iam.tf
```diff
@@ -153,6 +153,26 @@ resource "aws_iam_role_policy" "ses_send_email_from_batch" {
   policy = data.aws_iam_policy_document.ses_send_email.json
 }
 
+data "aws_iam_policy_document" "batch_s3_read_policy" {
```
roman-stolar, line 156: Why this changes needed, to get file?

src/django/api/views/facility/facility_list_view_set.py
```diff
@@ -206,176 +146,27 @@ def create(self, request):
...
 
-        header, rows = self._extract_header_rows(csv_file, request)
-
-        new_list = FacilityList(
+        new_list = FacilityList.objects.create(
             name=name,
             description=description,
-            file_name=csv_file.name,
-            file=csv_file,
-            header=header,
```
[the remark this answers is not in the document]
vladsha-dev: It will be filled in this file on row 49 -> https://github.com/opensupplyhub/open-supply-hub/pull/356/files#diff-4f818073994776a52735b5c329922f9ae8fd550f06d9529e79229abc1396c8b9R49
It will happen after receiving data from ContriCleaner.

deployment/terraform/iam.tf
vladsha-dev: Specifying the appropriate IAM policy for AWS Batch jobs is essential to ensure they have the necessary permissions to securely access S3 resources.
In AWS Batch jobs, we run different containers with different IAM permissions compared to the main ECS instances.

vladsha-dev: to get file?

> Yes, to get file from S3

src/django/api/views/facility/facility_list_view_set.py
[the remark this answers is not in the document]
roman-stolar: I see, thank you!

roman-stolar (approved):
LGTM

VadimKovalenkoSNF (approved):
Good work, approved.

PR #369, opened 2024-10-08
[OSDEV-1388] fix sharing anonymized snapshots with another account.

roninzp:
This changes increase timeout for copying shared snapshot to avoid the problem
```
botocore.exceptions.WaiterError: Waiter DBSnapshotCompleted failed: Max attempts exceeded
```

### DIFF
--- a/deployment/terraform/database_anonymizer_sheduled_task/docker/database_anonymizer.py
+++ b/deployment/terraform/database_anonymizer_sheduled_task/docker/database_anonymizer.py
@@ -113,7 +113,13 @@ response = source.copy_db_snapshot(
 )
 
 waiters = source.get_waiter('db_snapshot_completed')
-waiters.wait(DBSnapshotIdentifier=shared_snapshot_identifier)
+waiters.wait(
+    DBSnapshotIdentifier=shared_snapshot_identifier,
+    WaiterConfig={
+        'Delay': 15,
+        'MaxAttempts': 100
+    }
+)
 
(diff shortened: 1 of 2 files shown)


deployment/terraform/database_anonymizer_sheduled_task/docker/database_anonymizer.py
```diff
@@ -113,7 +113,13 @@
 )
 
 waiters = source.get_waiter('db_snapshot_completed')
-waiters.wait(DBSnapshotIdentifier=shared_snapshot_identifier)
+waiters.wait(
+    DBSnapshotIdentifier=shared_snapshot_identifier,
+    WaiterConfig={
+        'Delay': 15,
```
VadimKovalenkoSNF, line 119: Can we use something like AWS EventBridge instead of relying of delay of 15 seconds?

VadimKovalenkoSNF (commented):
Have a question.

deployment/terraform/database_anonymizer_sheduled_task/docker/database_anonymizer.py
roninzp: No, we must wait until the snapshot copy is complete before proceeding with the next steps.

VadimKovalenkoSNF: Ack.

VadimKovalenkoSNF (approved):
Approved.

roman-stolar (approved):
LGTM

PR #383, opened 2024-10-17
Increase open search version from 2.13 to 2.17

killev:
This pr increases Open Search version from 2.13 to 2.17 because 2.13 image [has been abandoned](https://opensearch.org/docs/2.13/install-and-configure/install-opensearch/docker/) and isn't maintained anymore


vladsha-dev (changes requested):
1. The version should also be updated in the TF code.
2. Are there any risks associated with updating it?

killev:
> 2. Are there any risks associated with updating it?

Not sure about risks. I consider them as low, but if we merge now, we have plenty time to test

docker-compose.yml
```diff
@@ -224,6 +224,7 @@ services:
       - POSTGRES_PASSWORD=opensupplyhub
       - POSTGRES_DB=opensupplyhub
       - LOGSTASH_UPDATE_INTERVAL_MINUTES=1
+      - LOG_LEVEL=INFO
```
vladsha-dev, line 227: It is by default `info`

vladsha-dev (changes requested):

roman-stolar (approved):

PR #385, opened 2024-10-18
[OSDEV-1387] Quickfix. Increased resources for the database in the production environment.

vladsha-dev:
- Increased the resources for the production database to `db.m6in.4xlarge` due to a new database downtime that occurred on _October 18, 2024, at 10:31 AM GMT+4_, following the initial increase of resources to `db.m6in.2xlarge`.
- The pre-production configurations of the database have been updated as well, since it should be a full copy of the production environment.


killev (approved):

mazursasha1990 (approved):
Approved.
Could you please rewrite the description in part "due to today's downtimes"? It's not clear what "today" means when someone will read it in the future.

PR #396, opened 2024-10-31
Updated RELEASE-NOTES.md file for 1.24 version

vladsha-dev:
Right now, to enable access from EC2 to OpenSearch, the IAM ARN of the special Terraform user needs to be added to the master user settings associated with the OpenSearch instance. Since it’s not possible to set multiple master users via Terraform, you will need to add one more manually through the AWS console, as this is permitted.

As the OpenSearch instance will be recreated during the release of version 1.24, the necessary instructions have been provided in the corresponding release notes 1.24 to reestablish access from EC2 to OpenSearch for clearing OpenSearch indexes.


Innavin369 (approved):

VadimKovalenkoSNF (approved):
Approved.

PR #407, opened 2024-11-06
[OSDEV-1158] API. Create POST endpoint for v1/production-locations to support SLC & API Mod.

vladsha-dev:
[[OSDEV-1158](https://opensupplyhub.atlassian.net/browse/OSDEV-1158)]
1. Introduced a new POST `/api/v1/production-locations/` endpoint based on the API v1 specification. This endpoint allows the creation of a new moderation event for the production location creation with the given details.
2. Removed redundant redefinition of paths via the `as_view` method for all the v1 API endpoints since they are already defined via `DefaultRouter`.
3. The following updates to the ModerationEvent table have been made:
    1. Set `uuid` as the primary key.
    4. Make `geocode_result` field optional. It can be blank if lat and lng
    have been provided by user.
    5. Remove redundant `blank=False` and `null=False` constraints, as these are
    the default values for model fields in Django and do not need to be
    explicitly set.
    6. Make `contributor` field non-nullable, as the field should not be left
    empty. It is required to have information about the contributor.
    7. Allow `claim` field to be blank. This change reflects the fact that
    a moderation event may not always be related to a claim, so the field can
    be left empty.

[OSDEV-1158]: https://opensupplyhub.atlassian.net/browse/OSDEV-1158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/views/v1/production_locations.py
```diff
@@ -16,19 +20,75 @@
...
+    def init_opensearch() -> Tuple[OpenSearchService, OpenSearchQueryDirector]:
+        opensearch_service = OpenSearchService()
+        opensearch_query_builder = ProductionLocationsQueryBuilder()
+        opensearch_query_director = OpenSearchQueryDirector(
+            opensearch_query_builder
         )
 
+        return (opensearch_service, opensearch_query_director)
+
+    def get_permissions(self):
```
VadimKovalenkoSNF, line 53: I suggest to separate concerns for this method like:

```
def get_permissions(self):
    """
    Returns the list of permissions for the ViewSet action methods.
    Combines action-specific and global application-level permissions.
    """
    action_permissions = self.get_action_permissions()
    combined_permission_classes = action_permissions + self.permission_classes
    return [permission() for permission in combined_permission_classes]

def get_action_permissions(self):
    """
    Returns the list of permissions specific to the current action.
    """
    if self.action == 'create':
        return [IsRegisteredAndConfirmed]
    return []
```

The `get_permissions` method now focuses solely on combining permissions.
The `get_action_permissions` method handles action-specific permissions logic.

src/django/api/views/v1/production_locations.py
```diff
@@ -16,19 +20,75 @@
...
     swagger_schema = None
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        self.opensearch_service = OpenSearchService()
-        self.opensearch_query_builder = ProductionLocationsQueryBuilder()
-        self.opensearch_query_director = OpenSearchQueryDirector(
-            self.opensearch_query_builder
+    @staticmethod
+    def init_opensearch() -> Tuple[OpenSearchService, OpenSearchQueryDirector]:
```
VadimKovalenkoSNF, line 44: Can you, pls, apply the same OpenSearch initialization for `src/django/api/views/v1/moderation_events.py` in sake of consistensy?

src/django/api/tests/test_production_locations_create.py
VadimKovalenkoSNF, line 1: I see that you added unit tests which is ok. But would you mind to apply couple of integration test cases in `src/tests/v1/` folder?

VadimKovalenkoSNF (changes requested):
Good work, thank you. I've added couple of comments and suggestions + do not forget to revise coderabbit insights.

src/django/api/models/moderation_event.py
```diff
@@ -94,8 +92,6 @@ class Source(models.TextChoices):
     )
 
     cleaned_data = models.JSONField(
-        null=False,
-        blank=False,
```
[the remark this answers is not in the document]
vladsha-dev: I removed redundant `blank=False` and `null=False` constraints, as these are
the default values for model fields in Django and do not need to be
explicitly set.

src/django/api/moderation_event_actions/creation/location_contribution/location_contribution.py
```diff
@@ -0,0 +1,48 @@
...
+        entry_location_data_processor = self.__setup_location_data_processors()
+
+        processed_location_data = entry_location_data_processor.process(
+            event_dto
+        )
+
+        return processed_location_data
+
+    @staticmethod
+    def __setup_location_data_processors() -> ContributionProcessor:
```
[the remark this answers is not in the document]
vladsha-dev: Thanks!
Fixed.

src/django/api/moderation_event_actions/creation/location_contribution/processors/production_location_data_processor.py
```diff
@@ -0,0 +1,158 @@
...
+from contricleaner.lib.exceptions.handler_not_set_error \
+    import HandlerNotSetError
+from api.moderation_event_actions.creation.location_contribution \
+    .processors.contribution_processor import ContributionProcessor
+from api.moderation_event_actions.creation.dtos.create_moderation_event_dto \
+    import CreateModerationEventDTO
+from api.constants import APIV1CommonErrorMessages
+
+
+# Initialize logger.
```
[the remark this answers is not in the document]
vladsha-dev: Done

src/django/api/views/v1/production_locations.py
vladsha-dev: Good point!
Fixed.

src/django/api/tests/test_production_locations_create.py
vladsha-dev: I don't think it makes a lot of sense, since the functionality of this endpoint is perfectly tested by Django's unit tests. This endpoint doesn't work via third-party apps, unlike the GET /api/v1/production-locations/ endpoint.

VadimKovalenkoSNF: Ack.

src/django/api/views/v1/production_locations.py
vladsha-dev: As for me, this class could be split:

- `validate_user_permissions`: This is validation that should be done by overriding the get_permissions function inside the ViewSet and creating a special class by instantiating permissions.BasePermission. See examples in the api.permissions module.
- `validate_uuid`, `fetch_moderation_event_by_uuid`, `validate_moderation_status`: These are methods that would fit perfectly in a serializer class.
- `handle_processing_error`: I think this can simply be raised directly since it only contains two lines of code.

These are just my thoughts to consider or not. I wouldn’t want to fix it within this PR; it’s just my thoughts.
I thought that service is something third-party.

vladsha-dev: If my changes align with what you were expecting, could we please resolve this thread?

VadimKovalenkoSNF: Yes, let's resolve it.

src/django/api/moderation_event_actions/creation/location_contribution/processors/production_location_data_processor.py
```diff
@@ -0,0 +1,158 @@
...
+                'detail': APIV1CommonErrorMessages.COMMON_INTERNAL_ERROR
+            }
+            event_dto.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
+
+            return event_dto
+
+        list_level_errors = cc_processed_data.errors
+        if list_level_errors:
+            transformed_list_level_errors = \
+                self.__trasform_cc_errors_to_api_v1_format(
```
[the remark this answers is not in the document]
vladsha-dev: Done

src/django/api/moderation_event_actions/creation/location_contribution/processors/geocoding_processor.py
```diff
@@ -0,0 +1,73 @@
...
+            self,
+            event_dto: CreateModerationEventDTO) -> CreateModerationEventDTO:
+        lat = event_dto.cleaned_data['fields'].get(
+            APIV1LocationContributionKeys.LAT
+        )
+        lng = event_dto.cleaned_data['fields'].get(
+            APIV1LocationContributionKeys.LNG
+        )
+
+        if not (lat and lng):
```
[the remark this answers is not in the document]
vladsha-dev: Done

VadimKovalenkoSNF (approved):

roman-stolar (approved):
LGTM

PR #422, opened 2024-11-25
Updated RDS instance for prod, preprod and test

Innavin369:
The RDS instance has been upgraded as follows: for `production` and `preprod`, it is now `db.m6in.8xlarge`, and for `test`, it has been upgraded to `db.t3.xlarge`.


vladsha-dev (approved):
Approved.
@Innavin369 could you please delete coderabits' description?

PR #432, opened 2024-11-28
Fix release notes

VadimKovalenkoSNF:
Remove duplicate section for release 1.26


doc/release/RELEASE-NOTES.md
vladsha-dev, line 1: Why do we need these release instructions for 1.26?

vladsha-dev (changes requested):

doc/release/RELEASE-NOTES.md
VadimKovalenkoSNF: I've splitted `search_after` into `search_after[value]` and `search_after[id]`. This will require reindexing of the OpenSearch. I took thees instructions from release `1.24`. Or you think they should be in release protocol only?

vladsha-dev: - If you are referring to [this PR](https://github.com/opensupplyhub/open-supply-hub/pull/420), then there is no need to clear OpenSearch, because you only updated the Django code and didn't change the index mapping.
- Regarding the updates you made to the `moderation-events` index mapping for the `contributor_id` field, there is no need to clear the index, because the `moderation-events` index has not been created in the production or staging environments. This is due to the fact that the `api_moderationevent` table is empty. That’s how Logstash and OpenSearch work.

vladsha-dev: As of now, I think we will need the same instructions as for the 1.25 release.

VadimKovalenkoSNF: @vladsha-dev Thanks for clarification, I will update release notes asap.

doc/release/RELEASE-NOTES.md
```diff
@@ -54,10 +27,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   - Split `search_after` parameter into `search_after_value` and `search_after_id` for better pagination control.
 
 ### Architecture/Environment changes
-* *Describe architecture/environment changes here.*
+* [OSDEV-1170](https://opensupplyhub.atlassian.net/browse/OSDEV-1170) - Added the ability to automatically create a dump from the latest shared snapshot of the anonymized database from Production environment for use in the Test and Pre-Prod environments.
 
 ### Bugfix
-* *Describe bugfix here.*
+* [OSDEV-1388](https://opensupplyhub.atlassian.net/browse/OSDEV-1388) - The waiter from boto3 cannot wait more than half an hour so we replaced it with our own.
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: There is only one section named ### Architecture/Environment changes

doc/release/RELEASE-NOTES.md
VadimKovalenkoSNF: Pls, re-check. I only left `reindex_database` command because we don't have migrations so far.

roman-stolar (approved):
LGTM

PR #440, opened 2024-12-04
Limit page number for api/facilities GET endpoint

vladsha-dev:
Limit the page number to 100 for the API/facilities GET endpoint.


VadimKovalenkoSNF (approved):
Approved.

PR #445, opened 2024-12-06
Fix OpenSearch index mapping updates

vladsha-dev:
It was found that clearing OpenSearch indexes didn’t work properly because the templates weren’t cleared. After updating the index mappings within the index template files, the index template remained unchanged because only the indexes were deleted during deployment, not both the indexes and their templates. This caused conflicts and prevented developers' updates from being applied to the OpenSearch indexes.

This issue has been fixed by adding additional requests to delete the appropriate index templates to the `clear_opensearch.sh.tpl` script, which is triggered when clearing OpenSearch during deployment to any environment.


VadimKovalenkoSNF (approved):
Good work! Approved.

VadimKovalenkoSNF (approved):
Re-approved.

mazursasha1990 (approved):
LGTM

PR #446, opened 2024-12-06
Increased RDS storage space in Production, Test, and Pre-prod environments.

vladsha-dev:
In light of recent instances(on 12/03/2024 UTC and 12/04/2024 UTC) where the current RDS disk storage space limit was reached in Production, the RDS storage size has been increased to `256 GB` in the Production, Test, and Pre-prod environments to accommodate the processing of larger volumes of data. The configurations for the Test and Pre-prod environments have also been updated to maintain parity with the Production environment.


mazursasha1990 (approved):
LGTM

PR #457, opened 2024-12-13
[OSDEV-1336] SLC. Implement PATCH api/v1/production-locations/{os_id}/

vladsha-dev:
[[OSDEV-1336](https://opensupplyhub.atlassian.net/browse/OSDEV-1336)]
- Introduced a new PATCH `/api/v1/production-locations/{os_id}/` endpoint based on the API v1 specification. This endpoint allows the creation of a new moderation event for updating the production location with the given details. Basically, the endpoint can be used to contribute to an existing location.
- Dynamic mapping for the new fields in the `moderation-events` index has been disabled for those that don't have an explicit mapping defined. This change helps avoid indexing conflicts, such as when a field is initially indexed with one data type (e.g., long), but later an entry with a different data type for the same field is indexed, causing the entire entry to fail indexing. After this change, fields with an explicit mapping will be indexed, while other fields will not be indexed or searchable, but will still be displayed in the document.

[OSDEV-1336]: https://opensupplyhub.atlassian.net/browse/OSDEV-1336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/release/RELEASE-NOTES.md
+++ b/doc/release/RELEASE-NOTES.md
@@ -16,7 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Code/API changes
 * [OSDEV-1336](https://opensupplyhub.atlassian.net/browse/OSDEV-1336) - Introduced a new PATCH `/api/v1/production-locations/{os_id}/` endpoint based on the API v1 specification. This endpoint allows the creation of a new moderation event for updating the production location with the given details. Basically, the endpoint can be used to contribute to an existing location.
-* [OSDEV-1336](https://opensupplyhub.atlassian.net/browse/OSDEV-1336) - Dynamic mapping for the new fields in the `moderation-events` index has been disabled for those that don't have an explicit mapping defined. This change helps avoid indexing conflicts, such as when a field is initially indexed with one data type (e.g., long), but later an entry with a different data type for the same field is indexed, causing the entire entry to fail indexing. After this change, fields with an explicit mapping will be indexed, while other fields will not be indexed or searchable, but will still be displayed in the document. The `clear_opensearch.sh.tpl` script has been edited to clear only resources related to the `moderation-events` index, so there will be no need to wait for the production-locations index to be refilled during deployment, as it will not be deleted.
+* [OSDEV-1336](https://opensupplyhub.atlassian.net/browse/OSDEV-1336) - Dynamic mapping for the new fields in the `moderation-events` index has been disabled for those that don't have an explicit mapping defined. This change helps avoid indexing conflicts, such as when a field is initially indexed with one data type (e.g., long), but later an entry with a different data type for the same field is indexed, causing the entire entry to fail indexing. After this change, fields with an explicit mapping will be indexed, while other fields will not be indexed or searchable, but will still be displayed in the document.
 
 ### Architecture/Environment changes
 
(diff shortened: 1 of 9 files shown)


src/django/api/views/v1/production_locations.py
```diff
@@ -184,3 +188,58 @@ def create(self, request):
...
             status=result.status_code
         )
+
+    @transaction.atomic
+    def partial_update(self, request, pk=None):
+        if switch_is_active('disable_list_uploading'):
+            raise ServiceUnavailableException(
+                APIV1CommonErrorMessages.MAINTENANCE_MODE
+            )
+        if not Facility.objects.filter(id=pk).exists():
```
roman-stolar, line 198: **Suggestion:** Move it ti ProductionLocationService as was implemented in https://github.com/opensupplyhub/open-supply-hub/pull/438 (ModerationEventService)

src/django/api/views/v1/production_locations.py
```diff
@@ -184,3 +188,58 @@ def create(self, request):
...
+            specific_error = APIV1LocationContributionErrorMessages \
+                .LOCATION_NOT_FOUND
+            return Response(
+                {'detail': specific_error},
+                status=status.HTTP_404_NOT_FOUND
+            )
+        if not isinstance(request.data, dict):
```
roman-stolar, line 205: **Suggestion:** Move it ti ProductionLocationService as was implemented in https://github.com/opensupplyhub/open-supply-hub/pull/438 (ModerationEventService)

vladsha-dev: If you don't mind, I would prefer to leave it as is, because it’s a simple check that I don’t think requires a special method.

roman-stolar (approved):
LGTM

src/django/api/moderation_event_actions/creation/dtos/create_moderation_event_dto.py
```diff
@@ -4,13 +4,16 @@
...
+from api.models.facility.facility import Facility
 
 
 @dataclass
 class CreateModerationEventDTO:
-    contributor_id: int
+    contributor: Contributor
     raw_data: Dict
     request_type: str
+    os: Facility = None
```
VadimKovalenkoSNF, line 16: Should it be `os_id`?

src/django/api/tests/test_location_contribution_strategy.py
```diff
@@ -70,7 +76,7 @@ def test_source_set_as_api_regardless_of_whether_passed(self, mock_get):
         self.assertNotIn('source', self.common_valid_input_data)
 
         event_dto = CreateModerationEventDTO(
-            contributor_id=self.contributor,
```
VadimKovalenkoSNF, line 73: Why did you decide to remove `_id` suffix for the contributor?

doc/release/RELEASE-NOTES.md
```diff
@@ -15,6 +15,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 #### Scheme changes
 
 ### Code/API changes
+* [OSDEV-1336](https://opensupplyhub.atlassian.net/browse/OSDEV-1336) - Introduced a new PATCH `/api/v1/production-locations/{os_id}/` endpoint based on the API v1 specification. This endpoint allows the creation of a new moderation event for updating the production location with the given details. Basically, the endpoint can be used to contribute to an existing location.
```
VadimKovalenkoSNF, line 18: This might be a strange question (I might have forgot something but I didn't find this in the ticket): since this endpoint will be triggered (mostly) from the SLC, will it implicitly set `source=SLC` to the request payload? Or this won't change `source` field at all?

VadimKovalenkoSNF (commented):
Have a few questions.

src/django/api/tests/test_location_contribution_strategy.py
vladsha-dev: Because it is a Contributor instance.

src/django/api/moderation_event_actions/creation/dtos/create_moderation_event_dto.py
vladsha-dev: No. It is Facility instance.

doc/release/RELEASE-NOTES.md
vladsha-dev: The source should be passed as `SLC`; otherwise, it will be set to `API`, which is the default behavior.

VadimKovalenkoSNF: > The source should be passed as SLC; 
Could you give me a reference in the codebase where this is happening?

vladsha-dev: Sure. Here it is ->
https://github.com/opensupplyhub/open-supply-hub/blob/main/src/django/api/serializers/v1/contribution_moderation_event_source_field_serializer.py#L12

roman-stolar (approved):

VadimKovalenkoSNF (approved):
Approved, good work!

PR #478, opened 2025-01-07
[OSDEV-1524] align salutations and signatures in automated emails

FrancescaOSH:
Updated salutations in automated emails to be consistent across all email templates. This ensures that customers have a consistent and professional experience of communication from OS Hub.


vladsha-dev (dismissed):
LGTM

PR #482, opened 2025-01-13
Remove the snapshot_identifier value for the Test environment

vladsha-dev:
The value has been removed from the `main` branch to prevent triggering a DB restore from the snapshot when it is not necessary. This value should only be set in a special branch whose purpose is to restore the DB from a snapshot, and it should be empty in the main branch to avoid triggering the restore.


Innavin369 (approved):

roninzp (approved):
LGTM

PR #490, opened 2025-01-24
[OSDEV-1515] Database. Upgrade the PostgreSQL version to 17

vladsha-dev:


roninzp (approved):
LGTM

deployment/environments/terraform-preprod.tfvars
```diff
@@ -14,8 +14,8 @@ bastion_ami = "ami-0bb3fad3c0286ebd5"
 bastion_instance_type = "t3.nano"
 
 rds_allocated_storage = "256"
-rds_engine_version = "13"
-rds_parameter_group_family = "postgres13"
+rds_engine_version = "16"
+rds_parameter_group_family = "postgres16"
```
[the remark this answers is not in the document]
vladsha-dev: > rds_allow_major_version_upgrade = true
> rds_apply_immediately = true
> 

These two variables should be set only for the environments that are currently running on Postgres 13. Since Pre-prod is always created from scratch, this means that Postgres 16 will be applied instantly, and there won't be an upgrade process like in Staging, Production, Dev, or Test environments.

deployment/environments/terraform-production.tfvars
```diff
@@ -13,8 +13,8 @@ bastion_ami = "ami-0bb3fad3c0286ebd5"
 bastion_instance_type = "t3.nano"
 
 rds_allocated_storage = "256"
-rds_engine_version = "13"
-rds_parameter_group_family = "postgres13"
+rds_engine_version = "16"
+rds_parameter_group_family = "postgres16"
```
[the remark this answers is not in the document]
vladsha-dev: For the sake of cost efficiency, it was decided not to enable multi-AZ deployment for RDS.

src/django/sqls/0163_refresh_pg_statistic_and_upgrade_postgres_extensions.sql
```diff
@@ -1,4 +1,54 @@
...
-ALTER EXTENSION pg_trgm UPDATE TO '1.5';
+DO $$
+DECLARE
+-- Variables for postgis.
+current_version_postgis TEXT;
+target_version_postgis TEXT := '3.4.2';
+
+-- Variables for pg_trgm.
+current_version_pg_trgm TEXT;
+target_version_pg_trgm TEXT := '1.5';
```
mazursasha1990, line 11: You wrote that `pg_trgm` extension has been upgraded to version 1.6 but here is version 1.5.
Should it be corrected?

src/django/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN set -ex \
     && deps=" \
     gdal-bin \
     gettext \
-    postgresql-client-13 \
+    postgresql-client-15 \
```
mazursasha1990, line 15: Is this version correct?

mazursasha1990 (changes requested):

src/django/sqls/0163_refresh_pg_statistic_and_upgrade_postgres_extensions.sql
vladsha-dev: This is the SQL script for migration 0163, where I upgraded the pg_trgm version to 1.5.
You can refer to another file for the upgrade to version 1.6 -> 0164_refresh_pg_statistic_and_upgrade_postgres_extensions_after_db_upgrade_to_postgres_16.py

src/django/Dockerfile
vladsha-dev: Yeah...
I was researching this yesterday and on Monday, and I found the following:
The [postgresql-client-16](https://packages.debian.org/search?keywords=postgresql-client-16&searchon=names&suite=all&section=all) package isn't available for the latest stable version of Debian, which is [Bookworm](https://www.debian.org/releases/), used in the Django container (see line 1 in the Dockerfile). The latest stable version available is [postgresql-client-15](https://packages.debian.org/search?keywords=postgresql-client-15&searchon=names&suite=all&section=all), which is why I used it. Generally, there shouldn't be any issues, as PostgreSQL versions are usually compatible with each other. Testing the Django app locally confirms that it works as expected.

src/django/sqls/0163_refresh_pg_statistic_and_upgrade_postgres_extensions.sql
mazursasha1990: Thank you!

src/django/Dockerfile
mazursasha1990: Thank you!

PR #491, opened 2025-01-28
[OSDEV-1515] Testing restoration from a snapshot

vladsha-dev:


vladsha-dev:
It was created for testing purposes related to [OSDEV-1515](https://opensupplyhub.atlassian.net/browse/OSDEV-1515) and is no longer necessary. Closed.

PR #501, opened 2025-02-05
Update release protocol

VadimKovalenkoSNF:
Assign responsible person for releases v.2.0 and v.2.1


vladsha-dev (approved):
LGTM!
Vadim, I’d recommend communicating the major version change with Pranali and other stakeholders to keep them informed of why we’re making this change.

Innavin369 (approved):

PR #524, opened 2025-02-18
[OSDEV-1696] Fix loader to the production location info

roman-stolar:
[OSDEV-1696](https://opensupplyhub.atlassian.net/browse/OSDEV-1696) Fix loader to the production location info

- Fixed the loader to be in the center as well as on the search page

[OSDEV-1696]: https://opensupplyhub.atlassian.net/browse/OSDEV-1696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/react/src/util/styles.js
+++ b/src/react/src/util/styles.js
@@ -1,4 +1,4 @@
-import { OARColor } from './constants';
+import { OARColor, HEADER_HEIGHT, FOOTER_HEIGHT } from './constants';
 import COLOURS from './COLOURS';
 import { multiValueBackgroundHandler } from './util';
 
@@ -1382,7 +1382,7 @@ export const productionLocationInfoStyles = theme =>
             display: 'flex',
             justifyContent: 'center',
             alignItems: 'center',
-            height: 'calc(100vh - 116px)',
+            height: `calc(100vh - ${HEADER_HEIGHT + FOOTER_HEIGHT}px)`,
         }),
     });
 
--- a/src/react/src/util/styles.js
+++ b/src/react/src/util/styles.js
@@ -1,4 +1,4 @@
-import { OARColor, HEADER_HEIGHT, FOOTER_HEIGHT } from './constants';
+import { OARColor, HEADER_HEIGHT } from './constants';
 import COLOURS from './COLOURS';
 import { multiValueBackgroundHandler } from './util';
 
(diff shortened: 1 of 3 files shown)


mazursasha1990 (changes requested):

src/react/src/util/styles.js
```diff
@@ -1378,6 +1378,12 @@ export const productionLocationInfoStyles = theme =>
         notchedOutlineStyles: Object.freeze({
             borderRadius: '0',
         }),
+        circularProgressStyles: Object.freeze({
+            display: 'flex',
+            justifyContent: 'center',
+            alignItems: 'center',
+            height: 'calc(100vh - 116px)',
```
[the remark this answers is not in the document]
roman-stolar: Got it!

src/react/src/util/styles.js
```diff
@@ -1378,6 +1378,12 @@ export const productionLocationInfoStyles = theme =>
         notchedOutlineStyles: Object.freeze({
             borderRadius: '0',
         }),
+        circularProgressStyles: Object.freeze({
+            display: 'flex',
+            justifyContent: 'center',
+            alignItems: 'center',
+            height: `calc(100vh - ${HEADER_HEIGHT + FOOTER_HEIGHT}px)`,
```
VadimKovalenkoSNF, line 1385: I tend to disagree with this approach:

1. Even though the header is fixed on all screens, the same does not apply to the footer. Please check how the footer behaves on mobile devices, for example. I would use a React `ref` to dynamically get the footer height.

2. For this specific case (a simple loader), I wouldn't over-engineer the CSS. Instead, I would simply set `height: 50vh`. This value is sufficient to render the footer properly while also being more maintainable.

VadimKovalenkoSNF (changes requested):
Request change.

src/react/src/util/styles.js
roman-stolar: Hi @VadimKovalenkoSNF, yesterday we have a conversation with @mazursasha1990, I will re-implement it in the same way as it work in SearchByNameAddressResult page.

VadimKovalenkoSNF: Ack.

VadimKovalenkoSNF (approved):
Approved.

mazursasha1990 (approved):
LGTM

PR #531, opened 2025-02-21
[OSDEV-899][Backmerge] Revert the splitting of the Django and React app.

vladsha-dev:
[[OSDEV-899](https://opensupplyhub.atlassian.net/browse/OSDEV-899)]
Backmerge of the [PR](https://github.com/opensupplyhub/open-supply-hub/pull/529).

The default redirection of 404 and 403 responses to the /index.html file in the S3 bucket, which CloudFront receives from the S3 bucket with the React app, causes 403 and 404 responses from Django API endpoints to be redirected to the same S3 bucket. This means that endpoints, such as those requiring admin access, won't return the correct JSON response for a 403 error. Instead, the HTML file will be returned with a 200 status code.

Additionally, removing this redirection rule for 403 errors will cause the React app to break, as CloudFront won't be able to locate the React paths that the client sends in the S3 bucket. Requests won't be rerouted to index.html, which is the main entry point for the React app.

For now, the splitting of Django and the React app for the AWS setup will be reverted to ensure the platform functions properly. Meanwhile, the DevOps engineer is investigating the possibility of implementing a handling function in CloudFront to properly manage the redirection in this case.

[OSDEV-899]: https://opensupplyhub.atlassian.net/browse/OSDEV-899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/deploy_to_aws.yml
+++ b/.github/workflows/deploy_to_aws.yml
@@ -122,7 +122,7 @@ jobs:
           SETTINGS_BUCKET: oshub-settings-${{ steps.get_env_name.outputs.lowercase }}
           AWS_DEFAULT_REGION: "eu-west-1"
 
-  build_and_push_react_app:
+  build_and_push_docker_image:
     needs: apply
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


VadimKovalenkoSNF (approved):
Approved.

protsack-stephan (approved):

PR #537, opened 2025-02-26
[OSDEV-1746] Implement auto-scaling for Django app instances

roninzp:
Auto-scaling ensures the optimal number of Django instances based on the current load. This helps efficiently allocate resources, reducing excessive costs while maintaining stable operation and application availability, even with fluctuating traffic levels.


vladsha-dev (changes requested):
@roninzp, could we control the memory usage so that it starts scaling once the memory reaches 80-85 percent?
I believe it is also important since some processes can take a little CPU but use a lot of memory, which could lead to the death of the ECS container.

deployment/environments/terraform-production.tfvars
vladsha-dev, line 1: Should `app_ecs_desired_count` be 10 instead of 5, to match `app_ecs_scale_target_min_capacity`?

deployment/environments/terraform-preprod.tfvars
vladsha-dev, line 1: Should `app_ecs_desired_count` be 10 instead of 5, to match `app_ecs_scale_target_min_capacity`?

deployment/environments/terraform-staging.tfvars
vladsha-dev, line 1: Should `app_ecs_desired_count` be 4 instead of 2, to match `app_ecs_scale_target_min_capacity`?

deployment/environments/terraform-test.tfvars
vladsha-dev, line 1: Should `app_ecs_desired_count` be 2 instead of 1, to match `app_ecs_scale_target_min_capacity`?

deployment/terraform/aws-ecs-service-autoscaling/outputs.tf
vladsha-dev, line 1: Do we need this empty file?

doc/release/RELEASE-NOTES.md
```diff
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Architecture/Environment changes
 * [OSDEV-1935](https://opensupplyhub.atlassian.net/browse/OSDEV-1935) - Added terraform module for creating IAM roles in production and test AWS accounts to enable integration with Vanta auditor.
 * [OSDEV-1895](https://opensupplyhub.atlassian.net/browse/OSDEV-1895) - Updated the rules to send the `csrftoken` cookie with the `HttpOnly` flag to the user. Implemented logic to retrieve the `csrftoken` upon user login and save it to `localStorage`. Added functionality to set the `csrftoken` from `localStorage` in Axios headers.
+* [OSDEV-1746](https://opensupplyhub.atlassian.net/browse/OSDEV-1746) - Implemented auto-scaling to dynamically adjust instance count based on load metrics for cost-efficient resource utilization and high availability.
```
vladsha-dev, line 27: Please, move it to version 2.4.0

deployment/terraform/task-definitions/app.json
```diff
@@ -60,6 +60,16 @@
...
+    },
+    "healthCheck": {
+        "command": [
+            "CMD-SHELL",
+            "/usr/bin/curl -f http://localhost:${app_port}/health-check/ || exit 1"
+        ],
+        "interval": 30,
+        "timeout": 5,
+        "retries": 3,
+        "startPeriod": 30
```
vladsha-dev, line 72: What is the purpose of adding that?
I thought the health check at the load balancer level triggers the same action within the ECS service: killing the container when it becomes unhealthy. Maybe it would be worth adding this configuration to that health check instead?

vladsha-dev (changes requested):

deployment/environments/terraform-production.tfvars
roninzp: They can be fixed. But autoscaler will do it itself if there is no sufficient load.
Perhaps it makes sense not to change the value of app_ecs_desired_count, so that the service will be in the maximum availability mode during the deployment.
But you can choose any other value equal to or greater than app_ecs_scale_target_max_capacity.

deployment/environments/terraform-preprod.tfvars
roninzp: They can be fixed. But autoscaler will do it itself if there is no sufficient load.
Perhaps it makes sense not to change the value of app_ecs_desired_count, so that the service will be in the maximum availability mode during the deployment.
But you can choose any other value equal to or greater than app_ecs_scale_target_max_capacity.

deployment/environments/terraform-staging.tfvars
roninzp: They can be fixed. But autoscaler will do it itself if there is no sufficient load.
Perhaps it makes sense not to change the value of app_ecs_desired_count, so that the service will be in the maximum availability mode during the deployment.
But you can choose any other value equal to or greater than app_ecs_scale_target_max_capacity.

deployment/environments/terraform-test.tfvars
roninzp: They can be fixed. But autoscaler will do it itself if there is no sufficient load.
Perhaps it makes sense not to change the value of app_ecs_desired_count, so that the service will be in the maximum availability mode during the deployment.
But you can choose any other value equal to or greater than app_ecs_scale_target_max_capacity.

deployment/terraform/aws-ecs-service-autoscaling/outputs.tf
roninzp: removed

deployment/terraform/task-definitions/app.json
roninzp: Adding a healthCheck to the task definition at the container level allows ECS to monitor the container’s health independently of the load balancer. This is particularly useful during the startup phase, when the container may not yet be ready to receive traffic but is already considered "running."

If the container fails the health check (e.g., curl does not return a 2xx response), ECS will restart the container regardless of the load balancer’s health check.

The load balancer’s health check (e.g., ALB) operates externally, checking whether the service is accessible from outside. It can deregister a target from the load balancer but does not restart the container. Therefore, both levels of health checks — container-level and load-balancer-level — complement each other and are not interchangeable.

Additionally, implementing a container-level health check helped us resolve an issue where ECS continuously launched new containers and terminated previous ones during autoscaling tests.

doc/release/RELEASE-NOTES.md
roninzp: done

deployment/terraform/task-definitions/app.json
vladsha-dev: Ok, thank you.

src/django/oar/settings.py
```diff
@@ -67,6 +67,7 @@
     '.openapparel.org',
     '.opensupplyhub.org',
     '.os-hub.net',
+    'localhost',
```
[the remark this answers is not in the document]
roninzp: Yes, it safe. The service is accessible by this name only from inside the container.

doc/release/RELEASE-NOTES.md
```diff
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Architecture/Environment changes
 * [OSDEV-1960](https://opensupplyhub.atlassian.net/browse/OSDEV-1960) - Disabled deletion protection and final snapshot creation for the RDS instance when it is deleted in the pre-prod environment.
+* [OSDEV-1746](https://opensupplyhub.atlassian.net/browse/OSDEV-1746) - Implemented auto-scaling to dynamically adjust instance count based on load metrics for cost-efficient resource utilization and high availability.
```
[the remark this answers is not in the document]
roninzp: done

src/django/oar/settings.py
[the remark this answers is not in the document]
vladsha-dev: Okay, thanks. Please remove the duplicate entry of `localhost` in the array on line 74 of this file. It simply adds the same value again to the `ALLOWED_HOSTS` array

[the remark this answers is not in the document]
roninzp: done

vladsha-dev (dismissed):
LGTM

vladsha-dev (approved):
LGTM

mazursasha1990 (approved):
LGTM

PR #538, opened 2025-02-26
[OSDEV-1782] Create a pop-up window to enter a text message for the rejection action of contribution.

mazursasha1990:
[OSDEV-1782](https://opensupplyhub.atlassian.net/browse/OSDEV-1782) - [SLC] Moderation Queue: Record rejection options & enable pop-up window to enter text.

In this PR, the following changes were made:

- New fields action_reason_text_cleaned and action_reason_text_raw were added to the api_moderationevent table to store text messages received when a moderator takes an action on a moderation event.
- Additional validation was added for the fields action_reason_text_cleaned and action_reason_text_raw when using the PATCH api/v1/moderation-events/{moderation_id} endpoint. These fields are required in the request body when the status field is set to 'REJECTED'. The minimum length for the values of these fields is 30 characters.
- A confirmation dialog window was added that appears when a user tries to reject a moderation event. The dialog includes a WYSIWYG text editor where entering a message of at least 30 characters is required to confirm the rejection. If a user does not enter the required number of characters, the 'Reject' button is disabled, and a tooltip with a clear message appears when the mouse hovers over it."
- Was added sanitization on the server side by using the `Django-Bleach` library for the HTML content that is stored in the `action_reason_text_raw` field. The bleach filter was applied to the `action_reason_text_raw` value in the `slc_contribution_rejected_body.html` template.

Screenshot from 2025-03-04 22-21-09
Screenshot from 2025-03-04 19-51-02

[OSDEV-1782]: https://opensupplyhub.atlassian.net/browse/OSDEV-1782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/react/src/actions/dashboardContributionRecord.js
```diff
@@ -130,15 +130,24 @@ export function fetchPotentialMatches(data) {
...
         dispatch(startUpdateSingleModerationEvent());
 
         return apiRequest
-            .patch(makeModerationEventRecordURL(moderationID), { status })
+            .patch(makeModerationEventRecordURL(moderationID), {
+                status,
+                action_reason_text_cleaned: textCleaned,
+                action_reason_text_raw: textRaw,
+            })
             .then(({ data }) => {
```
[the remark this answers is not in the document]
mazursasha1990: I don't see the described issue.

src/django/api/models/moderation_event.py
```diff
@@ -147,6 +147,16 @@ class ActionType(models.TextChoices):
...
 
+    action_reason_text_cleaned = models.TextField(
+        blank=True,
+        help_text='Cleaned version of the action reason text.'
+    )
+
+    action_reason_text_raw = models.TextField(
+        blank=True,
+        help_text='Raw version of the action reason text.'
+    )
```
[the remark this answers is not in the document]
mazursasha1990: I don't want to add a maximum length constraint for the `action_reason_text_cleaned` and `action_reason_text_raw` fields.

src/django/api/templates/mail/slc_contribution_rejected_body.html
```diff
@@ -22,7 +22,7 @@
           <b>Review the feedback below:</b>
         </p>
         <ul>
-          <li>{{ action_reason_text_raw }}</li>
+          <li>{{ action_reason_text_raw | safe }}</li>
```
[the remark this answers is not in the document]
mazursasha1990: Applied Django-Bleach library.

src/django/api/serializers/v1/moderation_event_update_serializer.py
```diff
@@ -78,12 +84,58 @@ def __validate_status(self, value):
             })
         return value
 
+    def validate(self, data):
+        status = data.get('status')
+        if status == ModerationEvent.Status.REJECTED:
```
protsack-stephan, line 89: Can we change this to be a reverse if statement something like:

```python
if  data.get('status') != ModerationEvent.Status.REJECTED:
  return data
```

src/django/api/templates/mail/slc_contribution_rejected_body.html
```diff
@@ -21,9 +22,7 @@
         <p>
           <b>Review the feedback below:</b>
         </p>
-        <ul>
-          <li>{{ action_reason_text_raw }}</li>
-        </ul>
+        {{ action_reason_text_raw|bleach }}
```
protsack-stephan, line 25: Should this be indented like this if possible? Just makes it a bit more readable.

```
{{ action_reason_text_raw | bleach }}
```

protsack-stephan (approved):
Few small comments. Great work!

src/django/api/models/moderation_event.py
```diff
@@ -147,6 +148,16 @@ class ActionType(models.TextChoices):
         )
     )
 
+    action_reason_text_cleaned = models.TextField(
```
VadimKovalenkoSNF, line 151: I suggest to add `max_length` attribute for fields introduced in this PR.

doc/release/RELEASE-NOTES.md
```diff
@@ -13,13 +13,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 * *Describe high-level database changes.*
```
VadimKovalenkoSNF, line 13: Pls, add here similar text that you've added in Migrations section.

src/django/api/templates/mail/slc_contribution_rejected_body.html
mazursasha1990: Done.

src/react/src/components/Dashboard/ConfirmActionDialog.jsx
```diff
@@ -0,0 +1,146 @@
+import React, { useState, useMemo } from 'react';
```
VadimKovalenkoSNF, line 1: Maybe it is better to rename this component to `RejectActionDialog.jsx` or `RejectModerationEventDialog.jsx`?

VadimKovalenkoSNF (changes requested):

src/django/api/serializers/v1/moderation_event_update_serializer.py
mazursasha1990: Done.

src/django/api/models/moderation_event.py
mazursasha1990: The TextField is designed to hold large amounts of text and doesn’t enforce a maximum length at the database level.
https://docs.djangoproject.com/en/5.1/ref/models/fields/#textfield

So, I believe it's not necessary to add a max_length attribute for these fields.

src/react/src/components/Dashboard/ConfirmActionDialog.jsx
mazursasha1990: Initially, I thought that this component would be common for all moderation event actions, therefore I called it `ConfirmActionDialog`.
But in the current stage, I agree that it would be better to rename it because it's unknown whether we will implement such a dialog window for other actions or not.

Renamed to the `RejectModerationEventDialog`.
Thank you.

src/django/api/serializers/v1/moderation_event_update_serializer.py
```diff
@@ -78,12 +84,61 @@ def __validate_status(self, value):
...
+                "detail": (
+                    "This field is required when rejecting a moderation "
+                    "event."
+                )
+            })
+        elif len(raw) < self.MIN_REASON_TEXT_LENGTH:
+            errors.append({
+                "field": "action_reason_text_raw",
+                "detail": "This field must be at least 30 characters."
+            })
```
protsack-stephan, line 123: This is just a comment for the future, no need to change it in this PR. I would suggest assigning the variable right before using it instead of at the top of the file, that means assigning the `raw` value right before this if else statement, like this:

```python
raw = data.get('action_reason_text_raw')

if not raw:
            errors.append({
                "field": "action_reason_text_raw",
                "detail": (
                    "This field is required when rejecting a moderation "
                    "event."
                )
            })
        elif len(raw) < self.MIN_REASON_TEXT_LENGTH:
            errors.append({
                "field": "action_reason_text_raw",
                "detail": "This field must be at least 30 characters."
            })
```

protsack-stephan (approved):

src/django/api/models/moderation_event.py
VadimKovalenkoSNF: Ack.

doc/release/RELEASE-NOTES.md
vladsha-dev: > Pls, add here similar text 

@VadimKovalenkoSNF, this section under the "Database Changes" title was added to include information that is not related to migrations and the DB schema, such as the Postgres version upgrade—anything beyond just migrations and DB schema. So, I wouldn’t add this text here either.

vladsha-dev: Since almost all the work around the DB is related to schema and migrations, everything else would definitely be more general and high-level. @VadimKovalenkoSNF, what do you think?

VadimKovalenkoSNF: Then I suggest to remove `* *Describe high-level database changes.*` text and combine sections like this. We did like that in some of previous release notes (see release 1.23.0). 

```
### Database changes
#### Migrations:
<you migrations go there>
```

mazursasha1990: I believe it is worth clearing the release notes in the last PR before the Release (or creating a special PR for these purposes).

VadimKovalenkoSNF: Makes sense, but this will require additional negotiations between devs to decide who is last.

vladsha-dev: For this release, yes, it can be removed. I would just keep it in the release notes template file.

VadimKovalenkoSNF: I mark this conversation as resolved.

PR #557, opened 2025-03-11
Wait for snapshot get status - available

roninzp:
In this pull request, we fix the snapshot status checking procedure. This will prevent a crash when trying to restore a database from an inaccessible snapshot.


VadimKovalenkoSNF (approved):
LGTM

vladsha-dev (approved):

PR #558, opened 2025-03-12
[OSDEV-1428] The Test, Development, and Pre-prod environments are shutting down for a while during deployment.

roninzp:


deployment/terraform/container_service.tf
```diff
@@ -254,6 +254,7 @@ resource "aws_ecs_task_definition" "app" {
   requires_compatibilities = ["FARGATE"]
   cpu                      = var.app_fargate_cpu
   memory                   = var.app_fargate_memory
+  skip_destroy              = true
```
VadimKovalenkoSNF, line 257: There is redundant space, `= true` is slightly misaligned to the right.

VadimKovalenkoSNF (changes requested):
Could you describe which environment changes prevent shutting down? You can leave this in ticket description + link the ticket there as well.

vladsha-dev:
The PR has been closed due to its irrelevance.

PR #580, opened 2025-04-03
[OSDEV-1895] [RBA Pen Test] Address pen test results of Cookies without HTTP-only

roman-stolar:
[OSDEV-1895](https://opensupplyhub.atlassian.net/browse/OSDEV-1895) **[RBA Pen Test] Address pen test results of Cookies without HTTP-only**

- Updated the rules to send the `csrftoken` cookie with the `HttpOnly` flag to the user. 
- Implemented logic to retrieve the `csrftoken` upon user login and save it to `localStorage`. 
- Added functionality to set the `csrftoken` from `localStorage` in Axios headers.

### DIFF
--- a/src/django/api/views/auth/login_to_oar_client.py
+++ b/src/django/api/views/auth/login_to_oar_client.py
@@ -3,6 +3,7 @@ from rest_framework.exceptions import AuthenticationFailed
 from rest_framework.response import Response
 from rest_auth.views import LoginView
 from django.contrib.auth import authenticate, login
+from django.middleware.csrf import get_token
 
 from ...serializers.user.user_serializer import UserSerializer
 
@@ -36,8 +37,12 @@ class LoginToOARClient(LoginView):
                 'Your account is not verified. '
                 'Check your email for a confirmation link.'
             )
+        
+        serialized_data = UserSerializer(user).data
+        csrf_token = get_token(request)
+        serialized_data['csrfToken'] = csrf_token
 
-        return Response(UserSerializer(user).data)
+        return Response(serialized_data)
 
     def get(self, request, *args, **kwargs):
         if not request.user.is_active:
--- a/src/django/oar/settings.py
+++ b/src/django/oar/settings.py
@@ -152,6 +152,9 @@ AUTH_USER_MODEL = 'api.User'
 CSRF_COOKIE_SECURE = True
 SESSION_COOKIE_SECURE = True
 
+# https://docs.djangoproject.com/en/5.1/ref/settings/#csrf-cookie-httponly
+CSRF_COOKIE_HTTPONLY = True
+
 REST_AUTH_SERIALIZERS = {
     'USER_DETAILS_SERIALIZER': 'api.serializers.UserSerializer',
     'PASSWORD_RESET_SERIALIZER': 'api.serializers.UserPasswordResetSerializer',
--- a/src/react/src/util/apiRequest.js
+++ b/src/react/src/util/apiRequest.js
@@ -11,12 +11,18 @@ const apiRequest = axios.create({
 
 // Not related to CSRF protection, but this is the appropriate place to set the
 // client key header used to authenticate anonymous API requests.
-apiRequest.interceptors.request.use(config =>
-    Object.assign({}, config, {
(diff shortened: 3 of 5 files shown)


src/django/oar/settings.py
```diff
@@ -152,6 +152,9 @@
 CSRF_COOKIE_SECURE = True
 SESSION_COOKIE_SECURE = True
 
+# https://docs.djangoproject.com/en/5.1/ref/settings/#csrf-cookie-httponly
+CSRF_COOKIE_HTTPONLY = True
```
VadimKovalenkoSNF, line 156: I have concerns about using local storage in terms of saving CSRF token. If an attacker manages to run JavaScript in your user's browser (e.g., via an XSS attack), they can read localStorage and steal the CSRF token, defeating its purpose.

What about to try set `CSRF_COOKIE_HTTPONLY: False`, but:
`CSRF_COOKIE_SECURE`: True
`CSRF_COOKIE_SAMESITE`: Strict or Lax

VadimKovalenkoSNF (changes requested):

src/django/oar/settings.py
roman-stolar: @VadimKovalenkoSNF I understand your concern, that why we waiting for response from PenTest team about our approach. In general we change one issue to another, the only different that after this changes `csrf token` could be access only by XSS attack.

**Also if we setup:**
`CSRF_COOKIE_HTTPONLY: False`
`CSRF_COOKIE_SECURE: True`
`CSRF_COOKIE_SAMESITE: Strict or Lax`

These settings don’t control JS access so it won't resolve an issue.

VadimKovalenkoSNF: Ack.

VadimKovalenkoSNF (dismissed):
LGTM

mazursasha1990 (dismissed):
LGTM

mazursasha1990 (approved):
LGTM

src/django/api/views/auth/login_to_oar_client.py
```diff
@@ -37,7 +37,14 @@ def post(self, request, *args, **kwargs):
                 'Check your email for a confirmation link.'
             )
 
-        return Response(UserSerializer(user).data)
+        # Adding the CSRF token to the serialized data response to store it
+        # in local storage for future use and include it in header requests,
+        # as access to cookies has been restricted by the HttpOnly flag.
```
vladsha-dev, line 42: Did you mean "request headers" instead of "header requests" ?

src/react/src/util/apiRequest.js
```diff
@@ -9,14 +11,21 @@ const apiRequest = axios.create({
...
+// Appropriate place to set the client key header 'X-OAR-Client-Key'
+// used to authenticate anonymous API requests.
+// Set 'X-CSRFToken' for CSRF protection when
+// making unsafe HTTP requests.
+apiRequest.interceptors.request.use(config => {
+    const headers = {
+        ...config.headers,
+        'X-OAR-Client-Key': window.ENVIRONMENT.OAR_CLIENT_KEY,
+        'X-CSRFToken':
+            config.headers['X-CSRFToken'] ||
```
vladsha-dev, line 23: What is the purpose of getting the value from the header if it is already in local storage?

vladsha-dev (changes requested):

src/django/api/views/auth/login_to_oar_client.py
roman-stolar: Will update

src/react/src/util/apiRequest.js
roman-stolar: We need this for old/previous session + to assign X-CSRFToken from local storage if it not exist yet.

vladsha-dev: Thank you for your explanation!

roman-stolar:
@vladsha-dev Just tested locally with scenarios you presented today on call, all things work properly and I can successfully logout even if I login on previous implementation of csrftoken.

mazursasha1990 (approved):
LGTM

vladsha-dev (approved):
LGTM

PR #590, opened 2025-04-15
Fix SQL script for DB anonymization

roninzp:


roninzp:
This fix is ​​not needed.

PR #598, opened 2025-04-18
[OSDEV-1927] Return additional identifiers in the production locations API response.

Innavin369:
[OSDEV-1927](https://opensupplyhub.atlassian.net/browse/OSDEV-1927) Return additional identifiers in the production locations API response.
Screenshot 2025-04-18 at 19 04 31
Screenshot 2025-04-18 at 19 04 19
Screenshot 2025-04-18 at 19 04 07
Screenshot 2025-04-18 at 19 03 42

PR #601, opened 2025-04-21
[OSDEV-1926] Allow submission of additional identifiers through the API

mazursasha1990:
[OSDEV-1926](https://opensupplyhub.atlassian.net/browse/OSDEV-1926) - [RBA Data Model] Allow submission of additional identifiers through the API.

The current PR introduces support for submitting additional identifiers when uploading a new production location or modifying an existing one. Additional identifiers can now be added via the API (POST `api/facilities/`, POST `api/v1/production-locations/`, PATCH `api/v1/production-locations/{os_id}/`) or through list uploads. The system currently supports three types of identifiers: DUNS (Data Universal Numbering System), LEI (Legal Entity Identifier), and RBA Online ID. The provided identifiers are stored as standalone fields in the `api_extendedfields` table.

[OSDEV-1926]: https://opensupplyhub.atlassian.net/browse/OSDEV-1926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/views/fields/create_nonstandard_fields.py
```diff
@@ -20,6 +20,7 @@ def create_nonstandard_fields(
         'address',
         'lat',
         'lng',
+        'additional_identifiers',
```
vladsha-dev, line 23: Why is the key named `additional_identifiers`? I thought it should be the actual field names like `lei_id`, `duns_id`, and `rba_id`, since you’re saving them separately as standalone fields, not under `additional_identifiers`, in the `row_additional_ids_serializer.py` file, line 93.  

Do we need these as standard fields? I’d check with Pranali, because I believe this could be a good option for Embedded Map customers to display DUNS, LEI, or RBA IDs on the production location profile page.

Just FYI, I tested a case where these identifiers were uploaded as non-standard fields, and everything worked without issues. They don’t appear on the location profile page by default, but you can enable them by checking the relevant boxes in the Embed Config tab, and they will eventually show up on the location profile page in Embed Map mode.

vladsha-dev (changes requested):

doc/release/RELEASE-NOTES.md
```diff
@@ -19,7 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 * *Describe schema changes here.*
 
 ### Code/API changes
-* *Describe code/API changes here.*
+* [OSDEV-1926](https://opensupplyhub.atlassian.net/browse/OSDEV-1926) - Introduced support for submitting additional identifiers when uploading a new production location or modifying an existing one. The system now supports three types of identifiers: DUNS (Data Universal Numbering System), LEI (Legal Entity Identifier), and RBA Online ID.
```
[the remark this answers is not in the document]
mazursasha1990: Modified the text.

src/django/api/views/fields/create_nonstandard_fields.py
mazursasha1990: You're right — `lei_id`, `duns_id`, and `rba_id` are passed separately as standalone fields within the `fields` object and are then saved into the `api_extendedfields` table. For creating extended fields, we use the `create_extendedfields_for_single_item` method, passing the data from `fields` as an argument.

However, for creating non-standard fields, the system uses the `create_nonstandard_fields` method, which takes `header_row_keys` as an argument. In this case, an `additional_identifiers` field is created in the `api_nonstandardfield` table. In the Embed Map, it would be associated with the object like this:
```json
{
    "duns_id": "150483823",
    "lei_id": "529900T83BM49AURSD55",
    "rba_id": "RBA-12345678"
}
```
However, the system doesn't support this type of value, and an error would occur. 
This is why I omitted adding `additional_identifiers` to the non-standard fields.

In conclusion, our system currently doesn't support extracting extended fields from an object passed in the request and simultaneously adding those same fields as non-standard fields without significant refactoring. 
As an alternative, we could pass the additional IDs (`lei_id`, `duns_id`, and `rba_id`) separately in the request body. However, the main requirement of this ticket was to allow passing these IDs inside the `additional_identifiers` object.

Innavin369: I added them separatly but from `additional_identifiers` object. Can you take a look at [my PR](https://github.com/opensupplyhub/open-supply-hub/pull/598/files)?

mazursasha1990: Thanks @Innavin369.
No matter how you add data to the extended fields, the root cause is how you upload additional ids data.

Innavin369: I believe we upload them wrapped in additional_identifiers.

mazursasha1990: Yes, and then our system will create an `additional_identifiers` entry in the `api_nonstandardfield` table.

src/django/contricleaner/lib/serializers/row_serializers/row_additional_ids_serializer.py
```diff
@@ -0,0 +1,79 @@
+import re
+from contricleaner.constants import AdditionalIDs
+from contricleaner.lib.serializers.row_serializers.row_serializer import (
+    RowSerializer,
+)
+
+
+class RowAdditionalIdsSerializer(RowSerializer):
+    __DUNS_RE = re.compile(r'^\d{9}$')
```
VadimKovalenkoSNF, line 9: What about validation of hyphens for DUNS ids?

> A DUNS number is sometimes formatted with embedded dashes to promote readability, such as "15-048-3782"

from https://en.wikipedia.org/wiki/Data_Universal_Numbering_System

mazursasha1990: Thanks @VadimKovalenkoSNF.

But one of the further sentences is:
> The dashes are not part of D&B's official definition of the DUNS number.

I highlighted this point in one of the conversations with Stephan, and we decided not to accept dashes in the DUNS values at this stage.

cc @protsack-stephan

VadimKovalenkoSNF: Ack.

roman-stolar (approved):
Looks good! Approved.

vladsha-dev (approved):
LGTM

VadimKovalenkoSNF (approved):
Approved.

PR #603, opened 2025-04-24
[OSDEV-1886] Disable script crash when getting exit code 254 from aws cli when deleting Lambda@Edge function

roninzp:
Adding the `set +e` command is necessary to ensure that the script does not terminate on an error, allowing us to handle the exit code of the Lambda function deletion command manually. This is especially useful in cases where the function might already be deleted or doesn't exist — the `aws lambda delete-function` command would return an error, but with `set +e`, we can catch that error and decide how to proceed.


vladsha-dev (approved):
LGTM

PR #625, opened 2025-05-13
[OSDEV-1981] GET v1/production-locations/{os_id}/ endpoint doesn't return additional identifiers "duns_id", "lei_id", "rba_id"

roman-stolar:
[OSDEV-1981](https://opensupplyhub.atlassian.net/browse/OSDEV-1981) **GET v1/production-locations/{os_id}/ endpoint doesn't return additional identifiers "duns_id", "lei_id", "rba_id"**

- Fixed an issue where the `updated_at` field in the `api_facility` table was not modified when related dependency data changed, resulting in outdated or invalid data being stored in `OpenSearch`.
- Updated unit test case to cover changes with `api_facility` field `updated_at`.

[OSDEV-1981]: https://opensupplyhub.atlassian.net/browse/OSDEV-1981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/django/api/views/v1/moderation_events.py
+++ b/src/django/api/views/v1/moderation_events.py
@@ -217,6 +217,4 @@ class ModerationEvents(ViewSet):
         if event.source == ModerationEvent.Source.SLC:
             send_slc_contribution_approval_email(request, event, item)
 
-        Facility.update_facility_updated_at_field(item.facility_id)
-
         return Response({"os_id": item.facility_id}, status=status.HTTP_200_OK)
--- a/src/django/api/views/v1/moderation_events.py
+++ b/src/django/api/views/v1/moderation_events.py
@@ -16,6 +16,7 @@ from api.constants import (
     APIV1ModerationEventErrorMessages
 )
 from api.models.moderation_event import ModerationEvent
+from api.models.facility.facility import Facility
 from api.moderation_event_actions.approval.add_production_location \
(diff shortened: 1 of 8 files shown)


src/django/api/views/v1/moderation_events.py
```diff
@@ -216,4 +217,6 @@ def update_production_location(self, request, pk=None, os_id=None):
         if event.source == ModerationEvent.Source.SLC:
             send_slc_contribution_approval_email(request, event, item)
 
+        Facility.update_facility_updated_at_field(item.facility_id)
+
```
vladsha-dev, line 221: Could you please move it inside the `EventApprovalTemplate` class?
I would suggest making it a hook method - see `_create_new_facility` method as an example - and implement it only in `UpdateProductionLocation`, since it’s only needed for the update-type action.

vladsha-dev (changes requested):

src/django/api/views/v1/moderation_events.py
roman-stolar: Done.

vladsha-dev (approved):
LGTM

Innavin369 (approved):
LGTM

PR #631, opened 2025-05-20
[OSDEV-1939] SLC. Double click on Submit\Update button causes duplicated moderation events

roman-stolar:
[OSDEV-1939](https://opensupplyhub.atlassian.net/browse/OSDEV-1939) **SLC. Double click on Submit\Update button causes duplicated moderation events**

- Disable the submit button during SLC contribution submission or update to prevent duplicate requests
- Implemented a basic throttle class to prevent duplicate data for `POST` and `PATCH` requests on the `production-location` endpoints

[OSDEV-1939]: https://opensupplyhub.atlassian.net/browse/OSDEV-1939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/throttles.py
```diff
@@ -33,3 +43,33 @@ class SustainedRateThrottle(UserCustomRateThrottle):
 class DataUploadThrottle(UserCustomRateThrottle):
     scope = 'data_upload'
     model_rate_field = 'data_upload_rate'
+
+
+class DuplicateThrottle(BaseThrottle):
```
VadimKovalenkoSNF, line 48: What about to use this solution istead:

```
class DuplicateThrottle(BaseThrottle):
    cache = caches['api_throttling']
    MAX_REQUEST_SIZE = 1024 * 1024  # 1MB limit

    def _serialize_data(self, data):
        """Safely serialize request data to a consistent string format."""
        try:
            return json.dumps(data, sort_keys=True)
        except (TypeError, ValueError):
            # Fallback to string representation if JSON serialization fails
            return str(data)

    def allow_request(self, request, view):
        if request.method not in ["POST", "PATCH"]:
            return True

        if not request.user.is_authenticated:
            return False

        # Check if request has data
        if not request.data:
            return True

        # Check request size
        data_str = self._serialize_data(request.data)
        if len(data_str.encode()) > self.MAX_REQUEST_SIZE:
            raise Throttled(
                detail="Request data too large. Maximum size is 1MB."
            )

        data_hash = hashlib.sha256(data_str.encode()).hexdigest()
        cache_key = f"duplicate:{request.user.id}:{data_hash}"

        if self.cache.get(cache_key):
            raise Throttled(
                detail="Duplicate request submitted, please try again later."
            )

        self.cache.set(
            cache_key,
            timezone.now(),
            timeout=DUPLICATE_THROTTLE_TIMEOUT
        )
        return True
```
It provides:
1. Better data serialization (`_serialize_data` method that uses json.dumps with sort_keys=True for consistent serialization).
2. Request Validation (Added explicit check for empty request data, added a 1MB size limit on request data to prevent memory issues. Returns early if the request has no data)
3. Cache Backend Consistency (Now uses the same `api_throttling` cache backend as other throttles, Uses `self.cache` instead of the global cache object)

VadimKovalenkoSNF (changes requested):
Suggested few improvements for `DuplicateThrottle` class.

mazursasha1990 (approved):
LGTM

VadimKovalenkoSNF (approved):
Approved.

PR #642, opened 2025-06-03
[OSDEV-1865] NEW  Data Download Limits: 5,000 records annually per user.

Innavin369:
[OSDEV-1865](https://opensupplyhub.atlassian.net/browse/OSDEV-1865) Data Download Limits: 5,000 records annually per user.

 - added FacilityDownloadLimit model,
 - added 0172_add_facility_download_limit migration, creating a new database table named FacilityDownloadLimit (only regular free users will be saved there). Includes the following fields:
uuid: Unique identifier for the facility download limit record (UUIDField).
user_id: A user field associates each FacilityDownloadLimit entry with a single user.
updated_at: Stores the timestamp of the last free records update (DateTimeField).
free_download_records: The number of facilities the user can download per calendar year for free. (PositiveIntegerField, default=5,000)
paid_download_records: The number of paid facilities that the user can download. (PositiveIntegerField, default=0)
purchase_date: The date when paid facility records were purchased (blank=true).

**` For new users, the updated_at value will be the current timestamp. For existing users, it will be set to 12/07/2025 (the initial release date). Please, change it if the feature is released later.`**

**Regular users** are restricted by FACILITIES_DOWNLOAD_LIMIT = 5,000
**API users** are restricted by FACILITIES_DOWNLOAD_LIMIT = 5,000 on the UI and have no limits on the BE side
**Anonymous users** are redirected to the auth page. 

Implementation Details:

- The allowed_records_number field was added to userPropType, UserSerializer, and is now exposed in the UI to manage the Download button state.

- allowed_records_number is calculated as the sum of available free and paid download records.

- The system prioritizes consuming free records before using any available paid records.

 - added **update_expired_download_limits** command
**_How the script works:_**
The script calculates a date representing one calendar year ago from the current time, accounting for leap years. It then identifies all records where the updated_at timestamp is earlier than this cutoff. For those expired records, it resets the free_download_records to the default limit (e.g., 5,000) and updates the updated_at timestamp to the current time.

A similar process is applied to the purchase_date field. The script filters for records older than one year but excludes those where paid_download_records is already zero. For the resulting records, it resets the paid_download_records value to zero.

 * Added tests:
 -  **For UserSerializerTest:**
    test_get_default_allowed_records_number
    test_get_custom_allowed_records_number
 - **For FacilitiesDownloadViewSetTest**
    test_new_user_has_current_date_in_updated_at
    test_old_user_has_release_date_in_updated_at
 -  **For FacilityDownloadLimitModelTest**
    test_register_download
 -  **UI DownloadFacilitiesButton tests**
 - **FacilityDownloadLimitManagerTest**

[OSDEV-1865]: https://opensupplyhub.atlassian.net/browse/OSDEV-1865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/django/api/services/facilities_download_service.py
+++ b/src/django/api/services/facilities_download_service.py
@@ -83,8 +83,8 @@ class FacilitiesDownloadService:
                 limit.paid_download_records
             )
             raise ValidationError(
-                f'Downloads are only allowed for results containing '
-                f'{records_used} facilities or fewer.'
+                f'Downloads are supported only for searches '
+                f'resulting in {records_used} facilities or less.'
             )
 
     @staticmethod


src/django/api/facilities_download_view_set.py
```diff
@@ -67,13 +66,39 @@ def list(self, request):
...
+                request.user,
+                initial_release_date
+            )
+
+        current_page = int(request.query_params.get("page", 1))
+
+        if (
+            current_page == 1 and
+            facility_download_limit
+            and (facility_download_limit.free_download_records + facility_download_limit.paid_download_records) == 0  # noqa: E501
```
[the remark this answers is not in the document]
Innavin369: I can't calculate this earlier because in this condition we have a check on the existing facility_download_limit.

src/react/src/components/DownloadMenu.jsx
```diff
@@ -0,0 +1,31 @@
...
+DownloadMenu.propTypes = {
+    anchorEl: oneOfType([
+        instanceOf(typeof Element !== 'undefined' ? Element : Object),
+        oneOf([null]),
+    ]),
+    onClose: func.isRequired,
+    onSelectFormat: func.isRequired,
+};
+
+export default DownloadMenu;
```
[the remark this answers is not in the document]
Innavin369: done.

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,91 @@
...
+            else:
+                remaining_records = records_to_subtract - self.free_download_records  # noqa: E501
+                self.free_download_records = 0
+                self.paid_download_records -= remaining_records
+            self.save()
+
+    @staticmethod
+    def get_or_create_user_download_limit(
+        user,
+        custom_date
```
roman-stolar, line 78: There should be default value like -> **custom_date = None** for case when we didn't pass this like in 
**SubmitNewUserForm**.
`FacilityDownloadLimit.get_or_create_user_download_limit(user)`

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,91 @@
...
+    def register_download(self, records_to_subtract):
+        with transaction.atomic():
+            self.refresh_from_db()
+            if self.free_download_records >= records_to_subtract:
+                self.free_download_records = self.free_download_records - records_to_subtract  # noqa: E501
+            else:
+                remaining_records = records_to_subtract - self.free_download_records  # noqa: E501
+                self.free_download_records = 0
+                self.paid_download_records -= remaining_records
+            self.save()
```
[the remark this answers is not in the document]
Innavin369: The check exists before the calculation. It is not possible to receive negative numbers.

src/django/api/models/facility_download_limit_manager.py
```diff
@@ -0,0 +1,26 @@
...
+        expired_free_limits.update(
+            free_download_records=FacilitiesDownloadSettings.FACILITIES_DOWNLOAD_LIMIT,  # noqa: E501
+            updated_at=now,
+        )
+
+        # Update paid limits in bulk
+        expired_paid_limits = self.filter(
+            purchase_date__lt=one_year_ago_with_leap
+        ).exclude(paid_download_records=0)
+        expired_paid_limits.update(paid_download_records=0)
```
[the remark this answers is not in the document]
Innavin369: We want to keep the old date.

src/django/api/tests/facility_download_limit_manager_test.py
```diff
@@ -0,0 +1,109 @@
...
+        self.assertEqual(
+            expired_free.free_download_records,
+            FacilitiesDownloadSettings.FACILITIES_DOWNLOAD_LIMIT,
+        )
+        self.assertTrue(expired_free.updated_at > one_year_ago)
+
+        # Assert fresh free limits unchanged
+        self.assertEqual(fresh_free.free_download_records, 100)
+        self.assertEqual(fresh_free.updated_at, fresh_free.updated_at)
+
```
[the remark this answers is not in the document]
Innavin369: done

VadimKovalenkoSNF (changes requested):
@Innavin369 I noticed other files with disabled flake8, pls refactor them the right way:

1. `src/django/api/models/facility_download_limit_manager.py`
2. `src/django/api/models/facility_download_limit.py`

src/django/api/migrations/0171_added_origin_source_field.py
vladsha-dev, line 1: I believe this file should be edited, as it doesn’t seem related to the changes in this task.
You may need to rebase the `features/OSDEV-1865-1879-1919-data-downloads-limits` branch on top of `main`, and then rebase this branch on top of that updated `features/OSDEV-1865-1879-1919-data-downloads-limits` branch.

vladsha-dev (changes requested):

vladsha-dev (changes requested):

src/django/api/tests/facility_download_limit_manager_test.py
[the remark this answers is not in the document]
Innavin369: Created [the PR ](https://github.com/opensupplyhub/open-supply-hub/pull/663)with a fix.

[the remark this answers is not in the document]
vladsha-dev: @Innavin369
Not all files in the `tests` folder are actual test cases.
Some of them are helper classes, like FacilityAPITestCaseBase, which doesn’t contain tests but only helper methods or a setUp method used to initialize tests in child classes. Please double-check.
So [the PR](https://github.com/opensupplyhub/open-supply-hub/pull/663) is irrelevant.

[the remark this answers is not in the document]
Innavin369: Closed PR. Fixed my test file name.

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @staticmethod
+    def get_or_create_user_download_limit(
+        user,
+        custom_date=None
+    ) -> Optional["FacilityDownloadLimit"]:
+        is_api_user = not user.is_anonymous and user.has_groups
+
+        if is_api_user or user.is_anonymous:
+            # if user is an API user we don't want to impose limits
```
vladsha-dev, line 87: Could you please use consistent formatting for new comments? For instance, in facility_download_limit_manager.py, your comment starts with a capital letter, but here it starts with a lowercase letter. Also, it looks like the period is missing at the end of all comments.

vladsha-dev (changes requested):

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,95 @@
...
+                self.free_download_records = 0
+                self.paid_download_records -= remaining_records
+
+            self.save()
+
+    @staticmethod
+    def get_or_create_user_download_limit(
+        user,
+        custom_date=None
+    ) -> Optional["FacilityDownloadLimit"]:
```
vladsha-dev, line 83: To avoid wrapping it in quotes, you can use the following import:
`from __future__ import annotations`
You can check that the same approach is used in other files.

vladsha-dev (changes requested):

src/react/src/components/FilterSidebarFacilitiesTab.jsx
```diff
@@ -355,8 +354,8 @@ function FilterSidebarFacilitiesTab({
                         alternative={
                             <DownloadFacilitiesButton
                                 disabled={
-                                    facilitiesCount >=
-                                    FACILITIES_DOWNLOAD_DEFAULT_LIMIT
+                                    facilitiesCount >
+                                    user.allowed_records_number
```
VadimKovalenkoSNF, line 358: I'd better use here a safer check for allowed records:

```
disabled={
    typeof user?.allowed_records_number === 'number'
        ? facilitiesCount > user.allowed_records_number
        : true
}
```

src/django/api/models/facility_download_limit.py
Innavin369: changed

src/django/api/models/facility_download_limit.py
Innavin369: changed

src/react/src/components/NonVectorTileFilterSidebarFacilitiesTab.jsx
```diff
@@ -228,8 +228,8 @@ function NonVectorTileFilterSidebarFacilitiesTab({
                         alternative={
                             <DownloadFacilitiesButton
                                 disabled={
-                                    facilitiesCount >=
-                                    FACILITIES_DOWNLOAD_DEFAULT_LIMIT
+                                    facilitiesCount >
+                                    user.allowed_records_number
```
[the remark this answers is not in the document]
Innavin369: disabled={
facilitiesCount > user.allowed_records_number
}
is safe because allowed_records_number should always be defined as a number (not undefined or null). allowed_records_number is marked as number.isRequired in userPropType, which means React prop types guarantee it will be a number and always present.

src/react/src/components/FilterSidebarFacilitiesTab.jsx
VadimKovalenkoSNF: Ack.

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class FacilityDownloadLimit(models.Model):
+    """
+    Stores the number of facility records allowed for free download per
+    calendar year, the number of paid facility records for non-API
+    users, the date of last update of free facilities records, and the
+    date when paid facility records were purchased.
+    """
+    id = BigAutoField(
```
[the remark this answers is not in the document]
protsack-stephan: That's actually not a bad idea, I think if this is not to to complicated let's do so.

src/django/api/facilities_download_view_set.py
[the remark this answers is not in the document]
vladsha-dev: +1
Thanks!

src/django/api/services/facilities_download_service.py
```diff
@@ -0,0 +1,91 @@
...
+from api.models.facility_download_limit import FacilityDownloadLimit
+from api.serializers.facility.facility_query_params_serializer import (
+    FacilityQueryParamsSerializer)
+from api.exceptions import ServiceUnavailableException
+from api.constants import APIErrorMessages, FacilitiesDownloadSettings
+
+logger = logging.getLogger(__name__)
+
+
+class FacilitiesDownloadService:
```
vladsha-dev, line 18: @Innavin369 
Since you moved it to a standalone class, you no longer need `self` (which refers to the instance of the class), as the class itself is sufficient in this context. I suggest making all the methods in this class static by using the `@staticmethod` decorator.

Also, since the methods are static, there’s no need to mark them as protected by adding an underscore prefix.

Once updated, you’ll be able to call these methods directly like this: `FacilitiesDownloadService.check_if_downloads_are_blocked()`.
No need to instantiate the class, like `service = FacilitiesDownloadService()` followed by `service.check_if_downloads_are_blocked()`.

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class FacilityDownloadLimit(models.Model):
+    """
+    Stores the number of facility records allowed for free download per
+    calendar year, the number of paid facility records for non-API
+    users, the date of last update of free facilities records, and the
+    date when paid facility records were purchased.
+    """
+    uuid = models.UUIDField(
```
vladsha-dev, line 21: - This field must be indexed. Use `db_index=True`.
- This field is primary key. Please, specify it `primary_key=True`.

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,91 @@
...
+        null=False,
```
vladsha-dev, line 22: `null` is `False` by default. See the docs -> https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.null

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,91 @@
...
+        default=uuid.uuid4,
+        unique=True,
+        editable=False,
+        help_text='Unique identifier for the facility download limit record.'
+    )
+    user = models.OneToOneField(
+        'User',
+        null=False,
```
vladsha-dev, line 30: `null` is `False` by default. See the docs -> https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.null

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,91 @@
...
+        on_delete=models.CASCADE,
+        help_text='The user to whom the download limit applies.'
+    )
+    free_download_records = models.PositiveIntegerField(
+        null=False,
```
vladsha-dev, line 35: `null` is `False` by default. See the docs -> https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.null

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,91 @@
...
+        blank=False,
```
vladsha-dev, line 36: `blank` is `False` by default. See the docs -> https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.blank

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,91 @@
...
+        default=FacilitiesDownloadSettings.FACILITIES_DOWNLOAD_LIMIT,
+        help_text=('The number of facilities the user '
+                   'can download per calendar year for free.')
+    )
+    paid_download_records = models.PositiveIntegerField(
+        null=False,
+        blank=False,
```
vladsha-dev, line 43: - `null` is `False` by default. See the docs -> https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.null
- `blank` is `False` by default. See the docs -> https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.blank

src/django/api/models/facility_download_limit.py
```diff
@@ -0,0 +1,91 @@
...
+        default=0,
+        help_text=('The number of paid facilities that the user can download.')
+    )
+    updated_at = models.DateTimeField(
+        null=False,
+        blank=False,
```
vladsha-dev, line 49: - `null` is `False` by default. See the docs -> https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.null
- `blank` is `False` by default. See the docs -> https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.blank

src/django/api/models/facility_download_limit.py
Innavin369: removed

src/django/api/services/facilities_download_service.py
Innavin369: done

PR #655, opened 2025-06-11
[OSDEV-1951] Fix the issue when script can't mark rows red

protsack-stephan:
Change the name from `__mark_row` to `_mark_row` because the script was failing during the execution. Also re-thrown the error from database insert instead of just breaking the loop.


VadimKovalenkoSNF (approved):
Approved.

Innavin369 (approved):

PR #660, opened 2025-06-19
Upgraded the RDS instance type for the Test environment to improve performance.

vladsha-dev:
The RDS instance type for the Test environment has been upgraded to `db.t3.2xlarge` to handle search requests across over 1.2 million production locations now present in the database.


roman-stolar (approved):
Approved.

PR #664, opened 2025-06-20
[RBA][OSDEV-2029] Create synchronization script that will merge the data between OS Hub and RBA instances

VadimKovalenkoSNF:
Fix [OSDEV-2029](https://opensupplyhub.atlassian.net/browse/OSDEV-2029)
List of cases that need to be done or re-considered for this script:

1. Prepare full list of tables that can possibly sync. Probably, it is better to check UUID column presence. - **DONE** in related [doc](https://opensupplyhub.atlassian.net/wiki/spaces/SD/pages/780926979/RFC+-+One-Way+Data+Synchronization+from+OS+Hub+to+RBA+Platform).
2. Verify that `origin_source` column will have `rba` value if it has been added by RBA itself. - Probably not in scope of this script.
3. What if creting updating production location takes time on OS Hub, and we need to wait for it to be created / updated before we can sync it to RBA?
4. Shall this script always run in a separate AWS job?
5. Shall we block RBA database from being updated while this script is running?
6. What if pipeline crashed while running this script?
7. How to monitor the status of this script and diffs between OS Hub and RBA?
8. Shall we have an ability to stop this script?
9. Is it possible to revert this script to the previous revision?
10. It looks like we need a separate database table to store the status of the sync process (and related information).
11. If there will be a table to store sync status, we need to apply respected UI (probably, in Django admin panel).
12. Shall it be bulk sync impicitly or it is better to check specific tables to be updated? For this, we need to apply specific pattern. I mean, for the sake of better optimization we can check only `updated_at` fields and compare them with RBA before final sync.
13. How to test this locally? Shall we use Docker with local DB that will emulate RBA DB? - **DONE**. There is a separate RBA DB instance connected to local Django by exposing same local network whereby we can run Django commands, including this sync sript.
14. Need adapter class to communicate with RBA DB from OSHub Django OR listern to RBA DB events (like trigger, etc). - Might be irrelevant for MVP+.
15. Need to Yes/No prompt to confirm sync if confict found. Highlight confilcted fields. - Might be irrelevant, there should not be conflicts in a perfect scenario.
16. Add tests for this script.
17. Measure the time of script execution. - **DONE**. Example of script execution summary:
image

18. Preserve sync logs in S3 bucket or/and connect CloudWatch.
19. Make sure deletion are not syncing.
20. Need to stop DBs for OS Hub and RBA while sync process to avoid conflicts.
21. Sync script might require refactoring if we switch to another version of PostgreSQL.

[OSDEV-2029]: https://opensupplyhub.atlassian.net/browse/OSDEV-2029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vlad-shapik:
Closed the PR in favor of https://github.com/opensupplyhub/open-supply-hub/pull/693.

PR #675, opened 2025-07-02
[OSDEV-1913] Fix an error downloading results from the main search.

Innavin369:
[OSDEV-1913](https://opensupplyhub.atlassian.net/browse/OSDEV-1913) Error downloading results from main search.

The `max_length` for the `path` field in the `DownloadLog` model has been increased from 2083 to 4096 to fix the too long value error.
Added migration **0172_increase_path_max_length** which increases `max_length` for the `path` field in the `DownloadLog` model.

[OSDEV-1913]: https://opensupplyhub.atlassian.net/browse/OSDEV-1913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


doc/release/RELEASE-NOTES.md
```diff
@@ -49,8 +50,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Database changes
 * *Describe high-level database changes.*
 
-#### Migrations:
-* 0170_add_uuid_to_relevant_tables.py - A new `uuid` field was added to the following tables:
+#### Migrations
+* 0170_add_uuid_to_relevant_tables - A new `uuid` field was added to the following tables:
```
vladsha-dev, line 54: Do we need to remove .py?

vladsha-dev (changes requested):

doc/release/RELEASE-NOTES.md
vladsha-dev: Otherwise, it feels like the file name is incomplete without the .py extension. Using dashes in the name makes it seem like something is missing. Or why use dashes instead of proper spaces between words?

src/django/api/serializers/log_download_query_params.py
```diff
@@ -1,10 +1,22 @@
+from rest_framework.exceptions import ValidationError
 from rest_framework.serializers import (
     CharField,
     IntegerField,
     Serializer,
 )
 
+from api.models.download_log import DownloadLog
```
VadimKovalenkoSNF, line 8: Pls, move this to one line up. I see blank line above.

VadimKovalenkoSNF (changes requested):
Pls, apply minor fix.

PR #680, opened 2025-07-08
[OSDEV-2079] Introduced lead-in copy to inform users when the download limit has been reached on the search page

vladsha-dev:
[OSDEV-2079](https://opensupplyhub.atlassian.net/browse/OSDEV-2079)
Introduced lead-in copy to notify only logged-in users about how the download limit works on the main location search page (/facilities) for non-private instances and in non-embedded mode of the platform.

[OSDEV-2079]: https://opensupplyhub.atlassian.net/browse/OSDEV-2079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vladsha-dev:
> @vladsha-dev Please check also NonVectorTileFilterSidebarFacilitiesTab.jsx file, probably we also need to add those info there.

Hello @roman-stolar ,
There’s no need to add it to that component, because the change should be made in the parent component of NonVectorTileFilterSidebarFacilitiesTab.

src/react/src/util/styles.js
```diff
@@ -1721,3 +1721,24 @@ export const makeStyledTooltipStyles = () =>
...
     });
+
+export const downloadLimitInfoStyles = () =>
+    Object.freeze({
+        informativeWrapper: Object.freeze({
+            marginTop: '15px',
+        }),
+        informativeText: Object.freeze({
+            fontSize: '16px',
+            color: '#191919',
```
Innavin369, line 1732: Should we use constant JET_BLACK from COLOURS here?

vladsha-dev: Yes, good catch!
Thank you.

vladsha-dev: @Innavin369 
Done

Innavin369 (approved):

Innavin369 (approved):

PR #692, opened 2025-07-14
[OSDEV-2084] Enable Self Service Data Upload (Dromo)

mazursasha1990:
[OSDEV-2084](https://opensupplyhub.atlassian.net/browse/OSDEV-2084) - Enable Self Service Data Upload (Dromo).

The PR enables a new self-service data upload flow using the Dromo uploader, which is implemented behind the `enable_dromo_uploading` feature flag.

[OSDEV-2084]: https://opensupplyhub.atlassian.net/browse/OSDEV-2084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -867,9 +867,11 @@ variable "dark_visitors_token" {
 variable "dromo_license_key" {
   type        = string
   description = "Dromo license key for data management"
+  sensitive   = true
 }
 
 variable "dromo_schema_id" {
   type        = string
   description = "Dromo schema ID for data management"
+  sensitive   = true
 }


src/react/src/components/ContributeForm.jsx
```diff
@@ -140,17 +146,33 @@ const ContributeForm = ({
...
+                    </FeatureFlag>
+                </div>
                 <p style={contributeFormStyles.fileNameText}>{filename}</p>
+                <p
+                    style={{
+                        fontSize: '12px',
+                        marginTop: '8px',
+                    }}
+                >
+                    Upload CSV or Excel files using our enhanced file selector
```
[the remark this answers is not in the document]
mazursasha1990: It's correct behavior.

src/react/src/util/util.js
```diff
@@ -1719,3 +1719,32 @@ export const formatExtendedField = ({
...
+
+    const inputEl = fileInput.current;
+    if (inputEl) {
+        const dataTransfer = new DataTransfer();
+        dataTransfer.items.add(csvFile);
+        inputEl.files = dataTransfer.files;
+
+        updateFileName(fileInput);
+    }
+};
```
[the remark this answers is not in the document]
mazursasha1990: No need in the current implementation.

src/react/src/components/ContributeForm.jsx
```diff
@@ -140,17 +146,33 @@ const ContributeForm = ({
...
+                        onClick={selectFile}
+                        type="button"
+                        variant="outlined"
+                        color="primary"
+                        className="outlined-button"
+                        disableRipple
+                    >
+                        Select Facility List File
+                    </MaterialButton>
+                    <FeatureFlag flag={ENABLE_DROMO_UPLOADING}>
```
vladsha-dev, line 160: Doesn’t React throw errors locally if the flag is not active?
If it does, I’d recommend passing a stub for the `alternative` argument, like this: <></>

src/react/src/components/SelfServiceDromoUploader.jsx
```diff
@@ -0,0 +1,53 @@
...
+                type="button"
+                variant="contained"
+                style={uploaderButtonStyle}
+                disableRipple
+            >
+                Beta Self Service Upload
+            </MaterialButton>
+
+            <DromoUploader
+                licenseKey="ee427cf5-da27-4f28-a260-c9a17d02ad30"
```
vladsha-dev, line 38: As I understand it, the licenseKey is similar to an API key issued specifically for OS Hub and should be secured in some way. Is it restricted to the OS Hub domain name?
Would it work locally with this licenseKey, as well as in all other environments?

vladsha-dev (changes requested):

src/react/src/components/SelfServiceDromoUploader.jsx
mazursasha1990: `licenseKey` is a Dromo Frontend API Key, and it's designed to be embedded in a client app.
It is the same for all environments.

roman-stolar (approved):
LGTM

src/react/src/components/SelfServiceDromoUploader.jsx
mazursasha1990: Unfortunately, I don't have access to the OSHub Dromo account, so I can't check if there are restricted domains (I suggest there are no). 
But I created my own Dromo account for testing purposes, and there is a description under Whitelisted domains: "Not enforced in development mode."
It allows us to build and test integration locally without restrictions, even if some domain restrictions were enforced.

My suggestion is to leave it as it is.
Screenshot from 2025-07-18 12-45-20

vladsha-dev (approved):
LGTM

roman-stolar (approved):
Re-approved

PR #700, opened 2025-07-24
Test VPN server on Preprod

vlad-shapik:


vlad-shapik:
Testing of the VPN server has been completed in the Preprod environment. The PR is no longer needed and can be closed.

PR #721, opened 2025-08-19
[OSDEV-2077] [RBA Infra Setup][MVP+] Establish VPC peering between OS Hub and RBA environments to enable database access from AWS Batch job

vlad-shapik:


vlad-shapik:
No longer relevant

PR #725, opened 2025-08-20
Osdev 2156 update copy for premium data downloads email templates

FrancescaOSH:
Updated the copy for Premium Data Downloads email templates to include information about all other available options to obtain data (link to bulk download inquiry form and link to Free Premium Feature access policy for Civil Society Organizations).


roman-stolar (approved):
LGTM

PR #735, opened 2025-09-05
[Backmerge][OSDEV-1946] Adjusted infra configs for the RBA sync script and optimized logging in the script.

vlad-shapik:
[[OSDEV-1946](https://opensupplyhub.atlassian.net/browse/OSDEV-1946)]

Adjusted infra configs for the RBA sync script and optimized logging in the script:

- Added a filter in the `DB - Apply Anonymized DB` and `Deploy to AWS` GitHub workflows - for the logic that retrieves EFS storage during OpenSearch cleanup, ensuring it points to the correct storage since we now have at least two in RBA.
- Introduced the ability to access the DB sync EFS storage from the bastion EC2 instance for management purposes.
- Implemented native Django chunking for table record retrieval.
- Removed the `attempt_duration_seconds` setting for the DB sync batch job to avoid blocking runs that may last several days.
- Updated the `max_connections_percent`, `max_idle_connections_percent`, and `idle_client_timeout` settings for the RDS proxy to allow the Django sync script to remain idle when working with the RBA DB, but not with Production.
- Removed the `verbose` parameter from the DB sync script so that all logs are always shown, as they may be necessary for regular monitoring.

[OSDEV-1946]: https://opensupplyhub.atlassian.net/browse/OSDEV-1946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


roman-stolar (approved):
LGTM

PR #796, opened 2025-11-06
[RBA][OSDEV-2246] Create backfill script to insert isic_4 to api_extendedfield table

VadimKovalenkoSNF:
Fix [OSDEV-2246](https://opensupplyhub.atlassian.net/browse/OSDEV-2246)
1. Add `backfill_isic_4_extended_fields.py` to insert `isic_4` to `api_extendedfield table` of RBA instance.
2. Show `isic_4` field in GET `api/facilities` response.

Screenshot from 2025-11-18 13-14-25

Example of `isic_4` fileds in the output of `api/v1/production-locations/{os_id}`:

```
...
"isic_4": [
    {
        "id": 4875940,
        "is_verified": false,
        "value": {
            "raw_value": {
                "class": "282 - Manufacture of machinery for mining, quarrying and construction",
                "group": "251 -  Manufacture of tanks, reservoirs and containers of metal",
                "section": "C - Manufacturing",
                "division": "28 - Manufacture of machinery and equipment n.e.c."
            }
        },
        "updated_at": "2025-11-20T11:20:28.523245Z",
        "contributor_name": "RBAVadimKovalenkoSNF",
        "contributor_id": 17564,
        "value_count": 1,
        "is_from_claim": false,
        "field_name": "isic_4",
        "verified_count": 0,
        "source_by": null,
        "unit": null,
        "label": null
    },
    {
        "id": 4863701,
        "is_verified": false,
        "value": {
            "raw_value": {
                "class": "7020 - Management consultancy activities",
                "group": "702 - Management consultancy activities",
                "section": "M - Professional, scientific and technical activities",
                "division": "70 - Activities of head offices; management consultancy activities"
            }
        },
        "updated_at": "2025-11-20T11:11:12.327735Z",
        "contributor_name": "Responsible Business Alliance",
        "contributor_id": 14267,
        "value_count": 1,
        "is_from_claim": false,
        "field_name": "isic_4",
        "verified_count": 0,
        "source_by": null,
        "unit": null,
        "label": null
    }
]
...
```

[OSDEV-2246]: https://opensupplyhub.atlassian.net/browse/OSDEV-2246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/react/src/util/constants.jsx
+++ b/src/react/src/util/constants.jsx
@@ -1132,7 +1132,7 @@ export const EXTENDED_FIELD_TYPES = [
         fieldName: 'isic_4',
         formatValue: value => {
             const { section, division, group, class: isicClass } =
-                value?.raw_value || {};
+                value?.raw_value || value || {};
             return [
                 section && `Section: ${section}`,
                 division && `Division: ${division}`,


src/django/api/serializers/facility/facility_index_serializer.py
```diff
@@ -407,7 +407,15 @@ def get_extended_fields(self, facility):
...
-            grouped_data[field_name] = data
+            # For ISIC-4, return only the raw value objects (no metadata).
+            if field_name == ExtendedField.ISIC_4:
+                grouped_data[field_name] = [
+                    item.get('value')
+                    for item in data
+                    if item.get('value') is not None
+                ]
+            else:
+                grouped_data[field_name] = data
```
vlad-shapik: The response should be structured a little bit differently - similar to `number_of_workers`, where we return `value`, `updated_at`, `contributor_name`, etc., which are usually displayed as part of the data point on the location profile page. To keep it consistent, we should return the same fields for this metric as well. It will work well if the response of the GET `/facilities/{os_id}` endpoint has the following format to integrate smoothly into our platform:
```
{
  ...,
  "properties": {
    ...,
    "extended_fields": {
      ...,
      "isic_4": [
        ...,
        {
          "contributor_id": 4390,
          "contributor_name": "G-STAR RAW",
          "created_at": "2025-04-02T07:12:52.402106Z",
          "field_name": "number_of_workers",
          "id": 3313183,
          "is_from_claim": false,
          "is_verified": false,
          "updated_at": "2025-04-04T22:02:56.688767Z",
          "value": {
            "section": "blabla",
            "division": "blabla",
            "group": "blabla",
            "class": "blabla",
          },
          "value_count": 5,
          "verified_count": 0
        }
        ...,
      ]
    }
  }
}
```

See how it looks for `number_of_workers`:
Screenshot 2025-11-12 at 12 21 21

Then you won’t need this specific code you introduced.

vlad-shapik (changes requested):

src/django/api/serializers/facility/facility_index_serializer.py
```diff
@@ -414,7 +414,24 @@ def get_extended_fields(self, facility):
...
+                    if isinstance(value, list):
+                        if len(value) == 1:
+                            copied = dict(item)
+                            copied['value'] = value[0]
+                            normalized.append(copied)
+                        else:
+                            normalized.append(item)
+                    else:
+                        normalized.append(item)
+                grouped_data[field_name] = normalized
```
vlad-shapik: Isn’t `130_index_extended_fields.sql` already sufficient to build the value for that field so we don’t need to parse it manually, which increases endpoint latency? Keep in mind that the `isic_4` field is still indexed by the SQL function I mentioned, even though you’re parsing it here - so it’s essentially double work. I recommend saving it properly to the `ExtendedField` table so the trigger on `api_extendedfield` picks it up correctly. Then this custom transformation is unnecessary; we only need to store it properly in the ExtendedField table. If the field data is saved in a format compatible with `130_index_extended_fields.sql`, the `isic_4` field will be saved to the `data` variable in the `else` block on lines 411–415. Check how new fields like `duns_id`, `lei_id`, and `rba_id` are indexed and returned in the `get_extended_fields` method - you need to do it in the same way to preserve endpoint latency.

It basically means that since we need to unwrap it into an object, we should already save it as an object to the `ExtendedField` table in the script you created. Later, when this field is accepted by POST `/api/v1/production-locations/`, it must be saved in a way that can be easily indexed by the `130_index_extended_fields.sql` SQL function. Check how this is done for `duns_id`, `lei_id`, and `rba_id` in [this file](https://github.com/opensupplyhub/open-supply-hub/blob/main/src/django/api/extended_fields.py#L112) when they are saved, and follow the same approach in the custom script.

roman-stolar (commented):
In general LGTM, please fix CI

vlad-shapik (dismissed):
LGTM.
Thank you for taking the comments into account!

vlad-shapik (approved):
LGTM

PR #802, opened 2025-11-07
[OSDEV-2260] Update the logic about saving extended fields for the Claim flow

protsack-stephan:
Removed the unnecessary `facility_type` checks in the `facilities_view_set.py` to support proper saving of the data. The function assumed that the value is an array and that caused errors in the creation of extended fields.

On front-end data format has been changed to send array values as a **pipe-separated (`|`) string** instead of a comma-separated one. The `extended_fields.py` file has been updated to correctly parse this new pipe-delimited format, ensuring data is handled correctly.


VadimKovalenkoSNF (approved):
Re-approved

PR #805, opened 2025-11-07
[OSDEV-2264] Prevent saving empty lists for facility production types

protsack-stephan:
This commit addresses an issue where an empty list (`[]`) was being saved for the `facility_production_types` field during a facility claim, even if the user provided no data for it. It was causing empty `Processing Type` values on the facility profile page.

The `facilities_view_set.py` has been updated to remove the default empty list when accessing `facility_production_types` from the validated data. This ensures that a `None` value is passed if the field is not present, preventing an empty list from being saved to the database when no production types are specified.


VadimKovalenkoSNF (approved):
LGTM

PR #806, opened 2025-11-10
[RBA][OSDEV-2265] Update DB - Save Anonymized DB to make anon dump of RBA DB

VadimKovalenkoSNF:
Implement [OSDEV-2265](https://opensupplyhub.atlassian.net/browse/OSDEV-2265)

`DB - Save Anonymized DB` action: `Production` , `Rba` and `Development` environments can be selected for anonymized dump creation.

**Environment selection and bastion lookup**
Added ENVIRONMENT support (`Production`, `Rba`, `Development`) and logs for selected env/region.
Chooses AWS creds based on env (TEST for Development; PROD otherwise) to call aws ec2 describe-instances with Environment=<env> tag.
Why: resolve the correct bastion per environment without hardcoding.

**SSH key handling and authentication**
Uses mounted /keys/key only (no GitHub expressions in the script).
Ensures perms (chmod 600), logs safe fingerprint (if parsable), logs key size.
Tries common users (ec2-user, ubuntu) for SSH port-forward; logs outcome.
Why: avoid secret leakage, make SSH auth explicit and diagnosable across AMIs.

**Tunnel readiness and robust pg_dump**
Waits for localhost:5433 (`pg_isready`) before `pg_dump`.
Dynamic dump naming per env (`osh_prod_large`/`dev_large`/`rba_large`).
Logs dump sizes, warns if tiny.
Why: prevent “connection refused” and empty dump artifacts, make results observable.

**Local restore hardening**
Pre-drops PostGIS extensions (postgis/postgis_topology/postgis_tiger_geocoder) before pg_restore.
Captures `pg_restore` stderr to /dumps/restore.err; prints first error context.
Why: the postgis/postgis image auto-enables extensions; dropping them avoids `cannot drop extension postgis because other objects depend on it`, and error capture shows root causes.

**Anonymization SQL invocation**
Switched from a shell variable to a here-doc psql <<'SQL' block.
Why: avoid `SQL_SCRIPT: parameter not set` under `set -u` and keep SQL always defined.

**S3 upload and logging**
Logs S3 destination, prints anonymized dump size.
Why: traceability of outputs.

[OSDEV-2265]: https://opensupplyhub.atlassian.net/browse/OSDEV-2265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/release/RELEASE-NOTES.md
+++ b/doc/release/RELEASE-NOTES.md
@@ -19,6 +19,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Architecture/Environment changes
 * [OSDEV-2244](https://opensupplyhub.atlassian.net/browse/OSDEV-2244) - Added `backfill_isic_4_extended_fields.py` to insert `isic_4` to `api_extendedfield` table of RBA instance. Show `isic_4` field in `GET api/facilities` response.
+* [OSDEV-2281](https://opensupplyhub.atlassian.net/browse/OSDEV-2281) - Increased the CPU and memory allocation for the DedupeHub container to `8 CPU` and `40 GB` in the Terraform deployment configuration to address memory overload issues during production location reindexing for the `Production`, `Pre-prod`, and `RBA` environments.
+
+### Bugfix
+* [OSDEV-2284](https://opensupplyhub.atlassian.net/browse/OSDEV-2284) - Temporarily increase heap size for logstash plugin installation.
 
 ### What's new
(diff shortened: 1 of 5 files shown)


doc/release/RELEASE-NOTES.md
```diff
@@ -3,6 +3,21 @@ All notable changes to this project will be documented in this file.
...
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html). The format is based on the `RELEASE-NOTES-TEMPLATE.md` file.
 
+## Release 2.16.0
+
+## Introduction
+* Product name: Open Supply Hub
+* Release date: November 15, 2025
+
+### Architecture/Environment changes
+* [OSDEV-2265](https://opensupplyhub.atlassian.net/browse/OSDEV-2265) - `DB - Save Anonymized DB` GitHub action now supports Production/Rba selection.
```
roman-stolar: Not sure that release notes needed here as this changes will apply directly and not related to deployment process

roman-stolar (commented):
Minor comment

roman-stolar (commented):
Left question

roman-stolar (dismissed):
LGTM

vlad-shapik (changes requested):
Could you please add to the RELEASE-PROTOCOL.md that when we run the "DB - Save Anonymized DB" workflow, we need to select Production as the environment? Please refer to line 186 in that file where this workflow is mentioned.

PR #818, opened 2025-11-20
[Backmerge][OSDEV-2249] Fixed the issue where the claim-flow state was not resetting when the OS ID changed or after a successful submission

vlad-shapik:
[OSDEV-2240](https://opensupplyhub.atlassian.net/browse/OSDEV-2240)
Implemented form state reset in the new claim flow when the OS ID changes or after a successful submission, ensuring that users do not see previously entered data when they start a new form for another OS ID or begin a new claim flow for location A in the same browser tab session - even if they were previously completing the form for location B and exited mid-form. However, if the user exits mid-form for location A without refreshing the page and then returns to the claim form for the same location A in the same tab session, the previously entered data will still be present.

[OSDEV-2240]: https://opensupplyhub.atlassian.net/browse/OSDEV-2240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/release/RELEASE-NOTES.md
+++ b/doc/release/RELEASE-NOTES.md
@@ -19,6 +19,21 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
     * `reindex_database`
 
 
+## Release 2.15.2
+
+## Introduction
+* Product name: Open Supply Hub
+* Release date: November 21, 2025
+
+### Bugfix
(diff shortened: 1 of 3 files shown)


VadimKovalenkoSNF (approved):
Good work, approved!

PR #823, opened 2025-11-25
[OSDEV-2269] [Partner Fields Enhancement] Add URI (JSON Schema) support for Partner Fields

roman-stolar:
**[OSDEV-2269](https://opensupplyhub.atlassian.net/browse/OSDEV-2269) [Partner Fields Enhancement] Add URI (JSON Schema) support for Partner Fields**

- Added URI format support for partner fields with JSON schema. Properties defined with `"format": "uri"` are rendered as clickable links on production location profile pages.

[OSDEV-2269]: https://opensupplyhub.atlassian.net/browse/OSDEV-2269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/release/RELEASE-NOTES.md
+++ b/doc/release/RELEASE-NOTES.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Code/API changes
 * [OSDEV-2280](https://opensupplyhub.atlassian.net/browse/OSDEV-2280) - Introduced a new reusable `ImportantNote` component in the new claim flow to replace custom implementations of important notes across multiple components (`BusinessStep`, `ContactInfoStep`, and `ClaimInfoSection`), improving code maintainability and consistency.
 * [Follow-up][OSDEV-2066](https://opensupplyhub.atlassian.net/browse/OSDEV-2066) - Removed the `null=True` parameter from the `Contributor.partner_fields` ManyToManyField definition in the model class to resolve Django system check warning W340. The parameter had no effect on the field behavior as ManyToManyFields store relationships in an intermediary table rather than as a database column that could contain NULL values.
+* [OSDEV-2114](https://opensupplyhub.atlassian.net/browse/OSDEV-2114) - Added the `reindex_locations_with_environmental_data` management command to perform a one-time reindexing of locations that have approved claims containing the new environmental data fields (opening/closing dates, throughput, energy consumption). This ensures that the already contributed environmental data will be displayed on the location profile once the environmental data display feature is released in version `2.17.0`. The command also includes a `--dry-run` flag to preview the affected location IDs before execution.
 
 ### What's new
 * [OSDEV-2280](https://opensupplyhub.atlassian.net/browse/OSDEV-2280) - Added prominent PII (Personally Identifiable Information) warning notes at file upload stages throughout the new claim flow to inform users that they should NOT submit documents containing personal information, home addresses, personal utility bills, or personal phone numbers, enhancing data security and user privacy protection.
@@ -28,7 +29,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 * Ensure that the following commands are included in the `post_deployment` command:
     * `migrate`
     * `reindex_database`
-    * `index_facilities_new`
+    * `reindex_locations_with_environmental_data`
 
 
 ## Release 2.16.0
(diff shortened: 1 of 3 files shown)


src/react/src/util/util.js
```diff
@@ -1898,9 +1899,18 @@ const formatRawValues = rawValues => {
     return rawValues.toString().split('|');
 };
 
-export const formatPartnerFieldValue = value => {
+export const formatPartnerFieldValue = (value, item) => {
```
vlad-shapik: Could you please improve parameter naming for clarity?

Consider renaming the parameters for better readability:
- `value` → `fieldValueData` (contains the actual field value with `raw_value`/`raw_values`)
- `item` → `fieldDefinition` (contains the field configuration/schema)

src/react/src/components/PartnerFieldSchemaValue.jsx
```diff
@@ -0,0 +1,150 @@
...
+        !value ||
+        typeof value !== 'object' ||
+        Array.isArray(value)
+    ) {
+        return value;
+    }
+
+    const schemaProperties = jsonSchema?.properties || {};
+
+    const renderedItems = Object.keys(value)
```
VadimKovalenkoSNF: The pipeline is doing too much things. Check this version suggested by AI:

```
const PartnerFieldSchemaValue = ({ value, jsonSchema }) => {
  if (
    !jsonSchema ||
    !value ||
    typeof value !== "object" ||
    Array.isArray(value)
  ) {
    return value;
  }

  const schemaProperties = jsonSchema.properties || {};

  const renderedItems = Object.entries(value).reduce((acc, [key, val]) => {
    if (shouldSkipProperty(key, schemaProperties)) {
      return acc;
    }
    const rendered = renderProperty(key, value, schemaProperties);
    if (rendered) {
      acc.push(rendered);
    }
    return acc;
  }, []);

  return <>{renderedItems}</>;
};
```

Why this is sharper:

• You iterate once, not three times.
• Object.entries makes the loop easier to read (property + its value always travel together).
• The skip and render happen in the same pass, so there’s no wasted work.
• No post-filtering needed.

src/react/src/components/PartnerFieldSchemaValue.jsx
```diff
@@ -0,0 +1,150 @@
...
+        return FORMAT_COMPONENTS[format];
+    }
+    return DefaultProperty;
+};
+
+/**
+ * Renders a single property based on format strategy.
+ */
+const renderProperty = (propertyKey, value, schemaProperties) => {
+    if (!Object.prototype.hasOwnProperty.call(value, propertyKey)) {
```
VadimKovalenkoSNF: Repeat: That `hasOwnProperty.call` line is a relic from defensive JavaScript. In your case it’s pointless: you’re already iterating over `Object.keys(value)` (or `Object.entries(value)` as I suggested in the comment below), so you only get own-properties anyway. You can drop this check.

src/react/src/components/PartnerFieldSchemaValue.jsx
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * Gets the format type from a property schema.
+ */
+const getFormatFromSchema = propertySchema => propertySchema?.format || null;
+
+/**
+ * Checks if a property key should be skipped.
+ */
+const shouldSkipProperty = (propertyKey, schemaProperties) => {
+    if (!Object.prototype.hasOwnProperty.call(schemaProperties, propertyKey)) {
```
VadimKovalenkoSNF: That `hasOwnProperty.call` line is a relic from defensive JavaScript. In your case it’s pointless: you’re already iterating over `Object.keys(value)` (or `Object.entries(value)` as I suggested in the comment below), so you only get own-properties anyway. You can drop this check.

src/react/src/util/util.js
```diff
@@ -1898,14 +1903,22 @@ const formatRawValues = rawValues => {
     return rawValues.toString().split('|');
 };
 
-export const formatPartnerFieldValue = value => {
-    const { raw_values, raw_value } = value;
+export const formatPartnerFieldValue = (fieldValueData, json_schema) => {
```
vlad-shapik: Could you rewrite the `json_schema` parameter in camelCase so it aligns better with JavaScript conventions?

PR #827, opened 2025-11-28
[Follow-up][OSDEV-2066] Removed null=true from the Contributor.partner_fields

vlad-shapik:
[OSDEV-2066](https://opensupplyhub.atlassian.net/browse/OSDEV-2066)
Removed the `null=True` parameter from the `Contributor.partner_fields` ManyToManyField definition in the model class to resolve Django system check warning W340. The parameter had no effect on the field behavior as ManyToManyFields store relationships in an intermediary table rather than as a database column that could contain NULL values.

[OSDEV-2066]: https://opensupplyhub.atlassian.net/browse/OSDEV-2066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


VadimKovalenkoSNF (approved):
Approved.

PR #902, opened 2026-03-05
[OSDEV-2121] Hide download limit lead-in copy on unfiltered search

TylerHeath1:
## What

When a user clicks "Find Facilities" without any search criteria or filters selected, the data download limits lead-in copy no longer displays above the search results. The banner continues to appear when filters are active and the result count exceeds the user's download limit.

## How

Added a `hasAppliedSearchFilters` utility that returns true when search text or at least one filter (excluding sort order) is active. This is computed during `fetchFacilities` and stored in the facilities Redux state as `hasAppliedFilters`. The `FilterSidebarHeader` component now requires `hasAppliedFilters` to be true (in addition to the existing conditions) before showing the download limit banner.

## How to test

1. Log in as a registered user
2. Click "Find Facilities" without entering any search criteria or selecting any filters
3. Verify the download limits lead-in copy does **not** appear
4. Enter search criteria or select a filter (e.g., a country), then click "Find Facilities"
5. If results exceed your download limit, verify the lead-in copy **does** appear

## Jira

[OSDEV-2121](https://opensupplyhub.atlassian.net/browse/OSDEV-2121)

[OSDEV-2121]: https://opensupplyhub.atlassian.net/browse/OSDEV-2121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TylerHeath1:
Responding to CodeRabbit's two nitpick suggestions:

**1. Centralizing active-filter key handling (FilterSidebarHeader.jsx lines 16-33):**
Intentionally kept as explicit checks rather than importing from the reducer to avoid coupling between the component and reducer internals. The filter keys change very rarely — the last filter was added over a year ago. If a new filter is added, the developer would naturally update this check alongside the reducer.

**2. Adding array-based active-filter test (FilterSidebarHeader.test.js lines 37-40):**
The `hasActiveFilters` function is a simple OR chain — if any field is non-empty it returns true. Testing one representative field is sufficient. Adding a test per filter type would be testing the OR operator, not our logic.

mazur-snf (approved):

vlad-shapik (approved):
LGTM

PR #949, opened 2026-04-07
[OSDEV-2423] Document partner fields in the facility retrieve API endpoint

protsack-stephan:
The `partner_fields` data was already being returned in the response but was invisible in the generated API docs because it lacked a serializer schema and was not referenced in the Swagger decorator. This made it difficult for API consumers to discover and understand the structure of Spotlight (partner) data.

Changes:
- Add `PartnerFieldEntrySerializer` to describe the shape of individual partner field entries (value, contributor info, verification status, metadata, etc.).
- Annotate `get_partner_fields` with `@swagger_serializer_method` so drf-yasg picks up the return type.
- Add `responses={200: FacilityIndexDetailsSerializer}` to the `retrieve` action's `@swagger_auto_schema` decorator.
- Extend the docstring sample response to include a `partner_fields` example showing nested value structures.


src/django/api/serializers/facility/facility_index_details_serializer.py
```diff
@@ -35,6 +37,31 @@
...
+    is_from_claim = serializers.BooleanField(read_only=True)
+    field_name = serializers.CharField(read_only=True)
+    source_by = serializers.CharField(
+        read_only=True, allow_null=True, allow_blank=True
+    )
+    unit = serializers.CharField(read_only=True, allow_blank=True)
+    label = serializers.CharField(read_only=True, allow_blank=True)
+    base_url = serializers.CharField(read_only=True, allow_blank=True)
+    display_text = serializers.CharField(read_only=True, allow_blank=True)
+    json_schema = serializers.JSONField(read_only=True, allow_null=True)
```
vlad-shapik: Nitpick:
It would be great to move it to a separate file so the current file does not get too large.

vlad-shapik (dismissed):
LGTM
Left one optional comment

src/django/api/serializers/facility/facility_index_details_serializer.py
protsack-stephan: Updated.

vlad-shapik (approved):
LGTM

PR #951, opened 2026-04-09
Suppress service output from CI deployment logs

protsack-stephan:
This PR removes additional informational log lines and suppresses ECS service update output that could leak infrastructure details (service names, cluster identifiers, distribution IDs) in CI logs.


VadimKovalenkoSNF (approved):
LGTM

PR #957, opened 2026-04-16
[OSDEV-2426] Added maintenance banner

VadimKovalenkoSNF:
Implements [OSDEV-2426](https://opensupplyhub.atlassian.net/browse/OSDEV-2426)

1. Add maintenance banner when `disable_list_uploading` flag is active.
2. Maintenance status can be verified by navigating to a different frontend route, without requiring a browser refresh.
3.  With Lodash `debounce` configured at 300ms, calls to `refreshFeatureFlags()` that arrive within a 300ms window of each other are collapsed - only the last one in that window actually fires the API request, and it fires 300ms after that last call.
4. Feature with the flag refetching will affect all flag switchers on the frontend.
5. Embed mode is not affected. When embed is true, `<Navbar />` is never rendered at all

image

[OSDEV-2426]: https://opensupplyhub.atlassian.net/browse/OSDEV-2426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/react/src/actions/featureFlags.js
```diff
@@ -32,3 +32,11 @@ export function fetchFeatureFlags() {
             );
     };
 }
+
+export function refreshFeatureFlags() {
+    return dispatch =>
+        apiRequest
+            .get(makeGetAPIFeatureFlagsURL())
+            .then(({ data }) => dispatch(completeFetchFeatureFlags(data)))
+            .catch(() => {});
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: I've introduced Lodash `debounce` wrapper to guard this request.

src/react/src/Routes.jsx
```diff
@@ -93,10 +93,19 @@ class Routes extends Component {
 
         window.setGridColorRamp = this.props.setRamp;
         this.props.getFeatureFlags();
+        this.unlistenHistory = history.listen(() =>
+            this.props.refreshFeatureFlags(),
```
protsack-stephan, line 97: Do I understand correctly, that this will make an API request on every page? Browser allows only for 6 concurrent API requests per page, need to double check if this might become an issue. Maybe it makes sense to cache the `feature-flags` endpoint on the CloudFront?

protsack-stephan (commented):
Good work! A comment from me.

src/react/src/Routes.jsx
VadimKovalenkoSNF: > Do I understand correctly, that this will make an API request on every page?

Yes, exactly

>  Browser allows only for 6 concurrent API requests per page, need to double check if this might become an issue.

The 6-connection browser limit applies to HTTP/1.1 only — CloudFront serves this app over HTTP/2, which multiplexes all requests over a single connection with no per-host cap. See `http_version    = "http2"` in our CDN settings.

The `refreshFeatureFlags` call is also debounced at 300ms, so at most one small GET fires per navigation pause rather than one per click. 

> Maybe it makes sense to cache the feature-flags endpoint on the CloudFront?

Caching `/api-feature-flags/` at CloudFront would break the feature — the current config intentionally bypasses CDN caching (`default_ttl = 0`, all headers forwarded) so flag changes are picked up immediately. But this is a really cheap request, I don't expect it will be a bottleneck.

protsack-stephan (dismissed):

vlad-shapik (dismissed):
LGTM

protsack-stephan (dismissed):

vlad-shapik (dismissed):

vlad-shapik (dismissed):

protsack-stephan (dismissed):

protsack-stephan (approved):

PR #958, opened 2026-04-17
[OSDEV-2525] Add Spotlight data to the user profile

protsack-stephan:
The public user profile page previously showed limited information about a contributor's data. This PR adds two new section **Spotlight Locations** and updates **Facility Lists** section, so visitors can see the facilities a contributor has provided spotlight data for, along with their published facility lists, directly from the profile.

- **New API endpoints** for the profile page:
  - `/user-profile/<pk>/facilities/?spotlight=true` — returns facilities where the contributor has spotlight partner field data (wage indicator, living wage, or other extended fields). Falls back to the contributor's full facility list when `spotlight` is not requested.
  - `/user-profile/<pk>/facility-lists/` — returns the contributor's approved, public facility lists.
  - Both endpoints are cursor-paginated and cached for 5 minutes via a new `view_cache` backend.

- **New React components** rendered on public (non-editable) profiles:
  - `UserProfileProductionLocations` — "Spotlight Locations" expandable panel showing OS ID, Name, and Address with a "Load More" button and an info tooltip linking to the data integrations page.
  - `UserProfileFacilityLists` — expandable panel listing facility lists by Name and Description.

- **Facility lists removed from the profile serializer** — `get_facility_lists` now returns a stub `[]` since the data is served by the dedicated paginated endpoint instead.

- **CDN TTL increase** — min TTL raised from 0 to 120s and default from 0 to 240s to improve caching for these and other profile responses.


src/react/src/components/UserProfileProductionLocations.jsx
```diff
@@ -0,0 +1,444 @@
...
+    },
+    loaderContainer: {
+        minHeight: '60px',
+    },
+    viewButton: {
+        fontSize: '14px',
+    },
+});
+
+function UserProfileProductionLocations({
```
vlad-shapik: Could you please use ES6 arrow functions for consistency with the other parts of the new React code?
I noticed a few other similar cases in the new code in this PR as well.

src/react/src/components/UserProfileFacilityLists.jsx
```diff
@@ -0,0 +1,392 @@
...
+    },
+    loaderContainer: {
+        minHeight: '60px',
+    },
+    viewButton: {
+        fontSize: '14px',
+    },
+});
+
+function UserProfileFacilityLists({
```
vlad-shapik: Could you please use ES6 arrow functions for consistency with the other parts of the new React code?
I noticed a few other similar cases in the new code in this PR as well.

src/react/src/actions/profile.js
```diff
@@ -137,6 +139,156 @@ export function createAPIToken() {
...
+                        err,
+                        'An error prevented fetching facility lists',
+                        failFetchFacilityLists,
+                    ),
+                ),
+            );
+    };
+}
+
+export function fetchMoreFacilityLists(nextPageUrl) {
```
vlad-shapik: Could you please use ES6 arrow functions for consistency with the other parts of the new React code?
I noticed a few other similar cases in the new code in this PR as well.

src/react/src/components/UserProfileProductionLocations.jsx
```diff
@@ -0,0 +1,444 @@
...
+const cellHeader = theme =>
+    Object.freeze({
+        fontWeight: '700',
+        padding: `0 ${theme.spacing.unit}px`,
+        fontSize: '18px',
+    });
+
+const cell = theme =>
+    Object.freeze({
+        fontSize: '18px',
```
vlad-shapik: I think it would be better to place the styles in a separate file, such as styles.js, and create a folder for this component to keep everything there.

src/react/src/components/UserProfileFacilityLists.jsx
```diff
@@ -0,0 +1,392 @@
...
+
+const cell = theme =>
+    Object.freeze({
+        fontSize: '18px',
+        padding: `0 ${theme.spacing.unit}px`,
+        fontWeight: '500',
+    });
+
+const muiStyles = theme => ({
+    appGridContainer: {
```
vlad-shapik: I think it would be better to place the styles in a separate file, such as styles.js, and create a folder for this component to keep everything there.

vlad-shapik (changes requested):

src/react/src/components/UserProfileFacilityLists.jsx
vlad-shapik: Just to make this file a little easier to read.

src/react/src/components/UserProfileProductionLocations.jsx
vlad-shapik: Just to make this file a little easier to read.

src/react/src/components/UserProfileFacilityLists.jsx
protsack-stephan: Thanks, updated!

src/react/src/components/UserProfileProductionLocations.jsx
protsack-stephan: Thanks, updated!

src/react/src/actions/profile.js
protsack-stephan: Thanks, updated!

src/react/src/components/UserProfileFacilityLists.jsx
protsack-stephan: Thanks, updated!

src/react/src/components/UserProfileProductionLocations.jsx
protsack-stephan: Thanks, updated!

vlad-shapik (dismissed):

vlad-shapik (approved):
GREAT!

PR #960, opened 2026-04-20
[Hotfix][OSDEV-2529] Fix duplicate throttle cache key collision for production location endpoints

protsack-stephan:
Fixed `DuplicateThrottle` incorrectly rejecting requests to different production locations as duplicates when the same user sends identical payloads (e.g., PATCH to `/api/v1/production-locations/123/` and `/api/v1/production-locations/456/`).

Added the view's `pk` to the throttle cache key so that requests targeting different resources are keyed independently.

The `DuplicateThrottle` cache key was `duplicate:{user_id}:{data_hash}`, which meant two requests with the same body but directed at different production locations would collide. This caused legitimate update requests to be throttled as duplicates.

The fix includes the resource `pk` (when present) in the cache key:

```
# Before
duplicate:{user_id}:{data_hash}

# After (with pk)
duplicate:{user_id}:{pk}:{data_hash}

# After (without pk, e.g. POST create)
duplicate:{user_id}:{data_hash}
```


mazur-snf (approved):
LGTM

PR #961, opened 2026-04-21
[Backmerge][OSDEV-2529] Fix duplicate throttle logic for production location creation and updates

protsack-stephan:
Updated the duplicate throttle cache key to include the primary key (`pk`) to prevent collisions when the same user makes multiple requests for the same production location. This change ensures that the throttle logic functions correctly and improves the handling of concurrent requests.


vlad-shapik (approved):
LGTM

PR #962, opened 2026-04-21
[OSDEV-2396] Fix edge case in Spotlight partner data validation

protsack-stephan:
Fix `hasPartnerData` logic in `PartnerDataContainer` to cross-reference partner field values against configured `partnerGroups`, preventing the Spotlight section from incorrectly showing the "data available" message when partner fields exist but don't match any known partner group.

Add `partnerGroups` to the `useMemo` dependency array so the check re-evaluates when partner group configuration changes.

Previously, `hasPartnerData` only checked whether `partner_fields` contained any non-empty arrays. This meant that if stale or unrecognized fields were present in the data (fields not belonging to any configured partner group), the component would display the "partner data available" description and attempt to render partner groups - even though no matching groups existed.


vlad-shapik (approved):
LGTM!

PR #964, opened 2026-04-22
[OSDEV-2424] Add partner data fields to data downloads - Part 2

roman-stolar:
**[OSDEV-2424](https://opensupplyhub.atlassian.net/browse/OSDEV-2424) Add partner data fields to data downloads - Part 2**

Added partner-field columns to CSV and XLSX facility downloads. Every `PartnerField` with `active=True` and `system_field=False` is appended after the `is_closed` column, followed by download-only columns for the two system partner fields (MIT Living Wage and **WageIndicator**).
    * **WageIndicator (download-only, 6 columns)** — For each link type on `WageIndicatorCountryData.LinkType` (`living_wage_link_national`, `minimum_wage_link_english`, `minimum_wage_link_national`) one URL column (`wage_indicator.<link_type>`) and one display-text column (`wage_indicator.<link_type>_text`) are projected from `WageIndicatorProvider.raw_values`.
    * Both system-field blocks are synthesized via dedicated download-only helpers (`mit_living_wage_download_helper`, `wage_indicator_download_helper`) wired through `FacilityDownloadSerializerBase.get_mit_living_wage_*` / `get_wage_indicator_*`.

[OSDEV-2424]: https://opensupplyhub.atlassian.net/browse/OSDEV-2424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/serializers/facility/wage_indicator_download_helper.py
```diff
@@ -0,0 +1,65 @@
...
+        self.__provider = WageIndicatorProvider()
+
+    def get_cells(self, facility) -> List[str]:
+        '''
+        Return the six ordered cells matching
+        `WAGE_INDICATOR_DOWNLOAD_HEADERS`, or six empty strings if the
+        provider has no data for this facility.
+        '''
+        try:
+            data = self.__provider.fetch_data(facility)
```
protsack-stephan, line 51: We probably need to do some performance optimization here. As currently designed, we’ll be querying contributor data and extended fields for every row. For a 5,000 - location download with a pagination size of 50, this results in 10,100 queries per page - potentially increasing by another 10,000 when accounting for the Living Wage provider. Should we look into caching these responses in Memcached to mitigate the load? Also, maybe we could make it so that we don't have to query the contributor data as we don't really need it in the data download.

This can be merged as is and we can do the caching in the followup PR.

protsack-stephan (approved):
Good work!

protsack-stephan (approved):

vlad-shapik (approved):

PR #967, opened 2026-04-23
[OSDEV-2525] Fix UserProfileFacilities queryset return when no partner fields are present

protsack-stephan:
Updated the UserProfileFacilities view to return an empty queryset instead of the base queryset when no partner fields are provided. This change ensures that the API behaves correctly by not returning unintended results in such cases.


vlad-shapik (approved):
LGTM

roman-stolar (approved):

PR #977, opened 2026-04-28
Add dockerized Playwright updater for WageIndicator admin links.

vlad-shapik:
This introduces a small standalone automation tool to apply targeted WageIndicatorCountryData link updates from CSV with assignee and status filtering, plus quick Docker run instructions for safe dry-runs and execution.


vlad-shapik:
No longer relevant

PR #978, opened 2026-04-28
[Hotfix-v.2.22.2][OSDEV-2563] Fix partner fields data source in download serializer

protsack-stephan:
Fix `FacilityDownloadSerializer.get_partner_fields_row` to use the unfiltered `facility.extended_fields` instead of `extended_fields_raw`, and add a 10-minute cache to the `all_contributors` view for improved performance.

The `get_partner_fields_row` call was incorrectly receiving `extended_fields_raw`, which filters out extended fields from non-claim contributors via `check_contributor`. Partner fields should consider all extended fields regardless of claim status, so the call now uses `facility.extended_fields` directly - matching how `get_mit_living_wage_row` and `get_wage_indicator_row` already work in the same method.

The `all_contributors` endpoint returns a simple list of contributor IDs, names, and types that changes infrequently. Adding `@cache_page(60 * 10, cache="view_cache")` avoids hitting the database on every request, reducing load on a frequently-called endpoint.

### DIFF
--- a/doc/release/RELEASE-NOTES.md
+++ b/doc/release/RELEASE-NOTES.md
@@ -3,6 +3,22 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html). The format is based on the `RELEASE-NOTES-TEMPLATE.md` file.
 
+## Release 2.22.2
+
+## Introduction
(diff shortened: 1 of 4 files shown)


roman-stolar (approved):
LGTM

vlad-shapik (approved):
LGTM

PR #985, opened 2026-05-01
[OSDEV-2547] Create automatic feedback in SLC submission form

jamesjohnson4395:
Refactor field requirements for the SLC submission form in `src/react/src/util/util.js` to more easily add new tests in the future. Add requirements for non-Latin characters and for values to fall under a maximum length (currently 200 characters).

Add warning dialogue in `src/react/src/components/Contribute/ProductionLocationInfo.jsx` to display a bypass-able pop-up when users click submit if certain tests are not met. Add warning dialogues for PO Boxes and for values under a minimum length (currently 25 characters) in the address field.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes SLC form validation rules (Latin-only/max length) and alters the submit flow to gate submission behind new warning dialogs, which could block or change user submissions if edge cases are missed.
> 
> **Overview**
> Adds a configurable validation registry for the SLC submission schema, introducing **Latin-character-only** and **max-length (200)** checks across key text fields and Latin-only validation for select-list array fields, while keeping number-of-workers format/range validation.
> 
> Updates the production location contribution form to show a bypassable **pre-submit warning dialog** (new `ContributionWarningDialog`) for suspicious addresses (PO Box or too short, via new `containsPOBox`/`isShortAddress` helpers and `MIN_ADDRESS_LENGTH` constraint), and improves Formik select UX by marking `locationType`/`processingType` as touched and rendering their validation errors consistently.
> 
> Extends test coverage with a dedicated `slcValidationSchema` test suite and updates an existing test assertion for the trimmed-text error message.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 312ac5c4eaf015f7c5a8a55218d8f877df5c0713. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/react/src/__tests__/utils/slcValidationSchema.test.js
+++ b/src/react/src/__tests__/utils/slcValidationSchema.test.js
@@ -1,0 +1,248 @@
+import { slcValidationSchema } from '../../util/util';
+
+// Minimal valid base values — only name, address, and country are required.
+// Optional string fields use undefined so Yup skips their tests (empty string
+// fails the meaningful-characters check for non-required slcTextFieldValidation
+// fields, and fails the format-and-range check for numberOfWorkers).
+const validBase = {
+    name: 'Test Facility',
+    address: '123 Main Street',
+    country: { value: 'US', label: 'United States' },
+    sector: [],
+    productType: [],
+    locationType: [],
+    processingType: [],
+    numberOfWorkers: undefined,
+    parentCompany: undefined,
+};
+
+// Helper: validate a partial override against the full schema.
+const validate = (overrides = {}) =>
+    slcValidationSchema.validate(
+        { ...validBase, ...overrides },
+        { abortEarly: true },
+    );
+
+const isValid = (overrides = {}) =>
+    slcValidationSchema.isValid({ ...validBase, ...overrides });
+
+describe('slcValidationSchema — non-Latin character validation', () => {
+    // -------------------------------------------------------------------------
+    // name
+    // -------------------------------------------------------------------------
+    describe('name field', () => {
+        it('accepts plain ASCII text', async () => {
+            await expect(isValid({ name: 'Acme Factory' })).resolves.toBe(true);
+        });
+
+        it('accepts accented Latin characters (é, ü, ñ, ø)', async () => {
+            await expect(
+                isValid({ name: 'Ärger über niño café' }),
+            ).resolves.toBe(true);
+        });
+
+        it('rejects Chinese characters', async () => {
+            await expect(
+                validate({ name: '工厂名称' }),
+            ).rejects.toThrow('must contain only Latin characters');
+        });
+
+        it('rejects Cyrillic characters', async () => {
+            await expect(
+                validate({ name: 'Фабрика' }),
+            ).rejects.toThrow('must contain only Latin characters');
+        });
+
+        it('rejects Arabic characters', async () => {
+            await expect(
+                validate({ name: 'مصنع' }),
+            ).rejects.toThrow('must contain only Latin characters');
+        });
+
+        it('rejects a mixed Latin and non-Latin string', async () => {
+            await expect(
+                validate({ name: 'Acme 工厂' }),
+            ).rejects.toThrow('must contain only Latin characters');
+        });
+    });
+
+    // -------------------------------------------------------------------------
+    // address
+    // -------------------------------------------------------------------------
+    describe('address field', () => {
+        it('accepts a standard Latin address', async () => {
+            await expect(
+                isValid({ address: '10 Downing Street, London' }),
+            ).resolves.toBe(true);
+        });
+
+        it('accepts accented Latin characters in address', async () => {
+            await expect(
+                isValid({ address: '12 Rue de l\'Église, Montréal' }),
+            ).resolves.toBe(true);
+        });
+
+        it('rejects Chinese characters in address', async () => {
+            await expect(
+                validate({ address: '北京市朝阳区' }),
+            ).rejects.toThrow('must contain only Latin characters');
+        });
+
+        it('rejects Arabic characters in address', async () => {
+            await expect(
+                validate({ address: 'شارع الملك فهد' }),
+            ).rejects.toThrow('must contain only Latin characters');
+        });
+    });
+
+    // -------------------------------------------------------------------------
+    // parentCompany
+    // -------------------------------------------------------------------------
+    describe('parentCompany field', () => {
+        it('accepts a Latin parent company name', async () => {
+            await expect(
+                isValid({ parentCompany: 'Global Holdings Ltd.' }),
+            ).resolves.toBe(true);
+        });
+
+        it('accepts accented Latin characters in parent company', async () => {
+            await expect(
+                isValid({ parentCompany: 'Société Générale' }),
+            ).resolves.toBe(true);
+        });
+
+        it('rejects Japanese characters in parent company', async () => {
+            await expect(
+                validate({ parentCompany: '株式会社' }),
+            ).rejects.toThrow('must contain only Latin characters');
+        });
+
+        it('rejects Cyrillic characters in parent company', async () => {
+            await expect(
+                validate({ parentCompany: 'Компания' }),
+            ).rejects.toThrow('must contain only Latin characters');
+        });
+    });
+
+    // -------------------------------------------------------------------------
+    // numberOfWorkers
+    // -------------------------------------------------------------------------
+    describe('numberOfWorkers field', () => {
+        it('accepts standard ASCII digits', async () => {
+            await expect(isValid({ numberOfWorkers: '500' })).resolves.toBe(
+                true,
+            );
+        });
+
+        it('accepts a valid range with ASCII digits', async () => {
+            await expect(
+                isValid({ numberOfWorkers: '100-500' }),
+            ).resolves.toBe(true);
+        });
+
+        it('rejects fullwidth (Japanese) digits', async () => {
+            await expect(
+                validate({ numberOfWorkers: '１００' }),
+            ).rejects.toThrow('Only Latin characters are accepted');
+        });
+
+        it('rejects Arabic-Indic digits', async () => {
+            await expect(
+                validate({ numberOfWorkers: '١٠٠' }),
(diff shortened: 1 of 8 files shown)


src/react/src/util/util.js
```diff
@@ -1743,80 +1742,201 @@ const isCleanValueMeaningful = value => {
...
+// Text-field tests receive the raw string value (possibly null/undefined).
+// Array-field tests receive the full array of { label, value } objects.
+// Message functions receive Yup params — `label` is set via .label() below.
+
+const SLC_TEXT_FIELD_TESTS = Object.freeze({
+    'is-trimmed': Object.freeze({
+        message: 'Remove spaces at start and end of text.',
+        test: value => value == null || value === value.trim(),
+    }),
+    'not-a-number': Object.freeze({
```
ColumbusLeonardOSHub: This name is a bit confusing -- it could be saying "the field needs to be a number" or "the field can't be a number". I recommend renaming for clarity.

src/react/src/util/util.js
```diff
@@ -1743,80 +1742,201 @@ const isCleanValueMeaningful = value => {
...
+const SLC_FIELD_VALIDATION_CONFIG = Object.freeze({
+    name: Object.freeze({
+        type: 'text',
+        tests: [
+            'is-trimmed',
+            'not-a-number',
+            'meaningful-characters',
+            'latin-characters-only',
+        ],
+    }),
```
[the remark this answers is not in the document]
ColumbusLeonardOSHub: @jamesjohnson4395 , I think this needs to be addressed.

ColumbusLeonardOSHub (changes requested):

src/react/src/util/util.js
jamesjohnson4395: Updated test name to 'valid-non-number' to reflect that the field is valid when it is not a number.

src/react/src/util/util.js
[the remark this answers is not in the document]
jamesjohnson4395: I originally built this test to check all 3 fields, but updated it to reflect contribot only checking the address field for maximum length. It's probably a good idea to check the name and parent company fields as well, though, so I've added the test back to these fields!

src/react/src/util/util.js
```diff
@@ -1743,80 +1742,203 @@ const isCleanValueMeaningful = value => {
...
+
+const SLC_TEXT_FIELD_TESTS = Object.freeze({
+    'is-trimmed': Object.freeze({
+        message: 'Remove spaces at start and end of text.',
+        test: value => value == null || value === value.trim(),
+    }),
+    'valid-non-number': Object.freeze({
+        message: ({ label }) => `${label} cannot be a number.`,
+        test: value => value == null || notANumber(value),
+    }),
```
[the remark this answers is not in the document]
jamesjohnson4395: The name was updated to address the previous comment, and the PR discussion request was resolved.

src/react/src/util/util.js
```diff
@@ -1743,80 +1742,203 @@ const isCleanValueMeaningful = value => {
...
+            value == null ||
+            value.length <= SLC_FORM_CONSTRAINTS.MAX_STRING_LENGTH,
+    }),
+    'valid-format-and-range': Object.freeze({
+        message:
+            'Enter a single positive number (e.g., 5) or a valid range ' +
+            '(e.g., 3–10). In a range, the minimum value must be less ' +
+            'than or equal to the maximum, and both must be at least 1.',
+        test: value => value == null || validNumWorkersFormat(value),
+    }),
```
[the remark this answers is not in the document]
ColumbusLeonardOSHub: @jamesjohnson4395 , have you seen this comment?

ColumbusLeonardOSHub:
@jamesjohnson4395 , what support do you need to merge this by eod wednesday?

src/react/src/util/util.js
```diff
@@ -1743,80 +1742,203 @@ const isCleanValueMeaningful = value => {
...
-            'valid-format-and-range',
-            'Enter a single positive number (e.g., 5) or a valid range ' +
-                '(e.g., 3–10). In a range, the minimum value must be less ' +
-                'than or equal to the maximum, and both must be at least 1.',
-            value => {
-                if (value == null) return true;
+const containsOnlyLatinCharacters = value => {
+    if (typeof value !== 'string') return true;
+    const normalized = value.normalize('NFC');
+    return !/[\u0250-\u1FFF\u2070-\u2E7F\u2E80-\uFFFF]/g.test(normalized);
```
VadimKovalenkoSNF: The regex range `\u0250–\u1FFF` includes Latin Extended Additional (`\u1E00–\u1EFF`),
which covers ~500 valid precomposed Latin characters used in Vietnamese
(e.g. `ắ = U+1EAF`). "Đắk Lắk" is rejected as non-Latin right now.

Polish is unaffected — ą, ć, ę, ł, ń, ś, ź, ż are all below U+0250.

Suggested fix: replace the range-based test with Unicode property escapes:
  `/^[\p{Script=Latin}\p{M}\p{N}\p{P}\s]+$/u`
This approach is future-proof and won't break for any other Latin-based script.

src/react/src/util/util.js
```diff
@@ -1743,80 +1742,203 @@ const isCleanValueMeaningful = value => {
...
+        test: value => value == null || value === value.trim(),
+    }),
+    'valid-non-number': Object.freeze({
+        message: ({ label }) => `${label} cannot be a number.`,
+        test: value => value == null || notANumber(value),
+    }),
+    'meaningful-characters': Object.freeze({
+        message: ({ label }) =>
+            `${label} cannot contain only spaces or symbols.`,
+        test: value => value == null || isCleanValueMeaningful(value),
```
VadimKovalenkoSNF: The guard is `value == null`, but optional fields (`parentCompany`, `numberOfWorkers`)
are initialised as '' in the form. Empty string is `not null/undefined`, so it
reaches `isCleanValueMeaningful('') → false`, immediately invalidating the form
on page load and keeping the Submit button disabled.

The test file works around this by using `undefined` instead of `''` (see
slcValidationSchema.test.js lines 16–17), which masks the real bug.

Fix options:
  1. Extend the guard using Lodash method: `isEmpty(value) || isCleanValueMeaningful(value)`
  2. Use `Yup's .transform()` in buildFieldValidation to coerce `''` → undefined
     before tests run.

src/react/src/util/util.js
```diff
@@ -1743,80 +1742,203 @@ const isCleanValueMeaningful = value => {
...
+        test: value =>
+            value == null ||
+            value.length <= SLC_FORM_CONSTRAINTS.MAX_STRING_LENGTH,
+    }),
+    'valid-format-and-range': Object.freeze({
+        message:
+            'Enter a single positive number (e.g., 5) or a valid range ' +
+            '(e.g., 3–10). In a range, the minimum value must be less ' +
+            'than or equal to the maximum, and both must be at least 1.',
+        test: value => value == null || validNumWorkersFormat(value),
```
VadimKovalenkoSNF: `validNumWorkersFormat('')` returns false (neither pattern matches).
The guard `value == null` does not cover the empty-string initial state of the
`numberOfWorkers field`. Same fix as the meaningful-characters guard above.

src/react/src/__tests__/utils/slcValidationSchema.test.js
```diff
@@ -0,0 +1,288 @@
...
+// fails the format-and-range check for numberOfWorkers).
+const validBase = {
+    name: 'Test Facility',
+    address: '123 Main Street',
+    country: { value: 'US', label: 'United States' },
+    sector: [],
+    productType: [],
+    locationType: [],
+    processingType: [],
+    numberOfWorkers: undefined,
```
VadimKovalenkoSNF, line 16: Using `undefined` here masks a real bug: the actual form initialises these
fields as `''`, not `undefined`. A test that reflects the real initial state
(empty string) would fail today and make the bug visible.

Consider adding a dedicated test case:
```
  it('accepts an empty optional field (real initial value)', async () => {
      await expect(isValid({ parentCompany: '', numberOfWorkers: '' })).resolves.toBe(true);
  });
```

jamesjohnson4395:
> @jamesjohnson4395 , what support do you need to merge this by eod wednesday?

@ColumbusLeonardOSHub I'll be implementing Vadim's comments from this morning and should be able to merge after that!

src/react/src/__tests__/utils/slcValidationSchema.test.js
```diff
@@ -0,0 +1,288 @@
...
+
+// Helper: build an array of N valid { label, value } items.
+const makeItems = count =>
+    Array.from({ length: count }, (_, i) => ({
+        label: `Item ${i + 1}`,
+        value: `item-${i + 1}`,
+    }));
+
+describe('slcValidationSchema', () => {
+    // -------------------------------------------------------------------------
```
VadimKovalenkoSNF: I don't think those comments are needed. You have `describe('name field'...` for that case.

src/react/src/components/Contribute/ProductionLocationInfo.jsx
```diff
@@ -388,11 +414,25 @@ const ProductionLocationInfo = ({
         contributionForm.validateField('country');
```
VadimKovalenkoSNF, line 414: This useEffect manually re-triggers Formik validation after chip removal in
react-select (documented in the comment above). The new `locationType` and
`processingType` fields use the same StyledSelect pattern but are not included
here. A stale error can linger after a chip is removed, keeping Submit disabled
when it shouldn't be.

Add both fields to the effect:
```
  useEffect(() => {
      contributionForm.validateField('productType');
      contributionForm.validateField('country');
      contributionForm.validateField('locationType');
      contributionForm.validateField('processingType');
  }, [
      contributionForm.values.productType,
      contributionForm.values.country,
      contributionForm.values.locationType,
      contributionForm.values.processingType,
  ]);
```

VadimKovalenkoSNF (changes requested):
I left few comments, pls check.
Also, make sure that this change is reflected in `doc/release/RELEASE-NOTES.md`

VadimKovalenkoSNF (approved):
Approved.

PR #989, opened 2026-05-07
[OSDEV-2659] Fix download limit handling for API users and harden Stripe webhook

protsack-stephan:
Resolve the discrepancy where API users who purchased additional data downloads via the UI checkout never received their credits, requiring manual admin intervention.

**Problem (from [OSDEV-2659](https://opensupplyhub.atlassian.net/browse/OSDEV-2659)):** The UI enforces a 5,000-record download limit for all authenticated users (including API users) and redirects them to Stripe checkout when they exceed it. However, the backend assumed API users have unlimited UI downloads and returned `None` for their download limit — which meant the Stripe webhook callback was ignored and purchased credits were never applied. A paying API user purchased $600 worth of downloads but was still capped at 5,000 records until support manually updated the limit via Django admin.

**Solution (Option 2 from the ticket):** Update the backend to match the UI behavior — API users are now subject to standard UI data download limits and can properly receive purchased credits through Stripe checkout.

Changes:

- **`FacilitiesDownloadService.get_download_limit`** — Return `None` early when the request carries an API token (`request.auth is not None`), so API-token requests bypass the data download limit (they're counted against their API quota instead). This moves the API-user check from the model to the service layer where the request context is available.
- **`FacilityDownloadLimit.get_or_create_user_download_limit`** — Simplified to only skip anonymous users. The previous `is_api_user` check here was causing the webhook to fail when trying to update the download limit after a successful purchase.
- **`DownloadLocationsCheckoutWebhookView`** — Two fixes:
  1. Use `get_or_create` instead of `get` for `FacilityDownloadLimit`, so the limit record is created automatically if it doesn't exist yet (which happens for API users making their first UI purchase).
  2. Check for existing `DownloadLocationPayment` before processing to handle duplicate webhook deliveries from Stripe — returns 200 without double-counting.
- **Tests** — Added test for duplicate webhook idempotency: verifies a second delivery returns 200, creates only one payment record, and doesn't inflate `paid_download_records`.

[OSDEV-2659]: https://opensupplyhub.atlassian.net/browse/OSDEV-2659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/release/RELEASE-NOTES.md
+++ b/doc/release/RELEASE-NOTES.md
@@ -25,7 +25,6 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 * Ensure that the following commands are included in the `post_deployment` command:
     * `migrate`
     * `reindex_database`
-    * `reindex_locations_with_approved_claim`
     * `backfill_moderation_event_os_id`
 
 
--- a/doc/release/RELEASE-NOTES.md
+++ b/doc/release/RELEASE-NOTES.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Introduction
 * Product name: Open Supply Hub
-* Release date: May 9, 2026
+* Release date: May 15 2026
 
 ### What's new
 * [OSDEV-1227](https://opensupplyhub.atlassian.net/browse/OSDEV-1227) - Replaced "Rejected" with "Feedback Phase" on user-facing list status pages (My Lists table, list detail page, and status summary message) to use more welcoming language.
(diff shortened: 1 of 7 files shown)


src/django/api/views/stripe/download_locations_checkout_webhook_view.py
```diff
@@ -43,6 +46,17 @@ def post(self, request, *args, **kwargs):
...
+                payment_exists = DownloadLocationPayment.objects.filter(
+                    stripe_session_id=stripe_session_id
+                ).exists()
+
+                if payment_exists:
+                    logger.info(
+                        f"Payment exists for session: {stripe_session_id}",
+                    )
+                    return HttpResponse(status=200)
+
```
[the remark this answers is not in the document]
protsack-stephan: This is true on one hand, but on another hand wrapping it onto the transaction will make this webhook hard to debug. Right now if user reaches out to us with an issue that their balance was not updated, we know exactly where to look and what could have happened within the applicaton.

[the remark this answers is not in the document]
protsack-stephan: Updated the code to place the API call on the top.

valencia-max (dismissed):

doc/release/RELEASE-NOTES.md
```diff
@@ -33,6 +33,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Bugfix
 * [OSDEV-2563](https://opensupplyhub.atlassian.net/browse/OSDEV-2563) - Fix partner fields in facility download to use unfiltered extended fields, and add 10-minute cache to the all_contributors endpoint.
 
+### Code/API changes
+* [OSDEV-2659](https://opensupplyhub.atlassian.net/browse/OSDEV-2659) - Updated download limits logic: exempt API-token requests from data download limits, add idempotent duplicate webhook handling, and use `get_or_create` for `FacilityDownloadLimit` in the checkout webhook.
+
```
vlad-shapik: Probably it should be in the 2.23.0 version

vlad-shapik (changes requested):

doc/release/RELEASE-NOTES.md
protsack-stephan: Good catch, will update.

protsack-stephan: Updated.

vlad-shapik (approved):
LGTM

PR #990, opened 2026-05-07
Cache is all you need

protsack-stephan:
The home page and facility search endpoints are among the most-hit paths in production, yet every request was hitting the database directly. This PR adds Memcached-backed response caching (`cache_page`) to seven read-heavy views and centralizes the timeout behind a single configurable environment variable (`MEMCACHED_VIEW_CACHE_TIMEOUT_SECONDS`, defaults to 600 seconds).

Cached views:
- `all_contributors`
- `ContributorFacilityListSortedViewSet.list`
- `ContributorFacilityListViewSet.list`
- `FacilitiesViewSet.list`
- `FacilitiesViewSet.retrieve`
- `UserProfileFacilities`
- `UserProfileFacilityLists`

Previously, `UserProfileFacilities` and `UserProfileFacilityLists` used hardcoded 5-minute timeouts and `all_contributors` used a hardcoded 10-minute timeout. All views now read from the shared `settings.MEMCACHED_VIEW_CACHE_TIMEOUT_SECONDS` setting, which is configurable per-environment via Terraform without a code change.

Infrastructure changes:
- New Terraform variable `memcached_view_cache_timeout_seconds` (default 600)
- Passed as `MEMCACHED_VIEW_CACHE_TIMEOUT_SECONDS` env var in the ECS app task definition
- Django settings reads the env var with a 600s fallback


src/django/api/views/contributor/contributor_facility_list_view_set.py
```diff
@@ -25,6 +27,10 @@ class ContributorFacilityListViewSet(ReadOnlyModelViewSet):
         type=TYPE_INTEGER,
         required=False
     )], responses={200: ''}, deprecated=True)
+    @cache_page(
+        settings.MEMCACHED_VIEW_CACHE_TIMEOUT_SECONDS,
+        cache="view_cache",
+    )
```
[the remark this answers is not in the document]
protsack-stephan: are you sure? I see that it's working locally]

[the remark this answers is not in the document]
protsack-stephan: But locally I have active Memcached running in docker. Are you absolutely sure?

src/django/api/views/user/user_profile_facilities.py
```diff
@@ -28,7 +29,13 @@ class FacilityIndexCursorPagination(CursorPagination):
     max_page_size = 20
 
 
-@method_decorator(cache_page(60 * 5, cache="view_cache"), name="dispatch")
+@method_decorator(
```
valencia-max, line 32: Previously the value for cache was `60 * 5`, now the default value is `60 * 10`. It is okay to increase 2x this value?

src/django/api/views/user/user_profile_facility_lists.py
```diff
@@ -19,7 +20,13 @@ class FacilityListCursorPagination(CursorPagination):
     max_page_size = 100
 
 
-@method_decorator(cache_page(60 * 5, cache="view_cache"), name="dispatch")
```
valencia-max, line 22: Same here

valencia-max (dismissed):
Left a question to better undestand this code, LGTM!

src/django/api/views/user/user_profile_facilities.py
protsack-stephan: Good catch. I’m overwriting this at the infrastructure level, also it can be modified for the local environment using an environment variable.

src/django/api/views/user/user_profile_facility_lists.py
protsack-stephan: Good catch. I’m overwriting this at the infrastructure level, also it can be modified for the local environment using an environment variable.

VadimKovalenkoSNF (approved):
Looks clear, approved!

vlad-shapik (approved):
LGTM

PR #991, opened 2026-05-07
Increase RDS work_mem from 20 MB to 64 MB to eliminate disk spilling during PostGIS queries

protsack-stephan:
Increase the PostgreSQL work_mem parameter from 20000 (≈19.5 MB) to 65536 (64 MB) in the RDS parameter group to resolve performance degradation caused by excessive temporary file I/O during PostGIS spatial operations.

During traffic spikes, the database was writing temporary files (8.5–11.7 MB each) to disk for spatial aggregation operations like ST_AsMVT (vector tile generation). Although these files appear smaller than the 19.5 MB limit, in-memory data structures (hash tables, sorting trees) consume 2–3x more space than their serialized on-disk representation. This caused PostgreSQL to spill to disk repeatedly, and when many concurrent queries hit simultaneously, disk I/O was overwhelmed.

### DIFF
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -291,7 +291,7 @@ variable "rds_cpu_credit_balance_threshold" {
 }
 
 variable "rds_work_mem" {
-  default = "20000"
+  default = "65536"
 }
 
 variable "rds_deletion_protection" {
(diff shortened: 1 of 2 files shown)


valencia-max (approved):

VadimKovalenkoSNF (approved):
LGTM

vlad-shapik (approved):
LGTM

PR #992, opened 2026-05-08
[OSDEV-2589] Apply JSON Schema defaults to partner field values in API responses and CSV downloads

protsack-stephan:
Partner Grievance Mechanism fields (Labor Solutions, Accord, Ulula) were appearing blank in data downloads and API responses when the underlying data didn't explicitly store default values defined in the field's JSON Schema. This made the dataset look incomplete despite the schema specifying what those defaults should be.

This PR introduces `apply_schema_defaults`, a recursive function that walks a JSON Schema's `properties` and fills in missing keys using their declared `default` values. It is applied in two places:
- **Production Locations API** (`ProductionLocations` view) — defaults are applied to `raw_values` before returning the response.
- **CSV download serializer** (`FacilityDownloadSerializerBase`) — defaults are applied to deep-copied entries before building CSV rows, so the stored data is never mutated.

The view was also refactored to replace the manual `PARTNER_FIELD_NAMES_KEY` cache with the existing `get_cached_all_partner_fields()` helper, removing a redundant cache layer and the now-unused constant. The former `__add_partner_field_value` static method was replaced by `__get_partner_field_value`, which returns the processed value instead of mutating a dict, making the control flow easier to follow.


vlad-shapik (approved):
LGTM

VadimKovalenkoSNF (approved):
Approved.

PR #994, opened 2026-05-08
[OSDEV-2591] Fix missing extended fields in downloads for claimed facilities

protsack-stephan:
Remove contributor-based filtering from `FacilityDownloadSerializer` so that `processing_type_facility_type_raw`, `facility_type`, `processing_type`, and `product_type` populate in downloads even when the facility is claimed.

The `get_extended_fields_raw` method (introduced over 2 years ago) excluded extended fields from non-claimant contributors whenever a facility had an approved claim. This caused a data discrepancy: fields visible on the location profile page (e.g. `BD2021113R7R87P`) appeared blank in the downloaded file.

This PR removes both `get_extended_fields_raw` and its helper `check_contributor`, and passes `facility.extended_fields` directly to `get_extended_fields` - matching the behavior already used by `get_partner_fields_row` on the same row.


vlad-shapik (approved):
LGTM

VadimKovalenkoSNF (approved):
Approved.

PR #995, opened 2026-05-08
[OSDEV-2654] Fix SLC dashboard UPDATE bug

jamesjohnson4395:
For all UPDATE SLC moderation events, if the intended existing OSID is not returned by the API call for potential matches, the existing OSID is manually added to the list. For all UPDATE requests, the existing OSID is moved to the top of the potential matches list.

### DIFF
--- a/src/react/src/actions/dashboardContributionRecord.js
+++ b/src/react/src/actions/dashboardContributionRecord.js
@@ -3,6 +3,7 @@ import apiRequest from '../util/apiRequest';
 import {
     makeModerationEventRecordURL,
     makeGetProductionLocationsForPotentialMatches,
+    makeProductionLocationURL,
     makeProductionLocationFromModerationEventURL,
     logErrorAndDispatchFailure,
 } from '../util/util';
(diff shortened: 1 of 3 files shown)


vlad-shapik (approved):
LGTM

vlad-shapik (changes requested):
Could you please update RELEASE-NOTES.md to include these changes as part of the 2.23.0 release?

VadimKovalenkoSNF (commented):
**Design concern: `pinOsIdToTop` has dual responsibility and a misleading name**

The function name `pinOsIdToTop` implies a pure reordering operation, but it actually does two very different things depending on whether the `osId` is already present in the matches:

1. **If `osId` is in the list** — reorders in-place, no network call.
2. **If `osId` is missing** — makes an additional API request to `GET /api/v1/production-locations/{osId}/` and prepends the result.

The name hides the fact that a network call is made at all. This is exactly what makes the function look redundant at first glance ("why re-fetch just to sort?") — and it will confuse future readers the same way.

A few suggestions:
- Rename to something like `ensureOsIdPinnedToTop` or `fetchAndPinOsId` to signal the potential side-effect.
- Or split into two functions: one that reorders (pure), and one at the `fetchPotentialMatches` level that decides whether to call the detail endpoint.
- Add a short JSDoc comment explaining the two code paths and why the secondary fetch exists (the fuzzy search endpoint may not return the existing location for UPDATE events).

VadimKovalenkoSNF (approved):
Approved. Left optional comment.

src/react/src/actions/dashboardContributionRecord.js
```diff
@@ -93,40 +94,70 @@ export function fetchSingleModerationEvent(moderationID) {
...
-                        failFetchPotentialMatches,
-                    ),
+            );
+
+            if (potentialMatches.data) {
+                const responseData = osId
+                    ? await pinOsIdToTop(potentialMatches.data, osId)
+                    : potentialMatches.data;
+
+                dispatch(completeFetchPotentialMatches(responseData));
```
protsack-stephan, line 150: Are we sure we want to do this before we put the data into the reducer? Maybe it makes more sense to perform this operation inside the reducer? Also, we are skewing the response data a bit by moving things around. Maybe it makes sense to add an additional piece to the UI that will show the OS ID match and allow you to accept it. This would also keep the original order of potential matches from the API in case the user submits data that is not related to this OS ID.

protsack-stephan (commented):
Good work! A comment from me.

src/react/src/actions/dashboardContributionRecord.js
jamesjohnson4395: A second implementation option based off of this comment has been created in a separated PR here: https://github.com/opensupplyhub/open-supply-hub/pull/1005

Only one of these PRs should be merged, and the other should be denied.

PR #997, opened 2026-05-12
[OSDEV-2360] Backfill os_id into OpenSearch for historical moderation events

daniellekain:
Add a management command to repair historical approved moderation events that are missing `os_id` in OpenSearch.

## Background

The post-save signal for `ModerationEvent` was previously writing the field to OpenSearch as `"os"` instead of `"os_id"` (fixed in #986). Logstash normally masked the bug by doing full document replacements on its schedule — but when Logstash fell out of sync (pipeline restarts, index rebuilds), events approved in that window ended up permanently missing `os_id` in OpenSearch. Since Logstash's watermark-based sync only catches records updated after the last run, those historical events were never corrected.

## Changes

**`api/management/commands/backfill_moderation_event_os_id.py`** — New management command that scrolls OpenSearch to find approved events missing `os_id`, fetches the correct `os_id` from PostgreSQL, and patches only those documents. A previous iteration of this command updated all approved events unconditionally, which timed out in staging due to nearly 300k records. This version uses the OpenSearch Scroll API to target only the affected subset, then batches the PostgreSQL lookups and OpenSearch updates in groups of 100. Events where `os_id` is also absent in the DB are skipped. Supports `--dry-run` to preview the count before writing.

**`api/management/commands/post_deployment.py`** — Wires the backfill into the post-deployment step so it runs automatically on each deploy until all affected records are resolved (it exits quickly once there is nothing left to fix).

## Testing

Tested locally:
- Created a test `ModerationEvent` in the DB with `os_id` set, indexed it in OpenSearch without `os_id` to reproduce the bug
- `--dry-run` correctly identified it as needing backfill
- Live run updated the OpenSearch document; confirmed `os_id` appeared in the document afterward
- Events with no `os_id` in the DB were correctly skipped

Ticket: OSDEV-2360

### DIFF
--- a/src/django/api/management/commands/backfill_moderation_event_os_id.py
+++ b/src/django/api/management/commands/backfill_moderation_event_os_id.py
@@ -5,6 +5,9 @@ from api.models.moderation_event import ModerationEvent
 from api.services.opensearch.opensearch import OpenSearchServiceConnection
 from api.views.v1.index_names import OpenSearchIndexNames
 
+BATCH_SIZE = 100
+SCROLL_TIMEOUT = '2m'
+
 
 class Command(BaseCommand):
     help = (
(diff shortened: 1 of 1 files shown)


VadimKovalenkoSNF (approved):
LGTM

PR #1001, opened 2026-05-13
OSDEV-2416: [Embedded Map] Percentage formatting issue

valencia-max:
## Summary of Changes
OSDEV-2416 — [Embedded Map] Percentage formatting issue

- Bug: Percent-formatted columns were detected by inspecting only row 2. If row 2's cell was blank, the entire column was treated as non-percent and stored as raw decimals (e.g. 0.5 instead of 50%).
- Fix: Removed the row-2 column scan and get_column_letter import. __format_row now checks cell.number_format per cell, applying __format_percent only when % is present.
- Null handling: __format_percent now returns '' for None (previously coerced through the string branch).

### Testing
- Tests: Added test_percent_formatting_detected_per_cell_not_only_from_row_2 — blank row-2 cell stays '', while 0.5 → '50%' and 0.75 → '75%' in subsequent rows.

- As you can see, executing this query in postgres console:
`select name,address,raw_json from api_facilitylistitem order by created_at desc limit 1;`
it returns this data, meaning that the data was processed correctly:

image

Facility that was uploaded:
image

image

### Known TODO Items
N/A

### Checklist
- [x] I have performed a self-review on my PR and am presenting my best work for my peers to review.

### DIFF
--- a/doc/release/RELEASE-NOTES.md
+++ b/doc/release/RELEASE-NOTES.md
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Architecture/Environment changes
 
 ### Bugfix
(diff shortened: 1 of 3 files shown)


daniellekain (approved):
thank you for making these changes -- I appreciate the test coverage!

daniellekain (dismissed):
thank you for making this change and adding testing!

VadimKovalenkoSNF (dismissed):
Looks good, approved.

daniellekain (approved):
lgtm!

protsack-stephan (approved):

PR #1005, opened 2026-05-14
[OSDEV-2654] Fix SLC dashboard UPDATE bug V2

jamesjohnson4395:
Add an "Existing OSID" section to the SLC dashboard for all UPDATE SLC submissions. This section contains the existing OSID that the UPDATE request was submitted for. The new section should not show up on the dashboard for any CREATE SLC submissions.

This is a second implementation option based on feedback received in the first implementation's PR: https://github.com/opensupplyhub/open-supply-hub/pull/995

Only one of these PRs should be merged, and the other should be denied.


daniellekain:
thanks for making this PR-- James! Would it be possible to add a few screenshots of the UI before and after?

jamesjohnson4395:
> thanks for making this PR-- James! Would it be possible to add a few screenshots of the UI before and after?

Yes, thanks for the suggestion! I've included screenshots of the previous and new UI in the ticket: https://opensupplyhub.atlassian.net/browse/OSDEV-2654

src/react/src/components/Dashboard/DashboardContributionRecord.jsx
```diff
@@ -76,18 +78,108 @@ const confirmPotentialMatchButtonDisabled = (classes, osId, matchOsId) => (
     </span>
 );
 
+const renderExistingOsIdContent = (
```
VadimKovalenkoSNF: `moderationEventStatus` is declared as the 6th parameter but is never referenced inside the function body. It should be removed from both the declaration and the call site.

src/react/src/components/Dashboard/DashboardContributionRecord.jsx
```diff
@@ -76,18 +78,108 @@ const confirmPotentialMatchButtonDisabled = (classes, osId, matchOsId) => (
...
+                                );
+                            }}
+                        >
+                            {confirmPotentialMatchButtonTitle}
+                        </Button>
+                    ) : (
+                        <DialogTooltip
+                            text="Moderation event data has been already matched to this production location."
+                            aria-label="Confirm existing OS ID button tooltip"
+                            childComponent={confirmPotentialMatchButtonDisabled(
```
VadimKovalenkoSNF: `confirmPotentialMatchButtonDisabled` renders 'Matched' when `osId === matchOsId`, which is always true here since the same value is passed for both. So a `REJECTED` event would show a "Matched" button label. Worse, the tooltip text says "Moderation event data has been already matched to this production location" — which is factually wrong when the event was rejected, not matched.

src/react/src/__tests__/components/DashboardContributionRecord.test.js
```diff
@@ -318,4 +330,133 @@ describe('DashboardContributionRecord component', () => {
 
     expect(queryByText('Close Dialog')).not.toBeInTheDocument();
   });
+
+  test('calls fetchPotentialMatches without osId for an UPDATE request', async () => {
```
VadimKovalenkoSNF: The tests `calls fetchPotentialMatches without osId for an UPDATE request'`and `calls fetchPotentialMatches without osId for a CREATE request` assert that `fetchPotentialMatches` is called without an osId property. That was already true before this PR — `fetchMatches` never received osId. These tests pass trivially and don't cover any behavior changed here.

src/react/src/__tests__/components/DashboardContributionRecord.test.js
```diff
@@ -318,4 +330,133 @@ describe('DashboardContributionRecord component', () => {
...
+  test('does not call fetchExistingOsIdLocation for a CREATE request', async () => {
+    renderComponent({
+      dashboardContributionRecord: {
+        singleModerationEvent: {
+          data: { ...data, os_id: null, request_type: 'CREATE' },
+        },
+      },
+    });
+
+    await waitFor(() => {
```
VadimKovalenkoSNF, line 445: `waitFor` is designed for eventually-true assertions. A `not.toHaveBeenCalled()` check will pass immediately (including before any async code runs), making it a weak guard.

VadimKovalenkoSNF (changes requested):
Hi, please take at the comments that I left using Cursor-assistance.

Also: The PR introduces a `fetchExistingOsIdLocation` thunk that makes a `GET /api/v1/production-locations/{osId}/` request to display the "Existing OSID" section. But the `os_id` is already available in `singleModerationEventItem` — it's part of the ModerationEventUpdateSerializer response.

Consider these 2 alternatives:

**Option A** (simpler — recommended): Remove the entire Redux async stack (`fetchExistingOsIdLocation`, `startFetchExistingOsIdLocation`, `failFetchExistingOsIdLocation`, `completeExistingOsIdLocation`, the new reducer state, the useEffect, existingOsIdLocationData/Fetching props). Use osId directly from the already-fetched `singleModerationEventItem` to render the "Existing OSID" section — just an OSID link and a Confirm button. All the data you need is right there.

**Option B** (if name/address/claim_status are genuinely needed): Instead of a separate front-end request, extend `ModerationEventUpdateSerializer` to include the existing production location's details (via SerializerMethodField reading from self.os). This keeps it one fetch and avoids the extra network round-trip.

src/react/src/components/Dashboard/DashboardContributionRecord.jsx
jamesjohnson4395: `moderationEventStatus` is now utilized in the function following the Confirm button bug fix

jamesjohnson4395:
> Hi, please take at the comments that I left using Cursor-assistance.
> 
> Also: The PR introduces a `fetchExistingOsIdLocation` thunk that makes a `GET /api/v1/production-locations/{osId}/` request to display the "Existing OSID" section. But the `os_id` is already available in `singleModerationEventItem` — it's part of the ModerationEventUpdateSerializer response.
> 
> Consider these 2 alternatives:
> 
> **Option A** (simpler — recommended): Remove the entire Redux async stack (`fetchExistingOsIdLocation`, `startFetchExistingOsIdLocation`, `failFetchExistingOsIdLocation`, `completeExistingOsIdLocation`, the new reducer state, the useEffect, existingOsIdLocationData/Fetching props). Use osId directly from the already-fetched `singleModerationEventItem` to render the "Existing OSID" section — just an OSID link and a Confirm button. All the data you need is right there.
> 
> **Option B** (if name/address/claim_status are genuinely needed): Instead of a separate front-end request, extend `ModerationEventUpdateSerializer` to include the existing production location's details (via SerializerMethodField reading from self.os). This keeps it one fetch and avoids the extra network round-trip.

Hey Vadim - thank you so much for the detailed review! The name, address, and claim data is needed for moderation, so Option A here would not be preferred. For Option B, I dove into `ModerationEventUpdateSerializer`, but it seems to be more focused for `PATCH` requests when a moderator updates the event. So, I'm not sure if it makes the most sense for the change to live there.

VadimKovalenkoSNF:
Approved. Plese, update PR with the base branch.

src/react/src/components/Dashboard/DashboardContributionRecord.jsx
```diff
@@ -63,31 +65,147 @@ const claimButtonDisabled = classes => (
...
-);
+const renderConfirmButton = (
+    classes,
+    matchOsId,
+    eventOsId,
+    moderationEventStatus,
+    moderationEventFetching,
+    isDisabled,
+    confirmPotentialMatch,
+    ariaLabel,
```
valencia-max: I agree with the below comment, we shouldn't introduce funtions with a lot of parameters, I think you could create a `RenderConfirmButtonSpec` as an interface/class/json, and destruct its properties.

src/react/src/components/Dashboard/DashboardContributionRecord.jsx
```diff
@@ -180,6 +301,12 @@ const DashboardContributionRecord = ({
         }
     }, [productionLocationName, countryCode, productionLocationAddress, osId]);
 
+    useEffect(() => {
+        if (requestType === REQUEST_TYPE_ENUM.UPDATE && osId) {
```
valencia-max: nitpick:

I saw `lodash` as dependency in the react app. I think we can use it for these cases when you want to validate a variable is not null/undefined.




```suggestion
        if (requestType === REQUEST_TYPE_ENUM.UPDATE && !_.isNil(osId)) {
```

valencia-max (dismissed):
Looks good to me!!

Please just update release-notes before merge it.

src/react/src/components/Dashboard/DashboardContributionRecord.jsx
valencia-max: Just don't forget to import it at the top.

VadimKovalenkoSNF (dismissed):
LGTM

src/react/src/components/Dashboard/DashboardContributionRecord.jsx
protsack-stephan: I moved this to a separate component and grouped the arguments to reduce the number of parameters. Thanks for the comment!

src/react/src/components/Dashboard/DashboardContributionRecord.jsx
protsack-stephan: Updated, thanks for the comment!

VadimKovalenkoSNF (approved):
LGMT

vlad-shapik (approved):

PR #1009, opened 2026-05-15
[OSDEV-1940] Embedded Map: It is impossible to hide Facility Type and Processing Type fields, fields are unchecked but still are displayed

valencia-max:
## Summary of Changes
A field is exposed only if BOTH:
- The contributor has at least one ExtendedField entry with that field_name, AND
- The corresponding EmbedField row for this embed config has visible=True (i.e. the user has not unchecked it in the /settings -> Embed page).

Previously, this method ignored condition (2) and returned every ExtendedField name for the contributor. As a result, fields the user had explicitly unchecked (e.g. Facility Type, Processing Type) continued to render in the embedded search form.

See [OSDEV-1940](https://opensupplyhub.atlassian.net/browse/OSDEV-1940).

### Testing

OSDEV-1940

OSDEV-1940-2

https://github.com/user-attachments/assets/f659dba3-2e73-44dc-b2ea-c0ed24d0b9c3

Unit testing:
image

### Known TODO Items
NA

### Checklist
- [x] I have performed a self-review on my PR and am presenting my best work for my peers to review.

[OSDEV-1940]: https://opensupplyhub.atlassian.net/browse/OSDEV-1940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/django/api/serializers/embed_config.py
+++ b/src/django/api/serializers/embed_config.py
@@ -47,12 +47,21 @@ class EmbedConfigSerializer(ModelSerializer):
 
     def get_extended_fields(self, instance):
         try:
-            extended_fields = (
+            contributor_field_names = (
                 ExtendedField
                 .objects
                 .filter(contributor_id=instance.contributor.id)
                 .values_list('field_name', flat=True).distinct()
             )
-            return extended_fields
+            visible_column_names = set(
+                EmbedField
+                .objects
+                .filter(embed_config=instance, visible=True)
+                .values_list('column_name', flat=True)
+            )
+            return [
+                field_name for field_name in contributor_field_names
+                if field_name in visible_column_names
+            ]
         except Contributor.DoesNotExist:
             return []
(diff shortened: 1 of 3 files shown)


src/django/api/serializers/embed_config.py
```diff
@@ -47,12 +47,21 @@ def get_embed_fields(self, instance):
...
+            visible_column_names = set(
+                EmbedField
+                .objects
+                .filter(embed_config=instance, visible=True)
+                .values_list('column_name', flat=True)
+            )
+            return [
+                field_name for field_name in contributor_field_names
+                if field_name in visible_column_names
+            ]
```
protsack-stephan, line 65: Looks good, I wonder if this method used across the codebase in other places. Also, are there any tests that exist for this? Should we add any?

valencia-max: I think we don't have tests for this, I'll create them and will check if this affect another place in the app.

daniellekain: thanks for adding tests here!

daniellekain (approved):
thanks for making this update!

src/django/api/serializers/embed_config.py
protsack-stephan: Great, thanks!

protsack-stephan (approved):

PR #1019, opened 2026-05-21
[OSDEV-1521]: Show "already active" message for already-verified email links

VadimKovalenkoSNF:
## What

When a user clicks an email confirmation link that has already been used, the platform previously showed a confusing generic "Not found" error. This PR replaces that with a clear "Your account is already active. Please log in with your username and password." message with a link to the login page.

## Why

`allauth`'s `EmailConfirmationHMAC.from_key()` queries `EmailAddress` with `verified=False`. If the address is already verified, it returns `None`, which causes `Http404`. The frontend's `logErrorAndDispatchFailure` hardcoded `['Not found']` for any 404 response, giving users no useful guidance.

## How

- **New `VerifyEmailView`** (`api/views/auth/verify_email_view.py`): subclasses `dj_rest_auth`'s `VerifyEmailView`. On `Http404`, it decodes the HMAC key via `django.core.signing` and checks whether the `EmailAddress` already exists with `verified=True`. If so, it returns HTTP 400 with `detail="already_confirmed"`. Truly invalid or expired keys still result in a 404.
- **URL registration** (`oar/urls.py`): the custom view is registered before the `dj_rest_auth.registration.urls` include so it takes precedence.
- **Frontend** (`ConfirmRegistration.jsx`): detects the `already_confirmed` code and renders a clear informational message and login link instead of the generic red error list.
- **Tests** (`api/tests/test_verify_email.py`): three cases — already-verified returns 400 + `already_confirmed`, unverified email is confirmed successfully (200), and a garbage key returns 404.

## Ticket

[OSDEV-1551](https://opensupplyhub.atlassian.net/browse/OSDEV-1551)

[OSDEV-1551]: https://opensupplyhub.atlassian.net/browse/OSDEV-1551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/react/src/components/ConfirmRegistration.jsx
```diff
@@ -20,6 +20,10 @@ import {
     facilitiesRoute,
 } from '../util/constants';
 
+const ALREADY_CONFIRMED = 'already_confirmed';
+const ALREADY_CONFIRMED_MESSAGE =
+    'Your account is already active. Please log in with your username and password.';
+
```
valencia-max, line 26: I think it's better whether we move these string constants to `../utils/constants`

src/django/api/views/auth/verify_email_view.py
```diff
@@ -0,0 +1,40 @@
...
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        key = serializer.validated_data['key']
+        self.kwargs['key'] = key
+        try:
+            confirmation = self.get_object()
+        except Http404:
+            try:
+                max_age = (
+                    60 * 60 * 24
```
valencia-max, line 25: Should we move this to an integer constant value? What does these numbers mean?

valencia-max (approved):
I left some comments, overall, LGTM!!

VadimKovalenkoSNF:
Closing: branch renamed to `OSDEV-1521-always-show-active-account` (correct ticket OSDEV-1521). Continued in the replacement PR.

PR #1024, opened 2026-05-26
[OSDEV-2542] Use Exists-based partner field filtering to improve facility query performance

protsack-stephan:
This PR was introduced to improve performance of `GET /api/facilities` for Spotlight Data Partners filtering.

### Why
The previous implementation had additional query-path complexity (including shared-facilities/combine behavior) that added overhead to facilities filtering. This change simplifies partner-contributor filtering logic while keeping Spotlight Data Partners support intact.

### What changed
- Removed the shared-facilities mode (`combine_partner_contributors`) from:
  - frontend filter UI/state/query-string handling
  - backend facilities query params and filtering
  - API docs for `GET /api/facilities`
- Kept Spotlight partner contributor filtering via `partner_contributor` with OR semantics.
- Updated partner contributor filter internals to:
  - use clearer variable naming
  - combine system-field country filters into one `country_code__in` path when both are requested
- Updated FE/BE tests to remove shared-facilities coverage and validate the simplified behavior.
- Updated frontend rendering so the **Spotlight Data Partners** filter is hidden when the `private_instance` feature flag is active (for example, private instances such as RBA).
- Updated release notes wording to match the API/UI behavior, including private-instance visibility.


src/django/api/models/facility/partner_contributor_filter.py
```diff
@@ -91,10 +97,16 @@ def apply_partner_contributors_filter(
...
             if not contributor_field_names:
                 qs = qs.none()
                 break
-            qs = apply_partner_fields_or_filter(qs, contributor_field_names)
+
+            qs = apply_partner_fields_or_filter(
+                qs,
+                contributor_field_names,
+                partner_contributors,
+            )
```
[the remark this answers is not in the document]
vlad-shapik: @coderabbit "AND" logic was removed

src/django/api/models/facility/partner_contributor_filter.py
```diff
@@ -78,34 +96,25 @@ def apply_partner_contributors_filter(
...
+        facilities_queryset = apply_partner_fields_or_filter(
+            facilities_queryset,
+            field_names,
+            partner_contributors,
         )
-        if field_names:
-            qs = apply_partner_fields_or_filter(qs, field_names)
 
-    return qs
+    return facilities_queryset
```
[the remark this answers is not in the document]
vlad-shapik: Good catch — fixed. Added an explicit guard to return `facilities_queryset.none()` when selected `partner_contributors` resolve to no active/grouped partner fields. Also added regression coverage in `test_partner_contributor_no_active_grouped_fields_returns_none`.

src/django/api/models/facility/partner_contributor_filter.py
```diff
@@ -78,34 +96,25 @@ def apply_partner_contributors_filter(
...
+                field_name,
+            ) in partner_fields
+        }
+    )
+    if field_names:
+        facilities_queryset = apply_partner_fields_or_filter(
+            facilities_queryset,
+            field_names,
+            partner_contributors,
         )
```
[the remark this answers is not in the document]
vlad-shapik: Partially fixed

src/django/api/models/facility/partner_contributor_filter.py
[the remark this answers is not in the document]
vlad-shapik: Not applicable for this PR scope anymore — `combine_partner_contributors` / shared-facilities behavior was removed from this ticket and is tracked separately in OSDEV-2756.

src/django/api/tests/test_facility_search.py
```diff
@@ -244,85 +247,95 @@ def test_partner_contributor_or_search(self):
...
+        contributor_one.partner_fields.add(inactive_field_one)
+        contributor_two.partner_fields.add(inactive_field_two)
 
         response = self.client.get(
-            "{}?partner_contributor={}&partner_contributor={}"
-            "&combine_partner_contributors=AND".format(
+            "{}?partner_contributor={}&partner_contributor={}".format(
                 self.base_url, contributor_one.id, contributor_two.id
             )
         )
```
VadimKovalenkoSNF, line 341: **Optional**
Missing isolation test. The three new tests cover OR across contributors, inactive-field filtering, and the all-inactive/empty case — good coverage for the happy paths. What's not covered: a facility with an ExtendedField for (contributor=A, field_name=field_b) (B's field name but submitted by A) should not appear when filtering for contributor B alone. Given the cross-join in the Exists query, this case would silently return a wrong result. A test for it would either confirm the behavior is acceptable or surface the bug.

src/django/api/models/facility/partner_contributor_filter.py
```diff
@@ -78,34 +95,34 @@ def apply_partner_contributors_filter(
...
-                    field_name,
-                ) in partner_fields
-            }
-        )
-        if field_names:
-            qs = apply_partner_fields_or_filter(qs, field_names)
+    if not partner_fields:
+        return facilities_queryset.none()
+
+    field_names = list(
```
VadimKovalenkoSNF, line 106: **Optional**
Redundant extraction of contributor_ids. The function fetches partner_fields as (contributor_id, field_name) tuples and then immediately re-splits them into two independent sets. contributor_ids is functionally equivalent to the subset of partner_contributors that have active grouped fields — but since partner_contributors without any active fields are already excluded from field_names, you can just pass partner_contributors directly.

This whole block could be flattened:

```
field_names = list(
    PartnerField.objects.filter(
        contributor__id__in=partner_contributors,
        active=True,
        group__isnull=False,
    ).values_list("name", flat=True).distinct()
)

if not field_names:
    return facilities_queryset.none()

return apply_partner_fields_or_filter(
    facilities_queryset,
    field_names,
    list(partner_contributors),
)
```

Same behavior, one fewer result to unpack, no intermediate tuple list or set comprehensions.

src/django/api/models/facility/partner_contributor_filter.py
```diff
@@ -1,75 +1,92 @@
...
-                    "FROM api_facilityindex fi "
-                    "WHERE EXISTS ("
-                    "  SELECT 1 FROM unnest(fi.extended_fields) AS ef "
-                    "  WHERE ef->>'field_name' = ANY(%s)"
-                    ")",
-                    [regular],
+            Exists(
+                ExtendedField.objects.filter(
+                    facility_id=OuterRef("id"),
+                    contributor_id__in=contributor_ids,
```
VadimKovalenkoSNF, line 38: **Optional**
Implicit cross-join between contributors and field names. The filter contributor_id__in=[A, B] AND field_name__in=[field_a, field_b] matches any combination — so an ExtendedField row for (contributor=A, field_name=field_b) would be included, even though contributor A owns field_a and contributor B owns field_b. In practice this doesn't matter because each contributor only creates ExtendedField rows with their own field names, but it's an implicit data-integrity assumption with no test or constraint backing it.

src/django/api/models/facility/partner_contributor_filter.py
```diff
@@ -1,75 +1,92 @@
...
-        # We still need SQL for unnest() over ArrayField(JSONField), but we
-        # keep it contained to a subquery and combine everything with ORM Q.
+    regular_field_names = [
+        field_name
+        for field_name in field_names
+        if field_name not in SYSTEM_PARTNER_FIELD_NAMES
+    ]
+
+    if regular_field_names:
+        # We need to use a subquery to match the extended fields because
```
VadimKovalenkoSNF: Stale comment. The ArrayField(JSONField) justification belonged to the old RawSQL/unnest() approach. The Exists subquery is used here for performance and correctness, not because of array-containment limitations. Consider replacing with a short explanation of why Exists is preferred (e.g. avoids a correlated raw-SQL unnest, plays well with the ORM query planner).

VadimKovalenkoSNF (dismissed):
Overall looks good. I left a few comments provided by AI assistant that are optional.

src/django/api/models/facility/partner_contributor_filter.py
protsack-stephan: This limits the query surface. Having them both included ensures that the query can narrow things down more quickly, so I would not recommend changing this. The query planner will determine the shortest path to the result, and this gives it more options.

vlad-shapik: > This limits the query surface. Having them both included ensures that the query can narrow things down more quickly, so I would not recommend changing this. The query planner will determine the shortest path to the result, and this gives it more options.

+1

src/django/api/models/facility/partner_contributor_filter.py
vlad-shapik: That simplification is nice, but keeping contributor_ids is a defensible safety guard because it scopes matching to contributors that actually have active grouped partner fields in the current query path.

src/django/api/tests/test_facility_search.py
vlad-shapik: I have added the test for that. Thanks!

src/django/api/models/facility/partner_contributor_filter.py
vlad-shapik: Done, thank you!

VadimKovalenkoSNF (approved):

PR #1028, opened 2026-05-28
[OSDEV-555] Fix list replacement workflow: deactivate at upload, restore on reject, clean Pending filter

VadimKovalenkoSNF:
Fix list replacement workflow bugs where replaced lists appeared in the wrong Admin Dashboard filters and ineligible lists were shown in the replacement dropdown.

**Problem**

When a contributor uploaded list B to replace list A, three distinct bugs existed in the replacement workflow:

List A appeared in the Admin Dashboard Pending filter. `REPLACED` is a computed status derived from the `replaced_by` reverse relation — the DB status column still holds `PENDING`. The Pending filter queried by DB status only, so replaced lists bled into it. This also affected legacy lists where the replacement FK was later cleared but is_active was never restored, leaving a "zombie" PENDING + inactive record with no replaced_by.
List A appeared in the Admin Dashboard Approved filter simultaneously with the Replaced filter. A replaced list retains its raw `APPROVED` DB status even after a replacement is uploaded, causing it to match both filters at once.

The "Select a list to replace" dropdown showed ineligible lists. Replaced, rejected, and inactive lists were visible in the dropdown, which was confusing and could lead to server-side 400 errors if selected.

**Changes**

`admin_facility_list_view.py`
When filtering by PENDING or APPROVED, additionally exclude lists that have a replaced_by (active replacement link) or source.is_active = False (broken-link zombie state). A replaced list now appears only in the Replaced filter.

`facility_list_view_set.py — create()`
Added a guard that rejects replacing a REJECTED or inactive list with HTTP 400.

`facility_list_view_set.py — reject()`
When a replacement is rejected, clear the replaces FK so the original list reverts from REPLACED back to its prior status and becomes eligible for future replacement uploads. The original list's source remains active throughout — it is only deactivated when the replacement is approved.

`ContributeForm.jsx / ContributeFormSelectListToReplace.jsx`
Filter facilityLists before passing to the dropdown, keeping only PENDING + is_active or APPROVED + is_active lists.
Removed the (inactive) label from dropdown items — ineligible lists no longer reach the component.

**Notes**

The six legacy lists from the ticket description were already fixed by migration 0208 shipped in PR #988 (OSDEV-463). No new migration is required.
REPLACED is never stored in the DB; it is computed by the serializer from the replaced_by reverse relation. Filtering by replaced_by__isnull is the correct way to exclude these lists from the Pending and Approved filters.
The replaced list's source stays is_active = True during the review window and is only set to is_active = False when the replacement is approved (existing approve() logic, unchanged).
Related

Jira: [OSDEV-555](https://opensupplyhub.atlassian.net/browse/OSDEV-555)

[OSDEV-555]: https://opensupplyhub.atlassian.net/browse/OSDEV-555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


doc/release/RELEASE-NOTES.md
```diff
@@ -3,6 +3,23 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html). The format is based on the `RELEASE-NOTES-TEMPLATE.md` file.
 
+## Release 2.25.0
+
+## Introduction
+* Product name: Open Supply Hub
+* Release date: *Provide release date*
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Fixed.

src/django/api/views/admin_facility_list_view.py
```diff
@@ -63,5 +63,10 @@ def get_queryset(self):
             facility_lists = facility_lists.filter(replaced_by__isnull=False)
         elif status is not None:
             facility_lists = facility_lists.filter(status=status)
+            if status in (FacilityList.PENDING, FacilityList.APPROVED):
+                facility_lists = facility_lists.filter(
+                    replaced_by__isnull=True,
+                    source__is_active=True,
+                )
```
protsack-stephan: This function is quite complicated. I understand adding an additional `if` statement here, but maybe we can move this to the same level as the previous statements as an `elif` instead of putting it here?

src/react/src/components/ContributeForm.jsx
```diff
@@ -134,14 +135,20 @@ const ContributeForm = ({
...
-                replaces={replaces}
-                handleChange={updateListToReplace}
-            />
-        ) : null;
+    const eligibleLists = (facilityLists || []).filter(
+        list =>
+            list.is_active &&
+            (list.status === facilityListStatusChoicesEnum.PENDING ||
+                list.status === facilityListStatusChoicesEnum.APPROVED),
+    );
```
protsack-stephan, line 145: I think this can be simplified a little bit, something like this would be much easier to read:

```javascript
const { PENDING, APPROVED } = facilityListStatusChoicesEnum;

const eligibleLists = (facilityLists || []).filter(({ is_active, status }) => 
    is_active && [PENDING, APPROVED].includes(status)
);
```

protsack-stephan (commented):
Great work, couple small comments.

src/react/src/components/ContributeFormSelectListToReplace.jsx
```diff
@@ -22,21 +22,7 @@ export default function ContributeFormSelectListToReplace({
...
-        if (active) {
-            return name || filename;
-        }
-
-        if (name) {
-            return `${name} (inactive)`;
-        }
-
-        return name ? `${name} (inactive)` : `${filename} (inactive)`;
-    };
```
valencia-max, line 39: As we are filtering to only see active list, I think that why you are removing this statements, right?

valencia-max (dismissed):
Would be nice to see how these changes are working in UI, LGTM!

src/react/src/components/ContributeFormSelectListToReplace.jsx
VadimKovalenkoSNF: Yes, exactly. The (inactive) label was removed because ineligible lists (inactive, replaced, rejected) are now filtered out in `ContributeForm.jsx` before they ever reach `ContributeFormSelectListToReplace`. By the time a list arrives at the dropdown component, it's guaranteed to be active and eligible — so the component never needs to know about `is_active` at all.

src/django/api/views/admin_facility_list_view.py
VadimKovalenkoSNF: Good point. I refactored this part.

src/react/src/components/ContributeForm.jsx
VadimKovalenkoSNF: Done.

protsack-stephan (approved):

src/django/api/views/admin_facility_list_view.py
```diff
@@ -61,6 +58,11 @@ def get_queryset(self):
             facility_lists = facility_lists.filter(source__in=sources)
         elif status == FacilityList.REPLACED:
             facility_lists = facility_lists.filter(replaced_by__isnull=False)
+        elif status in (FacilityList.PENDING, FacilityList.APPROVED):
+            facility_lists = facility_lists.filter(
+                replaced_by__isnull=True,
+                source__is_active=True,
+            )
         elif status is not None:
             facility_lists = facility_lists.filter(status=status)
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Fixed

vlad-shapik (approved):

PR #1036, opened 2026-06-02
[OSDEV-2657] Added admin Google Sheet uploads for partner-field PATCH moderation events

vlad-shapik:
## Summary

[OSDEV-2657](https://opensupplyhub.atlassian.net/browse/OSDEV-2657) Added a moderator-facing Django admin workflow and AWS Batch pipeline so partner-field production location updates can be submitted from Google Sheets without using the API.

- Add `PartnerDataFileUpload` model, migration, and admin UI. Saving a new upload submits a dedicated AWS Batch job and moves the record to `PROCESSING`.
- Add `process_partner_data_file_uploads` and the `api.partner_data_file_upload` package (Google Sheets client, header/column parsing, row processor, Batch submission). The command reads sheet rows, validates partner fields against contributor permissions and JSON schemas, creates pending `UPDATE` moderation events through the existing `LocationContribution` pipeline, and writes per-row `error` / `moderation_id` feedback back to the sheet.
- Provision dedicated Batch compute environment, queue, and job definition for partner uploads. Wire `BATCH_PARTNER_DATA_FILE_UPLOAD_JOB_QUEUE_NAME` and `BATCH_PARTNER_DATA_FILE_UPLOAD_JOB_DEF_NAME` into the ECS app so admin submission targets the partner job definition (`queueentryuuid`), not the default facility-list job definition.
- Remove the unused direct data load Batch job definition and trim the partner Batch worker env vars to essentials.
- Document local usage in `README.md` and add OSDEV-2657 entries to the 2.25.0 release notes.

## Test plan

- [x] Run migrations: `docker compose exec django python manage.py migrate`
- [ ] Set `GOOGLE_SERVICE_ACCOUNT_CREDS_BASE64` in `.env` for local sheet access
- [ ] In Django admin, create a **Partner Data File Upload** with a valid Google Sheet link and contributor (sheet must include `os_id` and authorized partner field columns)
- [ ] On a deployed environment, confirm an AWS Batch job is submitted and completes successfully
- [ ] Locally (or if Batch is unavailable), set the upload to `PROCESSING` and run: `./scripts/manage process_partner_data_file_uploads`
- [ ] Verify the sheet gets `error` / `moderation_id` columns, successful rows create pending moderation events, and failed rows show readable errors

[OSDEV-2657]: https://opensupplyhub.atlassian.net/browse/OSDEV-2657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/partner_data_file_upload/errors.py
```diff
@@ -0,0 +1,28 @@
+DEVELOPER_PROCESSING_ERROR_PREFIX = "Contact developers: "
+DEVELOPER_ERROR_HINTS = (
+    "Google Service Account",
+    "AWS credentials",
+    "BATCH_PARTNER_DATA_FILE_UPLOAD_JOB_QUEUE_NAME",
+    "BATCH_PARTNER_DATA_FILE_UPLOAD_JOB_DEF_NAME",
+)
```
valencia-max: We have `constants.py`, you can include these values there

valencia-max (dismissed):
LGMT! Would be so great to see this working, if you have a chance, can you please upload a short recording of this, please?

src/django/api/partner_data_file_upload/processing/processor.py
```diff
@@ -0,0 +1,200 @@
...
+                row_stats["rows_skipped_empty"] += 1
+                continue
+
+            try:
+                moderation_id = self.event_creator.create(
+                    queue_row.contributor,
+                    record,
+                    context.column_mappings,
+                )
+                self.sheets_client.mark_row(
```
daniellekain, line 158: is there a risk of being rate-limited by the sheets api here? should we add some retry/backoff logic?

daniellekain: (this applies everywhere that we call mark_row)

daniellekain (changes requested):
thanks for taking on this work! I had a quick question about the sheets api logic

src/django/api/partner_data_file_upload/parsing/parser.py
```diff
@@ -0,0 +1,470 @@
...
+
+class PartnerFieldSheetParser:
+    @staticmethod
+    def _raise_validation_errors(errors: List[str]) -> None:
+        if errors:
+            raise ValueError("; ".join(errors))
+
+    @staticmethod
+    def header_label(header) -> str:
+        return header_label(header)
```
protsack-stephan: Is there a specific reason we are wrapping the `header_label` function within a static method? This seems a bit redundant...

src/django/api/partner_data_file_upload/parsing/parser.py
```diff
@@ -0,0 +1,470 @@
...
+        for field_name, nested_value in object_values.items():
+            raw_data[field_name] = nested_value
+
+        if parse_errors:
+            cls._raise_validation_errors(parse_errors)
+
+        if not raw_data:
+            raise ValueError("No partner field values provided.")
+
+        return raw_data
```
protsack-stephan, line 419: Is there a way to restructure this? Do we really need to have this as a class full of static methods? Instead of having a set of independent static methods, is there a way to unify this logic? For example, something like:

```python
parser = PartnerFieldSheetParser(...)
parser.parse()

print(parser.headers) # holds the header values
print(parser.rows) # holds the row values
# etc...
```

src/django/api/partner_data_file_upload/processing/event_creator.py
```diff
@@ -0,0 +1,81 @@
...
+                        messages.append(str(item_detail))
+                    elif field:
+                        messages.append(str(field))
+                elif item:
+                    messages.append(str(item))
+
+        if messages:
+            return "; ".join(messages)
+
+        return json.dumps(errors)
```
protsack-stephan: Can we make this logic a bit more straightforward? Here's an example of what we could do:

```python
  @staticmethod
  def format_api_errors(errors: Mapping[str, object]) -> str:
      if not errors:
          return "Unknown error."

      messages = []
      detail = errors.get("detail")
      
      if isinstance(detail, str) and detail:
          messages.append(detail)

      error_items = errors.get("errors")
      if not isinstance(error_items, list):
          error_items = []

      for item in error_items:
          if not item:
              continue

          if not isinstance(item, dict):
              messages.append(str(item))
              continue

          field = item.get("field", "")
          item_detail = item.get("detail", "")

          if not field and not item_detail:
              continue

          if field and item_detail:
              messages.append(f"{field}: {item_detail}")
          else:
              messages.append(str(field or item_detail))

      return "; ".join(messages) if messages else json.dumps(errors)
```

src/django/api/admin.py
```diff
@@ -435,6 +440,76 @@ class USCountyTigerlineAdmin(admin.ModelAdmin):
...
+            messages.error(
+                request,
+                (
+                    "Partner data file was saved but "
+                    "Batch submission failed. "
+                    f"Error: {error_message}"
+                ),
+            )
+
+
```
protsack-stephan, line 520: Can we move this outside of this single large file? It would be nice to have a separate `PartnerDataFileUploadAdmin` model that is just included and registered here.

src/django/api/partner_data_file_upload/parsing/parser.py
```diff
@@ -0,0 +1,470 @@
...
+
+        if schema_type == "boolean":
+            if isinstance(value, bool):
+                return value
+            normalized = str(value).strip().lower()
+            if normalized in ("true", "1", "yes"):
+                return True
+            if normalized in ("false", "0", "no"):
+                return False
+            raise ValueError(f"Expected a boolean, not '{value}'.")
```
protsack-stephan: This can be extracted in a lookup dictionary, something like:

```python
def validate_integer(value):
    if isinstance(value, bool):
        raise ValueError("Expected an integer, not a boolean.")
    if isinstance(value, int):
        return value
    if isinstance(value, float) and value.is_integer():
        return int(value)
    try:
        return int(str(value).strip())
    except (TypeError, ValueError) as error:
        raise ValueError(f"Expected an integer, not '{value}'.") from error

def validate_number(value):
    if isinstance(value, bool):
        raise ValueError("Expected a number, not a boolean.")
    if isinstance(value, (int, float)):
        return float(value)
    try:
        return float(str(value).strip())
    except (TypeError, ValueError) as error:
        raise ValueError(f"Expected a number, not '{value}'.") from error

def validate_bool(value):
    if isinstance(value, bool):
        return value
    normalized = str(value).strip().lower()
    if normalized in ("true", "1", "yes"):
        return True
    if normalized in ("false", "0", "no"):
        return False
    raise ValueError(f"Expected a boolean, not '{value}'.")

type_validators = {
    "integer": validate_integer,
    "number": validate_number,
    "boolean": validate_bool,
}

# Then your main logic just becomes:
if schema_type in type_validators:
    return type_validators[schema_type](value)
```

src/django/api/partner_data_file_upload/parsing/parser.py
```diff
@@ -0,0 +1,470 @@
...
+            try:
+                return float(str(value).strip())
+            except (TypeError, ValueError) as error:
+                raise ValueError(
+                    f"Expected a number, not '{value}'."
+                ) from error
+
+        raise ValueError(
+            f"Unsupported partner field type '{partner_field.type}'."
+        )
```
protsack-stephan: This method contains a long chain of if statements that handle the parsing logic directly. To make the code cleaner and easier to maintain, we can extract the specific parsing logic into separate functions and use a dictionary as a lookup table (dispatch table) to map the field types to their respective parsers, something like:

```python
def _parse_int(value):
    if isinstance(value, bool):
        raise ValueError("Expected an integer, not a boolean.")
    if isinstance(value, int):
        return value
    if isinstance(value, float) and value.is_integer():
        return int(value)
    try:
        return int(str(value).strip())
    except (TypeError, ValueError) as error:
        raise ValueError(f"Expected an integer, not '{value}'.") from error

def _parse_float(value):
    if isinstance(value, bool):
        raise ValueError("Expected a number, not a boolean.")
    if isinstance(value, (int, float)):
        return float(value)
    try:
        return float(str(value).strip())
    except (TypeError, ValueError) as error:
        raise ValueError(f"Expected a number, not '{value}'.") from error


_CELL_PARSERS = {
    PartnerField.STRING: lambda v: str(v).strip(),
    PartnerField.INT: _parse_int,
    PartnerField.FLOAT: _parse_float,
}

@classmethod
def parse_cell_value(cls, value: object, partner_field: PartnerField):
    parser = cls._CELL_PARSERS.get(partner_field.type)
    
    if not parser:
        raise ValueError(f"Unsupported partner field type '{partner_field.type}'.")
        
    return parser(value)
```

protsack-stephan (commented):
Good work! Couple of comments from me.

README.md
```diff
@@ -201,6 +202,16 @@ Continue by accepting the list in the web browser dashboard. Then, in the django
 ./scripts/manage batch_process -a match -l 16
 ```
 
+### Process partner data file uploads
+
+Moderators submit partner Google Sheets via Django admin (**Partner Data File Uploads**); deployed environments run this through AWS Batch. Locally, create an upload pointing at a shared Google Sheet (status `PROCESSING`), then run:
+
+```
+./scripts/manage process_partner_data_file_uploads
+```
```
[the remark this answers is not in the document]
vlad-shapik: Handled — added `bash` to the fenced code block in README.md. Will be included in the next commit.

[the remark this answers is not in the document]
vlad-shapik: @coderabbitai

src/django/api/admin.py
```diff
@@ -435,6 +440,76 @@ class USCountyTigerlineAdmin(admin.ModelAdmin):
...
+                update_fields=["status", "processing_error", "updated_at"]
+            )
+            messages.error(
+                request,
+                (
+                    "Partner data file was saved but "
+                    "Batch submission failed. "
+                    f"Error: {error_message}"
+                ),
+            )
```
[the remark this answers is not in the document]
vlad-shapik: Handled — new uploads are now saved as `PROCESSING` before enqueue, and Batch submission runs in `transaction.on_commit` so the worker only starts after the row is committed. Submission failures still mark the record `FAILED` with `processing_error`. Will be included in the next commit.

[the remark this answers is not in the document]
vlad-shapik: @coderabbitai Handled. New uploads are persisted as `PROCESSING` before enqueue, and AWS Batch submission is deferred with `transaction.on_commit` so the worker cannot run while the row is still `PENDING`. Submission failures still set `FAILED` and `processing_error`. Covered by admin tests; will be included in the next commit.

src/django/api/partner_data_file_upload/sheets/client.py
```diff
@@ -0,0 +1,199 @@
...
+        spreadsheet_id = self.parse_spreadsheet_id(link)
+        tab_gid = self.parse_tab_gid(link)
+        metadata = self.service.spreadsheets().get(
+            spreadsheetId=spreadsheet_id
+        ).execute()
+        sheet_name, tab_id = self._resolve_sheet(metadata, tab_gid)
+        values = self.service.spreadsheets().values().get(
+            spreadsheetId=spreadsheet_id,
+            range=f"{sheet_name}!A:Z",
+        ).execute().get("values", [])
```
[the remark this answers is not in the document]
vlad-shapik: @coderabbitai Handled. `GoogleSheetClient.from_env` now builds the Sheets service with an `httplib2.Http` timeout (60s) wrapped in `google_auth_httplib2.AuthorizedHttp`, so all `get`/`values`/`batchUpdate` calls inherit the limit. Will be included in the next commit.

src/django/api/partner_data_file_upload/sheets/client.py
```diff
@@ -0,0 +1,199 @@
...
+    def load_workbook(self, link: str) -> SheetWorkbook:
+        spreadsheet_id = self.parse_spreadsheet_id(link)
+        tab_gid = self.parse_tab_gid(link)
+        metadata = self.service.spreadsheets().get(
+            spreadsheetId=spreadsheet_id
+        ).execute()
+        sheet_name, tab_id = self._resolve_sheet(metadata, tab_gid)
+        values = self.service.spreadsheets().values().get(
+            spreadsheetId=spreadsheet_id,
+            range=f"{sheet_name}!A:Z",
```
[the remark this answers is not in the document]
vlad-shapik: @coderabbitai Handled. `_column_letter` now supports multi-letter columns (AA, AB, …), and `load_workbook` builds the read range from the sheet's `gridProperties.columnCount` instead of hardcoding `A:Z`. Tests added for both. Will be included in the next commit.

src/django/oar/settings.py
```diff
@@ -73,25 +73,47 @@
...
+BATCH_PARTNER_DATA_FILE_UPLOAD_JOB_DEF_NAME = os.getenv(
+    'BATCH_PARTNER_DATA_FILE_UPLOAD_JOB_DEF_NAME')
+if (
+    BATCH_PARTNER_DATA_FILE_UPLOAD_JOB_DEF_NAME is None
+    and not DEBUG
+    and not BATCH_MODE
+):
+    raise ImproperlyConfigured(
+        'Invalid BATCH_PARTNER_DATA_FILE_UPLOAD_JOB_DEF_NAME provided, '
+        'must be set')
```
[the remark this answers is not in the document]
vlad-shapik: @coderabbitai Handled. Batch startup checks now reject unset, empty, and whitespace-only values via `(value or '').strip()` for all four settings, matching the runtime truthiness checks in `batch.py`. Also fixed the `BATCH_JOB_QUEUE_NAME` typo in the error message. Will be included in the next commit.

src/django/api/partner_data_file_upload/errors.py
vlad-shapik: Handled with a package-local `api/partner_data_file_upload/constants.py` instead of the main `api/constants.py` — this feature is fairly standalone and the shared constants file is already large. Moved constants from `errors.py`, `batch.py`, `sheets/client.py`, and `parsing/parser.py` into that module. Will be included in the next commit.

vlad-shapik:
@valencia-max Thank you — really appreciate the review!

I’ll try to put together a short recording if I can find the time. In the meantime, you should be able to try it locally:

- README: [Process partner data file uploads](https://github.com/opensupplyhub/open-supply-hub/blob/OSDEV-2657-automate-data-update-process-csv-excel-partners/README.md#process-partner-data-file-uploads)
- Confluence guide I put together: [Spotlight: Partner data Google Sheets upload (Django admin)](https://opensupplyhub.atlassian.net/wiki/spaces/SD/pages/1305083906/Spotlight+Partner+data+Google+Sheets+upload+Django+admin)

Happy to answer any questions if anything is unclear while walking through it.

src/django/api/partner_data_file_upload/processing/processor.py
vlad-shapik: Handled — good catch, and yes, we saw this in practice during a 10k-row load test (`429 Quota exceeded: Write requests per minute per user, limit 60` on `mark_row` → `batchUpdate`).

**Proactive throttling (primary fix)**
- Added `SHEETS_ROW_PROCESSING_DELAY_SECONDS = 1.1` in `api/partner_data_file_upload/constants.py` (~55 writes/min, under the 60/min Sheets limit).
- `process_partner_data_file_uploads` passes this into `PartnerDataFileUploadProcessor`.
- After each processed row (success or error — both call `mark_row`), we enforce a **minimum 1.1s interval from row start to row start**: if processing already took ≥1.1s we proceed immediately; otherwise we sleep only the remaining time.

**Resume support (reduces writes on re-run)**
- Rows that already have `moderation_id` or `error` are skipped (no `mark_row`, no moderation event), so a failed/partial upload can be safely re-submitted without duplicating work.

This covers every `mark_row` call in the processor loop. I have not added centralized 429 retry/backoff in `GoogleSheetClient` yet — throttling is the main guardrail.

src/django/api/partner_data_file_upload/parsing/parser.py
vlad-shapik: Handled — removed the redundant `PartnerFieldSheetParser.header_label` static wrapper and call `header_label` from `parsing.types` directly in `build_header_map` (same pattern as `GoogleSheetClient.ensure_tracking_columns`).

src/django/api/partner_data_file_upload/parsing/parser.py
vlad-shapik: Good point — an instance-based parser with headers / rows / parse() would be cleaner. I kept it as stateless classmethods for this PR since sheet state already lives in SheetWorkbook + SheetProcessingContext, and the processor orchestrates the pipeline. Happy to refactor in a follow-up task if we want that API shape.

src/django/api/partner_data_file_upload/processing/event_creator.py
vlad-shapik: Handled — refactored `format_api_errors` to use your suggested early-continue flow. Behavior is unchanged for normal API error payloads; empty dict items are skipped and non-list `errors` is normalized to `[]`.

src/django/api/admin.py
vlad-shapik: Good idea — I’d like to do this as part of converting api/admin.py into an api/admin/ package (like api/views/), rather than extracting just this one admin in this PR.
I would handle that in a follow-up task so we don’t mix a structural admin refactor with the upload feature.

src/django/api/partner_data_file_upload/parsing/parser.py
vlad-shapik: Done — extracted `_validate_integer`, `_validate_number`, and `_validate_boolean` module-level helpers and wired `parse_leaf_cell_value` through a `_LEAF_TYPE_VALIDATORS` dispatch dict. Reused the same integer/number validators for `parse_cell_value` via `_CELL_PARSERS` so scalar and nested parsing stay in sync.

src/django/api/partner_data_file_upload/parsing/parser.py
vlad-shapik: Done — `parse_cell_value` now uses `_CELL_PARSERS` with shared `_validate_integer` / `_validate_number` helpers (same refactor as the `parse_leaf_cell_value` thread).

src/django/api/partner_data_file_upload/parsing/parser.py
vlad-shapik: Update: validators and dispatch tables now live in `parsing/cell_value_parsers.py` (`_validate_integer`, `_validate_number`, `_validate_boolean`, `LEAF_TYPE_VALIDATORS`). `PartnerFieldSheetParser.parse_leaf_cell_value` remains on the parser class and uses that module.

src/django/api/partner_data_file_upload/parsing/parser.py
vlad-shapik: Update: validators and `CELL_PARSERS` now live in `parsing/cell_value_parsers.py`. `PartnerFieldSheetParser.parse_cell_value` remains on the parser class and delegates to `CELL_PARSERS` from that module (shared integer/number validators with the leaf parser).

protsack-stephan (approved):

daniellekain (approved):

PR #1042, opened 2026-06-04
[OSDEV-2662] Redirect info.openapparel.org to info.opensupplyhub.org via CloudFront

VadimKovalenkoSNF:
Adds a dedicated CloudFront distribution with a CloudFront Function that issues a 301 permanent redirect from info.openapparel.org to https://info.opensupplyhub.org, resolving a subdomain takeover risk (SOC 2 finding) caused by the subdomain pointing at a decommissioned Servd host. Includes ACM certificate and Route 53 A/AAAA alias records in the openapparel.org hosted zone.

Fix [OSDEV-2662](https://opensupplyhub.atlassian.net/browse/OSDEV-2662)

[OSDEV-2662]: https://opensupplyhub.atlassian.net/browse/OSDEV-2662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployment/terraform/cdn.tf
```diff
@@ -804,3 +804,138 @@ resource "aws_cloudfront_distribution" "cdn" {
...
+
+  providers = {
+    aws.acm_account     = aws.certificates
+    aws.route53_account = aws
+  }
+
+  domain_name           = "info.openapparel.org"
+  hosted_zone_id        = data.aws_route53_zone.openapparel[0].zone_id
+  validation_record_ttl = "60"
+}
```
protsack-stephan: Is there a way to avoid referencing a module on the `azavea` side? From what I see, this is something that we are referencing across the IaC codebase. We should take a look and see if we can avoid doing this at least in this PR, and create a ticket to remove those in the long run altogether. This is a bit of a risk and a surface of attack for our infrastructure.

protsack-stephan (commented):
Good work! One question, though: can we test this outside of production?

deployment/terraform/cdn.tf
VadimKovalenkoSNF: Good callout, I've replaced `azavea` module with three standard Terraform resources:

`aws_acm_certificate` — requests the cert in **us-east-1** via **aws.certificates** provider
`aws_route53_record` "info_openapparel_cert_validation" — creates the DNS validation CNAME records in the `openapparel.org` zone
`aws_acm_certificate_validation` — waits for validation to complete before the cert ARN is used by CloudFront

VadimKovalenkoSNF:
> can we test this outside of production?

@protsack-stephan , to answering your question:

Testing this in `Pre-prod` or `Staging` isn't possible — all three environments share the same AWS account, so only one CloudFront distribution can hold the `info.openapparel.org` alias at a time, and the `openapparel.org` Route 53 zone is shared too. 
A second environment applying the same resources would immediately conflict. The change is scoped to Production only intentionally. The `Pre-prod` Terraform run will still validate the configuration is correct — it just won't create any resources.

deployment/terraform/cdn.tf
protsack-stephan: Awesome, thanks!

protsack-stephan (dismissed):

vlad-shapik (dismissed):
LGTM

vlad-shapik (dismissed):

protsack-stephan (approved):

vlad-shapik (approved):

PR #1052, opened 2026-06-08
[OSDEV-2696] Add os_id_snapshot to ModerationEvent to survive facility deletion

daniellekain:
## What

Add `os_id_snapshot` — a `CharField` on `ModerationEvent`, written at approval time — so the approved OS ID survives later facility deletion or merging, and surface it on the contribution-record page.

## Why

The existing `os_id` FK uses `on_delete=SET_NULL`, which Django implements as a bulk `UPDATE` that bypasses ORM signals and OpenSearch indexing. When a facility is deleted or merged into another, `os_id` is silently set to `NULL` with no audit trail, making it impossible to trace which facility a moderation event approved.

`os_id_snapshot` survives facility deletion or merging because it is a plain string, not a FK.

## Changes

- **Field:** `os_id_snapshot` on `ModerationEvent` (migration `0213_add_os_id_snapshot_to_moderation_event`, blank, default `''`).
- **Write at approval:** `EventApprovalTemplate.__update_event` sets `os_id_snapshot = item.facility_id` only when empty — write-once, never overwriting the first-approval value.
- **Consumer wiring:** the moderation-events Logstash pipeline now indexes `os_id` as `COALESCE(NULLIF(os_id_snapshot, ''), os_id)` and exposes `os_id_snapshot` as its own keyword field. Because the contribution-record page renders the raw OpenSearch document, the OS ID now persists on that page after a facility delete/merge.

## Tests

`test_moderation_events_add_production_location.py`:
- `test_approval_sets_os_id_snapshot` — snapshot is populated to match `os_id` on approval.
- `test_os_id_snapshot_is_not_overwritten_on_reapproval` — approval does not overwrite an existing snapshot.
- `test_os_id_snapshot_persists_when_os_id_is_nulled` — simulates the `SET_NULL` cascade and verifies the snapshot survives.

## Backfilling existing rows

Backfilling existing `ModerationEvent` rows (forward-fill of the ~298k where `os_id` is still present, plus best-effort recovery of the ~36k already nulled) is split into **#1073** ([OSDEV-2878](https://opensupplyhub.atlassian.net/browse/OSDEV-2878)), stacked on this branch.

## Release instructions

See `doc/release/RELEASE-NOTES.md` (Release 2.26.0) for the post-deployment steps.

Part of [OSDEV-2696](https://opensupplyhub.atlassian.net/browse/OSDEV-2696)

[OSDEV-2878]: https://opensupplyhub.atlassian.net/browse/OSDEV-2878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/migrations/0214_add_os_id_snapshot_to_moderation_event.py
```diff
@@ -0,0 +1,25 @@
...
+        ('api', '0210_deactivate_rejected_list_sources'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='moderationevent',
+            name='os_id_snapshot',
+            field=models.CharField(
+                blank=True,
+                default='',
```
protsack-stephan, line 16: Quick question here, do I understand correctly that we are defaulting the database value to empty string? Does it make sense to make this field nullable instead?

src/django/api/moderation_event_actions/approval/event_approval_template.py
```diff
@@ -297,6 +297,8 @@ def __update_event(self, item: FacilityListItem) -> None:
         event.action_type = self._get_action_type()
         event.action_perform_by = self.__moderator
         event.os_id = item.facility_id
+        if not event.os_id_snapshot:
+            event.os_id_snapshot = item.facility_id
```
protsack-stephan, line 301: Is there a possibility that `os_id_snapshot` is set in advance and we don't want to update it because it is already set?

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot.py
```diff
@@ -0,0 +1,60 @@
...
+            return
+
+        # Bulk update is safe here: os_id_snapshot has no signal handlers,
+        # so bypassing the ORM save() path has no side effects.
+        updated = qs.update(os_id_snapshot=F('os_id'))
+
+        log.info('Backfill complete. Updated %s rows.', updated)
+        self.stdout.write(self.style.SUCCESS(
+            f'Done. Updated {updated} rows.'
+        ))
```
protsack-stephan: Is there a benefit to having this be a separate command? Does it make any sense to make it part of the migration instead?

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
```diff
@@ -0,0 +1,183 @@
...
+        if total == 0:
+            self.stdout.write(self.style.SUCCESS('Nothing to recover.'))
+            return
+
+        opensearch = OpenSearchServiceConnection()
+
+        self.stdout.write('Querying OpenSearch for os_id values...')
+        os_id_map = self.__get_os_id_map_from_opensearch(
+            target_uuids, opensearch
+        )
```
protsack-stephan: I'm not entirely sure if we can trust OpenSearch here. Although it should be faster than going through the facility list items, it might contain skewed data. @vlad-shapik , do you agree with me? Is there anything I'm missing?

protsack-stephan (commented):
Good work! Couple of minor comments from me.

src/django/api/migrations/0214_add_os_id_snapshot_to_moderation_event.py
```diff
@@ -0,0 +1,25 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0210_deactivate_rejected_list_sources'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='moderationevent',
+            name='os_id_snapshot',
```
VadimKovalenkoSNF, line 13: I have a question here: who is the consumer of this `os_id_snapshot`? I don't see it has been indexed to the OpenSeach, exposed in any serializer, or used by the contribution-record page.

VadimKovalenkoSNF (changes requested):

src/django/api/moderation_event_actions/approval/event_approval_template.py
vlad-shapik: I believe it can be done in advance here as well, since the OS ID may differ from the one set during event creation if another one is selected from the match list.

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot.py
vlad-shapik: +1

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
vlad-shapik: @protsack-stephan — I agree, we should not treat OpenSearch as a reliable source for the original OS ID.

**Why OpenSearch is fragile here**

When a `Facility` is deleted, Django nulls `ModerationEvent.os_id` via a bulk `UPDATE` that bypasses `save()` and the `post_save` signal. OpenSearch is **not** updated in that path, so it can temporarily hold a stale `os_id`. That stale data is what makes OpenSearch look useful for recovery — but it is drift, not an audit trail.

**We already saw this in Production**

We cleared the moderation-events index in Production some time ago (via `clear-opensearch`), and Logstash refilled it from Postgres. Logstash reads the current DB value (`ame.os_id`), so after reindex OpenSearch reflects whatever is in Postgres today — including `NULL` where the FK was cleared. Any previously stale OS IDs in OpenSearch would have been lost at that point.

So OpenSearch is not a single source of truth, and we cannot assume it still holds the original OS IDs for events where `os_id` was nulled in Postgres. In practice, that means some original OS IDs may have been lost inevitably if we were counting on OpenSearch to preserve them.

**Better sources in Postgres**

For approved/rejected events where `os_id` is already `NULL`, I would lean on Postgres instead:

1. **`FacilityListItem` linked via `moderation_event_id`** — on facility delete, `processing_results` records `deleted_os_id`, and the list item may still reflect the facility from the approval flow.
2. **`FacilityAlias`** — may map a deleted OS ID to a promoted/replacement facility after delete/merge.

OpenSearch might recover some rows only in the narrow window before a clear/reindex, but I would not build the backfill around it. I would treat `FacilityListItem` (and `FacilityAlias` where applicable) as the primary recovery source, with OpenSearch at most as an optional best-effort shortcut.

vlad-shapik: We should assume we **won't be able to recover the original OS ID for every moderation event**. Several things work against full recovery:

- When a facility is deleted, `ModerationEvent.os_id` is nulled via a bulk SQL update, so Postgres no longer holds the value for many approved/rejected events.
- OpenSearch is not a reliable archive: we cleared the moderation-events index in Production some time ago (`clear-opensearch`) and Logstash refilled it from Postgres, so OpenSearch now reflects current DB state, not historical OS IDs.
- `FacilityListItem.moderation_event_id` was added later (2.21.0) with no backfill, so many older approved events won't have that link even if a list item exists.
- `FacilityAlias` only helps once we already know the deleted OS ID — it can't reconnect an orphaned moderation event on its own.

Because of that, I'd like to **compare two recovery approaches** and measure how many events each one can actually fix:

1. **Postgres-based recovery** — e.g. events where `os_id` is still present, `FacilityListItem` / `processing_results.deleted_os_id`, `cleaned_data`, and other DB-side signals.
2. **OpenSearch-based recovery** — checking whether any moderation-event documents still contain an `os_id` that Postgres no longer has.

With access to both **Production Postgres** and **OpenSearch**, we can run both checks and compare counts before writing anything back. That should tell us:
- how many moderation events can be updated with a confident OS ID,
- how much overlap there is between the two sources,
- and how many rows are likely **unrecoverable** no matter which path we use.

I don't think we should treat either source as the single source of truth. The useful outcome is a side-by-side comparison of coverage, then only backfilling/updating rows where we have a strong match.

daniellekain:
Heads-up on scope: the **backfilling of existing data** (forward-fill data migration + the `os_id_snapshot` recovery command) has been split out into **#1073** (sub-task [OSDEV-2878](https://opensupplyhub.atlassian.net/browse/OSDEV-2878), stacked on this branch).

This PR now covers just the durable schema change and making it usable going forward:
- `os_id_snapshot` field (migration `0213`)
- write-once assignment at approval (`EventApprovalTemplate.__update_event`)
- moderation-events Logstash consumer: index `os_id` as `COALESCE(NULLIF(os_id_snapshot,''), os_id)` and expose `os_id_snapshot` as its own keyword field, so the contribution-record page shows the durable value after a facility delete/merge.

The OpenSearch-reliability discussion is carried over to #1073, where OpenSearch is treated as best-effort with a Postgres (`FacilityListItem`) fallback.

[OSDEV-2878]: https://opensupplyhub.atlassian.net/browse/OSDEV-2878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/django/api/migrations/0214_add_os_id_snapshot_to_moderation_event.py
daniellekain: Good catch — wired it up: the moderation-events Logstash pipeline now indexes `os_id` as `COALESCE(NULLIF(os_id_snapshot,''), os_id)` and exposes `os_id_snapshot` as its own keyword field. Since the contribution-record page renders the raw OpenSearch doc, the OS ID now survives facility deletion/merge with no frontend change, restoring the OSDEV-2360 symptom fix for the affected events.

src/django/api/moderation_event_actions/approval/event_approval_template.py
daniellekain: `os_id_snapshot` is only ever written at approval (`__update_event`, using `item.facility_id` — the *approved* match), never at creation, so it captures the final approved OS ID, not a stale creation-time value. The `if not` guard just preserves that first-approval value on any re-run, matching OSDEV-2696's Option A ("written once… never modified").

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
daniellekain: Confirmed with numbers — you were right. Measured against the Apr 2026 prod dump (36,007 target events: APPROVED / `os_id IS NULL` / empty snapshot):

- OpenSearch: **0** recoverable (cleared + refilled from Postgres in prod, so it mirrors the NULLs)
- `FacilityListItem` fallback: **16** (~0.04%) — the `moderation_event` link was added in 2.21.0 with no backfill
- `cleaned_data`: **0** (no OS ID stored)
- `FacilityAlias`: 0 usable (needs a known os_id to start from)

So ~16 / 36,007 with reliable sources. The `deleted_os_id` in `processing_results` is a larger pool (~24,968) but not joined to the events, so only a fuzzy `cleaned_data` heuristic could reach it — not trustworthy for an audit field.

Ive moved all the backfilling (forward-fill + this recovery) to #1073 ([OSDEV-2878](https://opensupplyhub.atlassian.net/browse/OSDEV-2878)) and labeled recovery best-effort, with these numbers in the command help + release notes. The deleted_os_id heuristic is noted there as a possible future spike rather than something this PR attempts.

src/django/api/migrations/0214_add_os_id_snapshot_to_moderation_event.py
daniellekain: Thanks for flagging this! Yes, you're right that it defaults to an empty string. I went with a non-nullable `CharField` (`default=''`) since that's the pattern Django generally recommends for string fields — it avoids ending up with both `NULL` and `''` meaning "no value." It also keeps things consistent with how the value is used elsewhere: the approval write, the forward-fill, and the recovery command all treat `os_id_snapshot=''` as "not captured yet."

That said, I'm happy to switch it to nullable if you'd prefer — let me know what you think!

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot.py
daniellekain: Good call — I moved the forward-fill into a data migration (`0214_backfill_moderation_event_os_id_snapshot`) so it runs automatically with `migrate` instead of being a separate manual step. To keep this PR focused on the schema change and the consumer wiring, I also split all the backfilling (that migration plus the recovery command) out into #1073 ([OSDEV-2878](https://opensupplyhub.atlassian.net/browse/OSDEV-2878)), stacked on this branch. Thanks for the suggestion!

doc/release/RELEASE-NOTES.md
```diff
@@ -19,6 +31,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 * Ensure that the following commands are included in the `post_deployment` command:
     * `migrate`
     * `reindex_database`
+* The `moderation-events` OpenSearch index template changed (new `os_id_snapshot` field + `os_id` fallback). Recreate/reindex the `moderation-events` index so existing documents pick up the new mapping and coalesced `os_id`.
+* Backfilling `os_id_snapshot` for existing moderation events is handled separately in [OSDEV-2878](https://opensupplyhub.atlassian.net/browse/OSDEV-2878).
+
+---
```
VadimKovalenkoSNF: I'm not sure this separator adds much value to the release notes. If we decide to use it, we should apply it consistently across the other release sections as well.

VadimKovalenkoSNF (dismissed):
Good work. I approved with minor comment.

VadimKovalenkoSNF (dismissed):
Re-approved.

src/django/api/moderation_event_actions/approval/event_approval_template.py
valencia-max: I agree with Vlad and Danielle.

valencia-max (dismissed):
Amazing work! Thanks for addessing this!

vlad-shapik (approved):
LGTM
Great work!

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot.py
protsack-stephan: Thanks!

src/django/api/moderation_event_actions/approval/event_approval_template.py
protsack-stephan: Thanks for the explanatinos!

src/django/api/migrations/0214_add_os_id_snapshot_to_moderation_event.py
protsack-stephan: Sounds good to me, thanks for taking a look.

protsack-stephan (dismissed):

protsack-stephan:
Good work! Thanks for addressing my comments!

src/django/api/migrations/0214_add_os_id_snapshot_to_moderation_event.py
vlad-shapik, line 1: The main already has 0213 migration. It should be 0214

vlad-shapik (changes requested):

vlad-shapik (approved):
LGTM

VadimKovalenkoSNF (approved):
Re-approved.

vlad-shapik:
@daniellekain
Please make sure to run a deployment for the Development environment from `main` with Reset OpenSearch checked, and please mention somewhere in the ticket description or QA notes for Anna that the same should be done for Test.

PR #1063, opened 2026-06-11
[OSDEV-2860] Remove base32-crockford dependency broken on PyPI

VadimKovalenkoSNF:
## Description

Remove the `base32-crockford` package from both services and inline its functionality directly.

PyPI's simple index for `base32-crockford` (https://pypi.org/simple/base32-crockford/) now returns an empty page — no distribution files are listed. Pip uses the simple index to resolve packages, so every Docker build that has `base32-crockford==0.3.0` in its requirements fails with:

```
ERROR: Could not find a version that satisfies the requirement base32-crockford==0.3.0 (from versions: none)
```

The package has not been maintained since 2015. Both affected services (`django` and `api-app`) use only two things from it: the 32-character Crockford symbol table and a simple integer-to-base32 encode function. Both are inlined directly in the respective `os_id.py` files, removing the external dependency entirely.

## Changes

- Removed `base32-crockford==0.3.0` from `src/django/requirements.txt`
- Removed `base32-crockford==0.3.0` from `src/dedupe-hub/api/requirements.txt`
- Replaced `import base32_crockford` with inline `_BASE32_SYMBOLS` constant and `_base32_encode()` function in `src/django/api/os_id.py`
- Same inline replacement in `src/dedupe-hub/api/app/utils/os_id.py`

## Jira ticket

[OSDEV-2860](https://opensupplyhub.atlassian.net/browse/OSDEV-2860)

## Test plan

- [x] Existing `api.tests.test_os_id` tests pass (4/4) with the inlined implementation
- [ ] CI Docker builds succeed for both `django` and `api-app` targets

[OSDEV-2860]: https://opensupplyhub.atlassian.net/browse/OSDEV-2860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


VadimKovalenkoSNF:
`base32-crockford==0.3.0` is now fully resolvable by pip. This means the original CI failure was a transient PyPI outage, not a permanent removal. The package is back and the dependency is valid. Closing the PR.

PR #1070, opened 2026-06-15
[OSDEV-2732] Add the ability to control display of the partner fields through admin panel

protsack-stephan:
## What

Introduces per-field visibility controls for partner fields so admins can independently hide a field from the public API and/or data downloads, without deactivating it entirely.

- Add `available_in_api` and `available_in_downloads` boolean flags to the `PartnerField` model (default `True`), with migration `0213_add_partner_field_availability_flags`.
- Move partner field admin config into a dedicated `partner_field_admin.py`:
  - Expose the new availability flags in the list display, list filters, and edit form.
  - Protect system fields (`name`, `type`, `json_schema`, `system_field` can't be edited; system fields can't be deleted).
- Respect the flags everywhere partner fields are surfaced:
  - **API** (`FacilityIndexDetailsSerializer`, `ProductionLocations` v1): filter out fields where `available_in_api` is false for API users.
  - **Downloads** (`FacilityDownloadSerializer`): only include fields where `available_in_downloads` is true.
- Refactor `FacilityDownloadSerializer` to build partner-field headers/rows (including object-schema dotted paths and the MIT living wage / wage indicator system fields) directly, removing the dependency on `FacilitiesDownloadService.get_active_partner_fields`.
- Remove the "available in profile" checkbox.

## Why

Previously a partner field was all-or-nothing: active and visible in both API and downloads, or inactive everywhere. Partners need finer control to expose a field in one surface while hiding it in another. The availability flags provide that, and centralizing the admin logic makes system fields safer to manage.

## Notes / Ticket

- [OSDEV-2732](https://opensupplyhub.atlassian.net/browse/OSDEV-2732)

[OSDEV-2732]: https://opensupplyhub.atlassian.net/browse/OSDEV-2732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

image


src/django/api/models/partner_field.py
```diff
@@ -99,6 +99,19 @@ class Meta:
...
     )
+    available_in_api = models.BooleanField(
+        default=True,
+        verbose_name="Available in API",
+        help_text=(
+            "Indicates if this partner field is available in the API."
+        ),
+    )
+    available_in_downloads = models.BooleanField(
+        default=True,
```
daniellekain, line 110: nit: should we set a verbose name here too?

src/django/api/serializers/facility/facility_index_details_serializer.py
```diff
@@ -370,25 +369,54 @@ def get_is_claimed(self, facility):
...
     def get_partner_fields(self, facility):
         request = self._get_request()
+        fields = get_cached_all_partner_fields()
+        partner_fields = [
+            field for field in fields
+            if field.active
+        ]
+
+        is_api_user = request is not None and request.auth
+        if is_api_user:
```
daniellekain, line 382: super nit: could combine 378 and 379 ex/

// checking for API user
if request is not None and request.auth:

though I do think there's some value in denoting why we're doing the check (hence the comment). Both are the same number of lines / similarly clear

src/django/api/serializers/facility/facility_index_details_serializer.py
```diff
@@ -370,25 +369,54 @@ def get_is_claimed(self, facility):
     )
     def get_partner_fields(self, facility):
         request = self._get_request()
+        fields = get_cached_all_partner_fields()
+        partner_fields = [
```
daniellekain, line 373: This now filters partner_fields to active (and available_in_api for API users) before serializing, so inactive fields with stored 
data — previously included via the full cached list — are no longer returned; assuming that's intended but just wanted to double check?

src/django/api/serializers/facility/facility_download_serializer.py
```diff
@@ -1,46 +1,116 @@
...
+        self.__wage_indicator_helper = WageIndicatorDownloadHelper()
+
+        partner_fields_override = partner_fields is not None
+        fields = (
+            partner_fields
+            if partner_fields_override
+            else [
+                field
+                for field in get_cached_all_partner_fields()
+                if field.active and field.available_in_downloads
```
daniellekain: with this change, MIT living wage / wage indicator columns are now gated on the system PartnerField being active and available_in_downloads (see get_mit_living_wage_headers/get_wage_indicator_headers, lines 209/225) -- just wanted to make sure this was intended?

src/django/api/views/v1/production_locations.py
```diff
@@ -301,7 +301,9 @@ def __get_partner_fields(self, pk):
         """
         all_partner_fields = get_cached_all_partner_fields()
         partner_field_names = [
-            field.name for field in all_partner_fields if field.active
+            field.name
+            for field in all_partner_fields
+            if field.active and field.available_in_api
```
daniellekain, line 306: Optional: the v1 available_in_api filter has no test asserting a field with available_in_api=False is excluded (the  details-serializer path has one, test_partner_fields_hidden_for_api_user_when_unavailable_in_api, but v1 doesn't). A small test would close the gap

src/django/api/serializers/facility/facility_download_serializer.py
```diff
@@ -55,3 +125,112 @@ def get_contributors(self, facility: FacilityIndexNewManager) -> str:
...
+                row.extend(build_object_field_cells(entries, paths, separator))
+            else:
+                row.append(build_primitive_field_cell(entries, separator))
+
+        return row
+
+    def get_mit_living_wage_headers(self) -> List[str]:
+        """Return MIT living wage columns when that system partner field is
+        active."""
+        if "mit_living_wage" not in self.__system_partner_fields:
```
daniellekain: optional: The system-field names are hardcoded here ("mit_living_wage"/"wage_indicator", also 218/225/234), and the same provider-name matching is duplicated in facility_index_details_serializer.py:393 and production_locations.py:349. Tests cover it, so not a blocker— but deriving these names from system_partner_field_registry instead of literals would remove the duplication.

daniellekain (changes requested):
thanks for adding this functionality -- it generally looks good to me! I had a question about a potential behavior change, and some nits

valencia-max (dismissed):
LGTM!

VadimKovalenkoSNF (dismissed):
Good work! A note from me:

Migration 0213 adds both flags with `default=True`, and there is no data migration that flips the relevant field. We must not forget to enable those flags manually after release if you don't plan to create a separate migration for this.

Pls, address comments from @daniellekain / Coderabbit and ping me to re-approve.

src/django/api/models/partner_field_admin.py
```diff
@@ -0,0 +1,174 @@
...
+    def clean(self):
+        """
+        Validate that protected fields of system fields are not modified.
+        """
+        cleaned_data = super().clean()
+
+        if self.instance and self.instance.pk and self.instance.system_field:
+            try:
+                original = PartnerField.objects.get(pk=self.instance.pk)
+
```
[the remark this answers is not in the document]
protsack-stephan: Not relevant.

src/django/api/models/partner_field_admin.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        if obj and obj.system_field:
+            messages.warning(
+                request,
+                f'Partner field \'{obj.name}\' cannot be deleted because it '
+                'is a system-defined field.'
+            )
+            return False
+
+        return super().has_delete_permission(request, obj)
```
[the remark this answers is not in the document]
protsack-stephan: Not relevant.

src/django/api/serializers/facility/facility_download_serializer.py
protsack-stephan: Yes, that's intentional. We want to control all of them from the admin panel. Although they have special logic associated with them, we want to make sure we can turn them off and on. Thanks for checking in!

src/django/api/serializers/facility/facility_download_serializer.py
protsack-stephan: Fair point, let me take a look.

src/django/api/serializers/facility/facility_index_details_serializer.py
protsack-stephan: Yes, that is intentional. Thanks for checking in!

src/django/api/serializers/facility/facility_index_details_serializer.py
protsack-stephan: Good point, I think this is in no way obvious why this if statement is here. Let me update this.

src/django/api/serializers/facility/facility_index_details_serializer.py
```diff
@@ -370,25 +369,54 @@ def get_is_claimed(self, facility):
...
+            field for field in fields
+            if field.active
+        ]
+
+        is_api_user = request is not None and request.auth
+        if is_api_user:
+            partner_fields = [
+                field for field in partner_fields
+                if field.available_in_api
+            ]
```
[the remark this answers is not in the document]
protsack-stephan: That's fine.

src/django/api/views/v1/production_locations.py
protsack-stephan: Thanks for flagging! It's definitely worth adding one.

src/django/api/models/partner_field.py
protsack-stephan: Good point, let me update that.

src/django/api/serializers/facility/facility_download_serializer.py
protsack-stephan: I don't want to change the logic too much in the scope of this PR. I've updated the logic so that they are being referenced directly from the providers instead of being hardcoded.

src/django/api/serializers/facility/facility_index_details_serializer.py
protsack-stephan: Updated, added a comment.

src/django/api/views/v1/production_locations.py
protsack-stephan: Updated.

src/django/api/models/partner_field.py
protsack-stephan: Uupdaed.

src/django/api/views/v1/production_locations.py
daniellekain: thank you!!

daniellekain (approved):
looks great to me! Thanks so much for being so thorough with tests, comments, responding to clarifications, etc!

VadimKovalenkoSNF (approved):
Re-approved.

PR #1072, opened 2026-06-16
[OSDEV-2786] Exclude Union-type data from programmatic facilities API

VadimKovalenkoSNF:
## Summary

Exclude trade union-linked data from the programmatic `GET /api/facilities/` endpoint while keeping it fully visible to the OS Hub web client (manual search, map, and facility profiles).

Trade union-linked datasets are identified by contributor type (`contrib_type='Union'`). A request is treated as programmatic — and has union facilities filtered out — when it either:
- authenticates with an API token (`Authorization: Token …`), or
- is **not** recognized as genuine web-client traffic (matching `X-OAR-Client-Key` + an allowed `Referer`).

The two signals are intentionally OR-combined so that neither a tokenless scraper nor a token caller spoofing the (public) web-client headers can slip through.

## Why this approach

- **No new schema / no reindex.** Exclusion is applied at query time via a new `exclude_trade_union` option on `FacilityIndex.filter_by_query_params`, keyed off the existing GIN-indexed `contributors_id` / `contrib_type` data. Facility profiles (`retrieve`) are intentionally left unfiltered so union data stays visible on the FE.
- **FE-vs-BE discriminator reuses proven signals.** `X-OAR-Client-Key` + `Referer` are already required by the global `IsAuthenticatedOrWebClient` permission and are explicitly whitelisted for `api/facilities/*` in CloudFront (`deployment/terraform/cdn.tf`), so they reliably reach Django and survive the WAF/CDN — no terraform change needed.

## Shortcomings / scope

- The web-client key is public (shipped to the browser), so the FE-vs-BE distinction is a best-effort heuristic, not a hard security boundary — accepted given the stakeholder threat model.
- This PR covers `GET /api/facilities/` only. Bulk download (`GET /api/facilities-downloads/`) exclusion and the governance "approval to include" override are tracked as follow-up work.

## Changes

- `permissions.py`: add reusable `has_api_token` and `is_web_client_request` predicates.
- `facilities_view_set.py`: exclude union data when `has_api_token(request) or not is_web_client_request(request)`.
- `facility_manager_index_new.py`: add `exclude_trade_union` filtering keyed off `contrib_type='Union'`.
- `contributor.py`: add `UNION_CONTRIB_TYPE` constant.
- Release notes entry under 2.26.0.

## Test plan

- [x] `api.tests.test_trade_union_api_exclusion` — manager exclusion, token request excluded, web-client (key + Referer) included, authenticated session without browser headers excluded.
- [x] Manual: union contributor data hidden for token API calls, visible in web app search and on profiles.

Ticket: [OSDEV-2786](https://opensupplyhub.atlassian.net/browse/OSDEV-2786)

[OSDEV-2786]: https://opensupplyhub.atlassian.net/browse/OSDEV-2786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/models/facility/facility_manager_index_new.py
```diff
@@ -213,4 +218,24 @@ def filter_by_query_params(self, params):
...
+        union_contributor_ids = list(
+            Contributor.objects
+            .filter(contrib_type=Contributor.UNION_CONTRIB_TYPE)
+            .values_list('id', flat=True)
+        )
+
+        if not union_contributor_ids:
+            return facilities_qs
+
+        return facilities_qs.exclude(
```
VadimKovalenkoSNF, line 243: Drop every facility whose contributor list contains **at least one** union-type contributor. It operates at the facility (row) level, not the field level. It does not strip the union contributor out of a facility while keeping the rest — it removes the entire facility from the result set.

src/django/api/models/facility/facility_manager_index_new.py
```diff
@@ -213,4 +218,45 @@ def filter_by_query_params(self, params):
...
+
+    @classmethod
+    def exclude_trade_union_linked(cls, facilities_qs):
+        union_contributor_ids = cls._union_contributor_ids()
+
+        if not union_contributor_ids:
+            return facilities_qs
+
+        return facilities_qs.exclude(
+            contributors_id__overlap=union_contributor_ids
```
VadimKovalenkoSNF, line 244: Adding another contributor does not remove the union-linked status. When a non-union contributor adds a contribution, the facility's contributors_id becomes [union_id, other_id, ...]. The union id is still present, the arrays still overlap, so the facility stays excluded from the API and downloads.

src/django/api/views/facility/facilities_view_set.py
```diff
@@ -256,6 +271,16 @@ def list(self, request):
 
         queryset = queryset.extra(order_by=order_list)
 
+        if exclude_trade_union or can_access_union:
+            excluded_from_download_count = 0
+        else:
+            excluded_from_download_count = (
```
VadimKovalenkoSNF, line 277: This is an extra query that can slow down a bit main search to addition ~10-20 ms, served entirely from the GIN index, no full scan.

PR #1073, opened 2026-06-16
[OSDEV-2878] Backfill os_id_snapshot for existing moderation events

daniellekain:
## What

Backfills `ModerationEvent.os_id_snapshot` for events that predate the write-at-approval logic added in #1052. Split out of #1052 so the schema change + consumer wiring can ship independently of this slower, investigative data backfill.

- **`0214_backfill_moderation_event_os_id_snapshot`** — data migration that forward-fills `os_id_snapshot = os_id` for approved events where the snapshot is empty but `os_id` is still present (~298k rows). Idempotent, no-op reverse.
- **`backfill_moderation_event_os_id_snapshot_recovery`** — **best-effort** management command that recovers `os_id_snapshot` for events whose `os_id` was already nulled by the `SET_NULL` cascade (~36k), from the linked **`FacilityListItem`**. Per-UUID error handling; supports `--dry-run`.

## Recovery coverage (measured on the Apr 2026 prod dump)

The forward-fill is solid (~298k). The **recovery is largely infeasible** — most original OS IDs are simply no longer in any reliable source. Of **36,007** target events (`APPROVED`, `os_id IS NULL`, empty snapshot):

| Source | Recoverable | Why |
|---|---|---|
| `FacilityListItem` link | 16 (~0.04%) | The `moderation_event` OneToOne link was added in 2.21.0 with no backfill — only 16 of the 36k have it. **This is the recovery source.** |
| OpenSearch | 0 | Prod index was cleared and refilled from Postgres, so it mirrors the NULLs. **Removed** — recovered nothing. |
| `cleaned_data` | 0 | Holds only pre-approval contribution data (name/address/sector); no OS ID. |
| `FacilityAlias` | 0 usable | Maps a *known* os_id → surviving facility; can't reconstruct an orphaned event's os_id. |

**Command total: ~16 / 36,007.** Recovery is **FacilityListItem-only** — the OpenSearch path was dropped since it recovered nothing in production (simpler command, same coverage). The `deleted_os_id` recorded in `FacilityListItem.processing_results` on delete is a sizable pool (~24,968), but those rows aren't joined to the moderation events (same missing 2.21.0 link) — associating them would require a fuzzy `cleaned_data` heuristic (contributor + name + address + country), which isn't trustworthy for an immutable audit field. Captured as a possible future spike, not done here.

Recovery is clearly **labeled best-effort** (command `help` + release notes); the bulk of the 36k backlog is expected to remain unrecoverable.

## Why

`ModerationEvent` is an audit record, but the `os` FK is `on_delete=SET_NULL`, so a facility delete/merge silently nulls `os_id`. #1052 adds the durable `os_id_snapshot` and starts populating it at approval going forward; this PR populates the forward-fillable backlog and recovers what little of the nulled backlog still has a reliable source.

## Notes

- **Stacked on #1052** (base = `feature/OSDEV-2696-add-os-id-snapshot-to-moderation-event`). `0214` depends on `0213` from that branch. Retarget to `main` once #1052 merges.

## Ticket

- [OSDEV-2878](https://opensupplyhub.atlassian.net/browse/OSDEV-2878) (sub-task of [OSDEV-2696](https://opensupplyhub.atlassian.net/browse/OSDEV-2696))

[OSDEV-2878]: https://opensupplyhub.atlassian.net/browse/OSDEV-2878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


VadimKovalenkoSNF (dismissed):
Approved.

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
```diff
@@ -0,0 +1,203 @@
...
+    def __get_os_id_map_from_opensearch(self, target_uuids, opensearch):
+        """
+        Fetch os_id values from OpenSearch for the given set of UUIDs.
+        Returns a dict of {uuid_str: os_id}.
+        """
+        os_id_map = {}
+        uuid_list = [str(u) for u in target_uuids]
+
+        for i in range(0, len(uuid_list), BATCH_SIZE):
+            batch = uuid_list[i:i + BATCH_SIZE]
```
valencia-max: I like this idea of doing this in batches! Just for curiosity, do we know what is the limit of objs that we can fetch from OpenSearch per query?

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
```diff
@@ -0,0 +1,203 @@
...
+)
+
+from api.models.facility.facility_list_item import FacilityListItem
+from api.models.moderation_event import ModerationEvent
+from api.services.opensearch.opensearch import OpenSearchServiceConnection
+from api.views.v1.index_names import OpenSearchIndexNames
+
+log = logging.getLogger(__name__)
+
+BATCH_SIZE = 100
```
valencia-max: Maybe this variable could be a parameter in the command

valencia-max (dismissed):
Amazing work!

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
```diff
@@ -0,0 +1,203 @@
...
+        self.stdout.write(
+            f'Found {total} approved events with null os_id and empty '
+            'os_id_snapshot to recover.'
+        )
+
+        if total == 0:
+            self.stdout.write(self.style.SUCCESS('Nothing to recover.'))
+            return
+
+        opensearch = OpenSearchServiceConnection()
```
protsack-stephan: **Nitpick:** This could potentially be moved to the constructor to avoid passing it as an argument into `__get_os_id_map_from_opensearch`. Alternatively, we could define it within the function itself, as I don't see `__get_os_id_map_from_opensearch` being used anywhere else in the code.

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self.stdout.write('Querying OpenSearch for os_id values...')
+        os_id_map = self.__get_os_id_map_from_opensearch(
```
protsack-stephan: Have we checked whether it makes sense to backfill from OpenSearch at all? I think we had this discussion in the previous PR, so I'm just wondering what we ended up deciding.

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
```diff
@@ -0,0 +1,203 @@
...
+            try:
+                rows = ModerationEvent.objects.filter(
+                    uuid=uuid,
+                    os_id__isnull=True,
+                    os_id_snapshot='',
+                ).update(os_id_snapshot=recovered_os_id)
+                updated += rows
+            except Exception as e:
+                failed_uuids.append(uuid)
+                log.error('Failed to update event %s: %s', uuid, e)
```
protsack-stephan: Can or should we batch these at all? Do we foresee this script taking a long time to run in production?

protsack-stephan (commented):
Good work! Couple of comments from me.

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
daniellekain: Good call — I measured it against the Apr 2026 prod dump: OpenSearch recovered **0 of ~36k** (the prod moderation-events index was cleared and refilled from Postgres, so it mirrors the NULLs); only the `FacilityListItem` link recovered anything (~16). So I dropped the OpenSearch path entirely — recovery is now FacilityListItem-only (72615198). That also removes the batching + the extra connection, so it resolves the related comments below.

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
daniellekain: Moot now — I removed the OpenSearch path (it recovered 0 in prod), so the batching is gone. For reference though: OpenSearch caps a single query at ~10k results (`index.max_result_window`), and we requested 100 per batch — well under. Thanks!

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
daniellekain: Moot now too — `BATCH_SIZE` went away with the OpenSearch path. Thanks for the suggestion!

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
daniellekain: Resolved by removing the OpenSearch path entirely — `__get_os_id_map_from_opensearch` (and the connection it took as an arg) are gone.

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
daniellekain: The update loop only iterates the *recoverable* entries (~16 on the prod dump), each a single PK-scoped `.update()`, so it runs fast — no batching needed. If that set ever grew large, `bulk_update` would be the move.

doc/release/RELEASE-NOTES.md
```diff
@@ -33,10 +34,13 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Release instructions
 * Ensure that the following commands are included in the `post_deployment` command:
-    * `migrate`
+    * `migrate` — runs `0215_backfill_moderation_event_os_id_snapshot`, which forward-fills `os_id_snapshot` for the ~298k approved events where `os_id` is still present in Postgres.
```
vlad-shapik: Let’s not focus on this migration here, as it also runs other migrations.

doc/release/RELEASE-NOTES.md
```diff
@@ -33,10 +34,13 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
     * `reindex_database`
 * The `moderation-events` OpenSearch index template changed (new `os_id_snapshot` field + `os_id` fallback). Recreate/reindex the `moderation-events` index so existing documents pick up the new mapping and coalesced `os_id`.
-* Backfilling `os_id_snapshot` for existing moderation events is handled separately in [OSDEV-2878](https://opensupplyhub.atlassian.net/browse/OSDEV-2878).
+* Run the following backfill command after deployment:
+    * `python manage.py backfill_moderation_event_os_id_snapshot_recovery` — **best-effort** recovery of `os_id_snapshot` for the ~36k events where `os_id` was already nulled, from the linked `FacilityListItem`. Run with `--dry-run` first. Note: coverage of this historical backlog is very low (~16 of ~36k on the Apr 2026 prod dump) — the `FacilityListItem`↔event link was only added in 2.21.0 with no backfill, so the majority of these events have no usable link and stay unrecoverable.
+
+---
```
vlad-shapik: Do we need it?

doc/release/RELEASE-NOTES.md
```diff
@@ -33,10 +34,13 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
-    * `migrate`
+    * `migrate` — runs `0215_backfill_moderation_event_os_id_snapshot`, which forward-fills `os_id_snapshot` for the ~298k approved events where `os_id` is still present in Postgres.
     * `reindex_database`
 * The `moderation-events` OpenSearch index template changed (new `os_id_snapshot` field + `os_id` fallback). Recreate/reindex the `moderation-events` index so existing documents pick up the new mapping and coalesced `os_id`.
-* Backfilling `os_id_snapshot` for existing moderation events is handled separately in [OSDEV-2878](https://opensupplyhub.atlassian.net/browse/OSDEV-2878).
+* Run the following backfill command after deployment:
+    * `python manage.py backfill_moderation_event_os_id_snapshot_recovery` — **best-effort** recovery of `os_id_snapshot` for the ~36k events where `os_id` was already nulled, from the linked `FacilityListItem`. Run with `--dry-run` first. Note: coverage of this historical backlog is very low (~16 of ~36k on the Apr 2026 prod dump) — the `FacilityListItem`↔event link was only added in 2.21.0 with no backfill, so the majority of these events have no usable link and stay unrecoverable.
```
vlad-shapik: The easiest method, which is already in place for running a command like this in a real environment, is to include it in the existing `post_deployment` Django command and after the release of that version, just remove it from there.

The mention of this command should go under "Ensure that the following commands are included in the `post_deployment` command:"

src/django/api/migrations/0216_backfill_moderation_event_os_id_snapshot.py
```diff
@@ -0,0 +1,33 @@
...
+        os_id_snapshot='',
+        os_id__isnull=False,
+    ).update(os_id_snapshot=F('os_id'))
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0213_add_os_id_snapshot_to_moderation_event'),
+    ]
```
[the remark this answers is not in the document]
vlad-shapik: cc @daniellekain ^^

vlad-shapik (changes requested):

doc/release/RELEASE-NOTES.md
daniellekain: Good call — reverted to plain `migrate`, since it runs every pending migration rather than just this one.

doc/release/RELEASE-NOTES.md
daniellekain: Thanks! I folded `backfill_moderation_event_os_id_snapshot_recovery` into the `post_deployment` command list with a note to remove it after this release, and dropped the separate manual section. Kept it (rather than cutting it) so the ~16 recoverable rows get picked up automatically on this deploy; it's a no-op on every run afterward. Let me know if you had something else in mind or would rather I dropped it altogether.

src/django/api/migrations/0216_backfill_moderation_event_os_id_snapshot.py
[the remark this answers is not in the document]
daniellekain: Thanks for the catch! The renumber to `0214` had only landed in my working tree and never got staged, so the committed migration still pointed at the deleted `0213`. Fixed in `0d49cb9f` — it now depends on `0214_add_os_id_snapshot_to_moderation_event`.

src/django/api/management/commands/backfill_moderation_event_os_id_snapshot_recovery.py
```diff
@@ -0,0 +1,129 @@
...
+
+        updated = 0
+        failed_uuids = []
+        for uuid, recovered_os_id in recovered_map.items():
+            try:
+                rows = ModerationEvent.objects.filter(
+                    uuid=uuid,
+                    os_id__isnull=True,
+                    os_id_snapshot='',
+                ).update(os_id_snapshot=recovered_os_id)
```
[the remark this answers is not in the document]
daniellekain: Added `status=ModerationEvent.Status.APPROVED` to the update filter so a row that leaves APPROVED between the candidate scan and the write is skipped. Thanks.

src/django/api/migrations/0216_backfill_moderation_event_os_id_snapshot.py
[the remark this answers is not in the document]
daniellekain: You were correct — the committed migration did depend on the nonexistent `0213`. Fixed in `0d49cb9f`; it now targets `0214_add_os_id_snapshot_to_moderation_event`.

doc/release/RELEASE-NOTES.md
daniellekain: Good catch — removed it. It was the only `---` in the file (other release sections don't use a separator), so it slipped in by mistake.

doc/release/RELEASE-NOTES.md
daniellekain: Followed up on this — beyond the release-notes mention, I've added the command to the actual `post_deployment` management command (`migrate` → `reindex_database` → recovery), with a comment to remove it after 2.26.0, so it runs automatically rather than as a manual step.

vlad-shapik: Thank you!

vlad-shapik (dismissed):
Approved!

vlad-shapik (approved):
re-approved 👍

valencia-max (approved):

valencia-max:
Amazing work!!

PR #1076, opened 2026-06-18
OSDEV-2197: SLC Contributed Data Contribution Disappears When Field is Promoted

valencia-max:
## Jira Ticket

[OSDEV-2197](https://opensupplyhub.atlassian.net/browse/OSDEV-2197)

---

## Summary of Changes

When an SLC reuses the existing name/address unchanged and is then promoted, the production location profile kept attributing the name/address to the original contributor instead of the promoting one.

Root cause: the profile resolves the name/address contributor by matching the displayed value and taking the first matching entry. The index data behind it had no notion of which contribution was promoted (`created_from`), so when two contributions share the same value, the older one always won the tie.

Fix: flag the promoted (`created_from`) contribution in the facility name/address index functions and give it priority when ordering those entries, so attribution follows the promoted contribution. Promotion itself already worked; this only corrects downstream attribution. No UI change.

---

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] UI / frontend change

---

## Testing

- Added a regression test: two contributions with identical name/address, promote the second, assert the promoted contributor is attributed first for both name and address.
- Manually verified: SLC with unchanged name/address → approve → promote now attributes the name/address to the promoting contributor; the modified-value case is unaffected.

Before promoting;

OSDEV-2197-before

After promoting;

OSDEV-2197-after

Both contributions appears:
image

---

## Known TODO Items

- Facilities promoted before this change need a reindex to show corrected attribution; new promotes self-heal automatically.
- A promoted contribution intentionally does not override a claimed or verified name/address.

---

### Implementation

### Testing & Validation

- [] All existing tests pass locally.

### Final Review

[OSDEV-2197]: https://opensupplyhub.atlassian.net/browse/OSDEV-2197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


VadimKovalenkoSNF:
Hi @valencia-max ! Thanks for the PR, I have a couple of questions here:
1. **Ticket scope mismatch**. The ticket description and AC still describe the extended-fields problem (parent_company, etc.), but per comments 30996/30999 the scope was re-clarified to name/address attribution — which is what this PR actually fixes. Can the description/AC be updated to the clarified scope before merge? As-is, the PR and ticket text disagree, which makes review/QA ambiguous.
2. Name/address are served from FacilityIndex via the `index_facility_*` functions, so `is_from_created_from` only lands after a reindex. Does `promote()` reindex the facility synchronously? And what's the backfill plan for already-promoted facilities mentioned in the TODO?
3. Francesca's clarified description says the contribution disappears "including from the additional data source panel." Does this fix restore the SLC contributor in that panel too, or only in the primary name/address attribution? A screenshot/test covering the panel would confirm it.

valencia-max:
> 2\. Does `promote()` reindex the facility synchronously? And what's the backfill plan for already-promoted facilities mentioned in the TODO?

Thanks fro review, @VadimKovalenkoSNF!
1. Yes, I just updated the ticket.
2. At the DB level, `facility.save()` runs an `UPDATE api_facility ... (new created_from_id, name, address, etc.)`. That UPDATE fires the `AFTER UPDATE` trigger` facility_post_update_insert_indexing_trigger`, which calls `index_facilities_by([facility.id])` to rebuild that facility's `api_facilityindex` row (including facility_names/facility_addresses with the `is_from_created_from` flag). All in the same transaction, so it commits atomically.
Yes, this PR is only fixing new promotions, I will create a PR for a backfill strategy on previous promoted facilities.
3. Taking a look! @VadimKovalenkoSNF done

valencia-max:
@VadimKovalenkoSNF, to backfill already-promoted facilities:

Running `index_facilities_new` (with migrate applied first) it rebuilds the `FacilityIndex` table so historically promoted facilities pick up the new `is_from_created_from` flag and show the correct name/address attribution:

Trade-offs / implications

- `index_facilities_new` with no args reindexes all facilities — it's a full rebuild, so it's heavy and time-consuming on production-sized data.
- Alternative if a full rebuild is too costly: a targeted reindex of only promoted facilities (`index_facilities_by(ids)` for facilities whose `created_from` item has a promote_match marker). Creating a command and only execute reindex for those already promoted facilities.

src/django/api/migrations/0215_attribute_promoted_contribution_name_address.py
vlad-shapik, line 1: Question: do we need to backfill historically promoted facilities?

From my reading, this fix is safe to deploy without touching existing api_facilityindex rows - the API won't fail. The serializer uses .get('is_from_created_from', False), so old JSON without that key just falls back to today's behavior. New promotions self-heal via the post-update trigger that calls index_facilities_by([id]).

The open question is whether we need to correct attribution for facilities that were promoted before this change (SLC submitted unchanged name/address, then promoted). Without a backfill, those records would keep showing the original contributor as the source.

If the answer is yes, I'd strongly prefer not running a full index_facilities_new on production and including it in post_deployment Django command. With ~2.5M facilities, that rebuilds every indexed column for every row and can run for hours - and we don't have a reliable runtime estimate for prod at that scale.

Instead, we could use the targeted backfill infrastructure in [#1077](https://github.com/opensupplyhub/open-supply-hub/pull/1077), which adds the facility_index_backfill package and backfill_facility_index management command. That approach refreshes only the affected columns (facility_names, facility_addresses) via the updated index_facility_names() / index_facility_addresses() functions, without recomputing the rest of the index row. On prod, a similar targeted backfill for the contributors column completed in ~3.5 minutes with --parallel 10.

Could you confirm:

Is correcting historical attribution in scope for this ticket, or is fixing new promotions enough?
If yes, let’s revisit this PR once [the aforementioned work ](https://github.com/opensupplyhub/open-supply-hub/pull/1077)is merged, probably tomorrow, and then you can simply update this PR to correct the already existing records.

vlad-shapik (changes requested):
Could these changes be assigned to the next release, 2.27.0, which is scheduled for July 11?
I do not think we will make it for 2.26.0, as we already have some chunks of work merged and not yet tested for this week’s release. Some of them may still be included, but they are still awaiting comment handling.

src/django/api/migrations/0215_attribute_promoted_contribution_name_address.py
valencia-max: Thanks for your thoughts, @vlad-shapik ! I'll confirm this with Francesca but my actual answer it's **NO**, we don't need to modify the existing facilities now, I'd say we can fix this issue this sprint (2.26.0) and work in a separate ticket to backfill the previous promoted facilities in the next one (2.27.0).

vlad-shapik: Ok, then let's merge it now

vlad-shapik: Could you handle merge conflicts?

vlad-shapik: You will likely need to re-tie the migration you created to a newer one and rename your migration and SQL files so they have a higher number prefix. There are already migrations with the 0213 and 0214 prefixes.

valencia-max: @vlad-shapik , On it!

valencia-max: Done, @vlad-shapik !

image

vlad-shapik (approved):
LGTM 🚀

daniellekain (approved):
looks good to me!

daniellekain:
Heads up @valencia-max — I'm working on #1073, which also adds a `0215` migration, so there's a slight chance we may need to renumber one of them depending on which merges first. Whichever lands second would bump to `0216` (and for this PR, the two `0215_index_facility_*.sql` files + `run_sql_files([...])` references would come along too). No action needed now — just flagging so it's on the radar.

PR #1077, opened 2026-06-18
[OSDEV-2390] Add contribution dates to Supply Chain Network drawer

vlad-shapik:
## Summary
- Add `list_uploaded_at` and `last_contributed_at` to `index_contributors()` (migration `0217`) and expose them via `FacilityIndexDetailsSerializer.get_contributors()` for public and anonymized contributor entries.
- Introduce the `facility_index_backfill` package with `BackfillOrchestrator` and `BackfillWorker` to refresh selected `FacilityIndex` fields in parallel using existing `index_*()` SQL functions. The orchestrator command is `backfill_facility_index`; parallel workers run via `backfill_facility_index_worker` subprocesses with explicit CLI args (no env-var dispatch).
- Wire post-deployment to run `migrate`, `reindex_database`, and `backfill_facility_index --fields contributors --parallel 10 --batch-size 10000`.
- Show contribution dates in the Production Location Supply Chain Network drawer: per-list upload dates for list contributions, and "Last contributed" for API-only public contributors and anonymized contributor types.
Screenshot 2026-06-23 at 16 43 04

## Deployment
Post-deployment runs automatically via `post_deployment.py`:
1. `migrate`
2. `reindex_database`
3. `backfill_facility_index --fields contributors --parallel 10 --batch-size 10000`

Expect the contributors backfill to add moderate RDS load (~+30% CPU, ~+10 connections for 10 workers) for roughly 3–4 minutes with no application downtime. See `src/django/api/facility_index_backfill/README.md` for operational notes.

## Test plan
- [ ] Open a production location with public list contributors and confirm list upload dates appear under each list
- [ ] Open a production location with API-only contributors and confirm "Last contributed" appears when the contributor has no lists
- [ ] Open a production location with both list and API matches for the same contributor and confirm only list upload dates are shown (no header last contributed)
- [ ] Confirm anonymized contributor types show "Last contributed"
- [ ] With `enable_production_location_page` disabled, open `/facilities/:osID` and confirm the legacy contributors drawer still lists contributor names and list names
- [ ] Deploy to Test and verify dates populate for previously indexed facilities after post-deployment completes
- [ ] Run Django tests: `api.tests.test_facility_index_backfill_specs`, `api.tests.test_facility_index_backfill_runner`, `api.tests.test_facility_index_backfill_integration`, `api.tests.test_facility_index_details_serializer`
- [ ] Run React tests: `ContributionDate.test.jsx`, `splitContributorsIntoPublicAndNonPublic` in `utils.tests.js`


src/react/src/components/ProductionLocation/Shared/ContributionDate/styles.js
```diff
@@ -0,0 +1,17 @@
+import COLOURS from '../../../../util/COLOURS';
+
+export default () =>
+    Object.freeze({
+        dateWithIcon: Object.freeze({
+            display: 'inline-flex',
+            alignItems: 'center',
+            fontSize: '16px',
```
VadimKovalenkoSNF, line 8: nitpick: I believe we should use `rem` units or extend `getTypographyStyles` here.

src/react/src/util/util.js
```diff
@@ -1021,28 +1021,46 @@ export const splitContributorsIntoPublicAndNonPublic = contributors =>
...
-                    /*
-                    Push the contributor to the array of public contributors if they don't
-                    exist there. Also, the code replaces the list_name property with
-                    list_names to make the object contain data more related to the
-                    contributor and their lists, not one contribution of the contributor.
-                    */
                     const {
                         list_name: listName,
-                        ...contributorWithoutListName
+                        list_uploaded_at: listUploadedAt,
```
VadimKovalenkoSNF, line 1026: Optional: It would be great to have a test suite for `splitContributorsIntoPublicAndNonPublic` utility function in `src/react/src/__tests__/utils.tests.js`

VadimKovalenkoSNF (dismissed):
Good work! I left few minor comments, feel free to skip.
Also, worth checking this investigation from AI agent:

**Bug**: anonymized "Last contributed" isn't the most recent date (AC #4)
In get_contributors, the anonymized date is captured inside the dedup-by-id block, so only the first-processed row per contributor is considered. Since index_contributors() emits one row per match (no ORDER BY) and rows sort by (id, fl_id or 0) with no date tiebreaker, "first" is typically the oldest contribution, not the newest.

**Repro**: anonymized contributor contributes via API (2020) and a later list (2024) → drawer shows 2020, AC #4 wants 2024. The frontend `mergeLastContributedAt` can't recover it since the backend already dropped the newer date. The public API-only path (AC #2) has the same issue when a contributor has multiple API matches.

**Fix**: separate counting (dedup by id) from date aggregation (max over all rows of the type), keeping the public/anonymous exclusion intact. Please add a serializer test for a contributor with multiple contributions — current tests don't assert on dates at all.

src/django/api/facility_index_backfill/runner.py
```diff
@@ -0,0 +1,281 @@
...
+
+        elapsed = time.monotonic() - started_at
+        self.stdout.write(
+            self.style.SUCCESS(
+                f'Worker {worker_label} ({field_name}) finished: '
+                f'{total_updated} rows updated in {elapsed:.1f}s.'
+            )
+        )
+        self._write_worker_result(total_updated)
+        return total_updated
```
protsack-stephan: This runner serves two purposes now: it orchestrates the parallelism and handles the subprocess-related work at the same time. I suggest we refactor this into two distinct classes. One will be responsible for the orchestration of subprocesses (`BackfillOrchestrator`), and the other will contain the logic for the specific worker (`BackfillWorker`). That way, we won't have to set environment variables and pass them into the subprocess; instead, we can just have two different commands - one for the worker and one for the orchestrator. Also, potentially the params validation can happen in the command and the classes can be responsible only for the logic.

For example, the main command can stay the same:

```bash
backfill_facility_index
  --fields <csv>          required   e.g. "contributors" or "contributors,claim_info"
  --parallel <N>          default 1  number of worker subprocesses per field group
  --batch-size <N>        default 10000
  --dry-run               flag       row counts only, no writes
```

But we can have another command that is called from within the main command, looking something like this:

```bash
backfill_facility_index_worker
  --field <name>          required   single field group (not csv)
  --worker-id <i>         required   0-based, 0 <= i < workers
  --workers <N>           required   total partitions (>= 1)
  --batch-size <N>        default 10000
  --dry-run               flag
  --result-file <path>    optional   where to write final row count for parent
```

In terms of code, it could look something like this:

```python
class BackfillOrchestrator:
    def run(field_names, parallel, batch_size, dry_run):
        pass


class BackfillWorker:
    def run(field, worker_id, workers, batch_size, dry_run, result_file):
        pass

# commands/backfill_facility_index.py
class Command(BaseCommand):
    # add_arguments: --fields, --parallel, --batch-size, --dry-run
    def handle(self, **opts):
        orchestrator = BackfillOrchestrator(self.stdout, self.style)
        orchestrator.run(
            fields, opts["parallel"], opts["batch_size"], opts["dry_run"])

# commands/backfill_facility_index_worker.py
class Command(BaseCommand):
    # add_arguments: --field, --worker-id, --workers, --batch-size,
    #                --dry-run, --result-file
    def handle(self, **opts):
        worker = BackfillWorker(self.stdout, self.style)
        worker.run(
            opts["field"], opts["worker_id"], opts["workers"],
            opts["batch_size"], opts["dry_run"], opts["result_file"])
```

protsack-stephan (commented):
Good work! A couple of things I'd like to mention:

- It would be great to add type annotations whenever possible to make the code easier to refactor.
- It would be great to add docstrings to the methods and classes to make the code easier to read for both LLMs and humans.

src/django/api/facility_index_backfill/specs.py
```diff
@@ -0,0 +1,85 @@
...
+    from_clause = spec.get('from_clause', 'FROM api_facilityindex afi')
+    id_column = spec.get('id_column', 'afi.id')
+    filter_sql = spec.get('filter_sql')
+    filter_part = f'AND {filter_sql}' if filter_sql else ''
+
+    return f"""
+SELECT COUNT(*)
+{from_clause}
+WHERE mod(abs(hashtext({id_column}::text)), %(workers)s) = %(worker_id)s
+  {filter_part}
```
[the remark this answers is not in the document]
vlad-shapik: @coderabbitai Addressed locally: both partition predicates in `specs.py` now use `abs(hashtext({id_column}::text)::bigint)` before `mod(...)`, and `test_facility_index_backfill_specs.py` asserts the cast in `build_count_sql` / `build_update_sql`. All 23 backfill tests pass. Will push in the next commit.

src/django/api/management/commands/backfill_facility_index.py
```diff
@@ -0,0 +1,63 @@
...
+            help=(
+                'Comma-separated field groups to backfill. '
+                f'Available: {", ".join(list_field_names())}.'
+            ),
+        )
+        parser.add_argument(
+            '--batch-size',
+            type=int,
+            default=10000,
+            help='Number of rows to update per batch (default: 10000).',
```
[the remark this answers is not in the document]
vlad-shapik: @coderabbitai Addressed locally: `handle()` in `backfill_facility_index.py` now raises `CommandError` when `--batch-size` is less than 1. Added `test_batch_size_must_be_at_least_one` in `test_facility_index_backfill_specs.py`. All 24 backfill tests pass. Will push in the next commit.

src/react/src/components/ProductionLocation/Sidebar/SupplyChain/SupplyChainNetworkDrawer/SupplyChainNetworkDrawer.jsx
```diff
@@ -69,13 +69,20 @@ SupplyChainNetworkDrawer.propTypes = {
             id: number.isRequired,
             contributor_name: string.isRequired,
             contributor_type: string,
-            list_names: arrayOf(string),
+            lists: arrayOf(
+                shape({
+                    name: string.isRequired,
+                    uploaded_at: string,
+                }),
```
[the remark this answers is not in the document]
vlad-shapik: @coderabbitai Addressed: `uploaded_at` is now `string.isRequired` in the `lists` shape in `SupplyChainNetworkDrawer.propTypes`, aligned with `PublicContributorsSection` always rendering `ContributionDate` for list rows. Will push in the next commit.

src/react/src/components/ProductionLocation/Sidebar/SupplyChain/utils.js
```diff
@@ -1,5 +1,10 @@
 import PLURAL_MAP from './constants';
 
+const mergeLastContributedAt = (existingDate, incomingDate) =>
+    new Date(incomingDate) > new Date(existingDate)
+        ? incomingDate
+        : existingDate;
```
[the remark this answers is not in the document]
vlad-shapik: @coderabbitai Thanks — skipping this one for now.

On the current API path, `last_contributed_at` is always passed through `format_date()` in `get_contributors()` (`facility_index_serializer.py`), so a missing value would fail serialization rather than reach the frontend as `undefined`/invalid. Anonymized contributors are also already aggregated per `contributor_type` on the backend, so `mergeLastContributedAt()` in `aggregateByType()` rarely runs with multiple rows of the same type from live data.

The more relevant date issue (if any) is backend dedup picking an older timestamp across duplicate contributor rows, which we are tracking separately. Happy to revisit null guards here if the API contract changes to allow missing dates.

src/react/src/components/ProductionLocation/Shared/ContributionDate/styles.js
vlad-shapik: Keeping `px` here — we use pixel values elsewhere in this area of the UI as well, and I prefer them for exact sizing in this component.

src/react/src/util/util.js
vlad-shapik: Added tests for `splitContributorsIntoPublicAndNonPublic` in `src/react/src/__tests__/utils.tests.js` (list split/merge, API-only last-contributed merge, and non-public routing). Will push in the next commit.

src/django/api/facility_index_backfill/runner.py
vlad-shapik: Done in c3e283d6. The refactor follows your suggested shape:

- **`BackfillOrchestrator`** (`backfill_orchestrator.py`) — coordinates field groups, spawns workers when `--parallel > 1`, aggregates results, and prints the summary.
- **`BackfillWorker`** (`backfill_worker.py`) — handles one hash partition (count + batched updates).
- **`backfill_facility_index`** — orchestrator CLI; validation for `--fields`, `--parallel`, and `--batch-size` stays here.
- **`backfill_facility_index_worker`** — worker CLI with explicit `--field`, `--worker-id`, `--workers`, `--batch-size`, `--dry-run`, and `--result-file` args.

The env-var subprocess pattern is removed. When `--parallel > 1`, the orchestrator spawns `backfill_facility_index_worker` subprocesses with CLI args instead of re-invoking the main command with environment variables.

Type annotations and docstrings were added to the backfill package as part of this change. All 26 backfill tests pass.

src/django/api/facility_index_backfill/backfill_orchestrator.py
```diff
@@ -0,0 +1,170 @@
...
+                '--result-file',
+                result_path,
+            ]
+            if dry_run:
+                cmd.append('--dry-run')
+            self.stdout.write(
+                f'  worker {format_worker_number(worker_id)}: '
+                f'{" ".join(cmd)}'
+            )
+            processes.append(subprocess.Popen(cmd))
```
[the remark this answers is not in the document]
vlad-shapik: Thanks — skipping for now.

In production this runs as a **one-off ECS task** (via post-deployment), not inside the long-lived app container. If a later `Popen` fails after earlier workers started, those orphan processes and temp files only live inside that task's filesystem. AWS tears down the task when the command finishes or fails, so nothing leaks on the cluster or host.

The same pattern existed in the previous env-var runner and we've already run this successfully on production RDS. Happy to add explicit spawn cleanup later if we start running parallel backfills in long-lived shells (e.g. local dev), but it's not a practical concern for the ECS deployment path.

src/django/api/facility_index_backfill/backfill_orchestrator.py
```diff
@@ -0,0 +1,170 @@
...
+            )
+            raise CommandError(
+                f'Backfill failed for {field_name}: {details}'
+            )
+
+        total_rows = 0
+        for result_path in result_files:
+            with open(result_path, encoding='utf-8') as result:
+                total_rows += int(result.read())
+            os.unlink(result_path)
```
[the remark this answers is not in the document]
vlad-shapik: Thanks — skipping for now.

Result files are created with `tempfile.mkstemp()` inside the same ephemeral ECS task that runs the orchestrator. If aggregation fails partway through, any unlinked temp files are removed when AWS stops the task anyway.

Workers always write a plain integer row count, so malformed content would only happen on an abnormal worker crash — in which case the deploy task is already in a failed state and gets torn down. A `finally` cleanup would be slightly nicer for local runs, but it's not needed for the production ECS lifecycle.

src/django/api/facility_index_backfill/backfill_orchestrator.py
```diff
@@ -0,0 +1,170 @@
...
+
+class BackfillOrchestrator:
+    """Coordinate backfill jobs across facility index field groups."""
+
+    worker_command_name = 'backfill_facility_index_worker'
+
+    def __init__(self, stdout: OutputWrapper, style: Style) -> None:
+        self.stdout = stdout
+        self.style = style
+        self._worker = BackfillWorker(stdout, style)
```
protsack-stephan: Does it make sense to pull in the worker as a dependency vs. just spawning a subprocess with one worker instead? That way, we don't have to initialize it here and can just wait until the subprocess ends regardless whether there are a single worker or five of them.

protsack-stephan (commented):
Good work, left one comment.

protsack-stephan (dismissed):

src/django/api/facility_index_backfill/backfill_orchestrator.py
vlad-shapik: Done in c40f8963. The orchestrator no longer holds a `BackfillWorker` instance — `_run_field_group` always spawns `backfill_facility_index_worker` subprocesses, including when `--parallel 1`. Worker logic now lives only in the worker command/class; the orchestrator just spawn → wait → aggregate.

protsack-stephan (approved):

VadimKovalenkoSNF (approved):
Re-approved.

PR #1082, opened 2026-06-20
OSDEV-2825: [Homepage] Modify current Terraform routing to link infosite and opensupplyhub domains

valencia-max:
## Jira Ticket
[OSDEV-2825](https://opensupplyhub.atlassian.net/browse/OSDEV-2825)

---

### React — Explicit `/map` route
Previously, the app used `mainRoute = '/'` both as the catch-all route and as the explicit target for "go to the map" navigation. This meant there was no way to distinguish between "the root path" and "the map page", and navigating to `/` would land directly on the map with no redirect.

Introduced a new `mapRoute = '/map'` constant in `util/constants.jsx` and updated all 16 components and navigation elements that used `mainRoute` to mean "the map" to use `mapRoute` instead. `mainRoute` is retained as the catch-all route in `Routes.jsx`.

This separation allows the homepage (Craft CMS) to own `/` without the React app intercepting it, while keeping all internal map navigation explicit.

### Terraform — Per-environment Craft CMS origin domain
Adds a CloudFront reverse proxy configuration to serve the Craft CMS homepage (`info.opensupplyhub.org/`) at `opensupplyhub.org/`, without redirecting the user's browser. The URL stays on `opensupplyhub.org/` while CloudFront fetches and serves the content from the Craft origin transparently.

Changes were made:
- **`deployment/terraform/cdn.tf`**: Adds a `aws_cloudfront_function` resource that rewrites the `URI` part to `/home-page`, and preserving query params on the URL, redirecting to `/map`, a new `origin` block for the Craft origin, and an `ordered_cache_behavior` for `path_pattern = "/"` that proxies the root path to Craft with TTL 0. All three are conditional on `enable_homepage_proxy`.
- **`deployment/terraform/variables.tf`**: Adds the `enable_homepage_proxy` boolean variable, and `craft_cms_origin_domain `variable and set the appropriate value

---

## Type of Change
- [ ] Backend change (API, database, business logic)
- [x] UI / frontend change

---

## Testing
- Verified that navigating to `/` redirects to `/map` in the dev environment.
- Verified that all existing map navigation links (navbar logo, auth menu, mobile nav, sidebar, contribute flow, claim flow) correctly point to `/map`.
- Confirmed that deep links (e.g. `/facilities/123`) still resolve correctly via the catch-all `mainRoute`.
- `EligibilityStep.test.js` updated to assert against `mapRoute` — all existing tests pass.

---

## Known TODO Items
NA

---

### Implementation

### Testing & Validation
- [ ] For backend changes: I have validated API contracts, error handling, and edge cases.
- [x] For architectural changes: I have documented the design decision and notified affected teams.

### Final Review

[OSDEV-2825]: https://opensupplyhub.atlassian.net/browse/OSDEV-2825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployment/terraform/cdn.tf
```diff
@@ -179,6 +210,22 @@ resource "aws_cloudfront_distribution" "cdn" {
     }
   }
 
+  # Craft CMS origin — only present when enable_homepage_proxy = true
+  dynamic "origin" {
+    for_each = var.enable_homepage_proxy ? [1] : []
+    content {
+      domain_name = "open-supply.staging.servd.dev"
```
protsack-stephan: Does it make sense to move this into a variable so that we can quickly change it in the future, or perhaps have a different configuration per environment?

deployment/environments/terraform-development.tfvars
```diff
@@ -75,3 +75,7 @@ app_logstash_fargate_memory = 2048
 instance_source= "os_hub"
 
 vpn_ec2_ami = "ami-0940c95b23a1f7cac"
+
+# Homepage proxy — routes / to Craft CMS (info.opensupplyhub.org)
+# Keep false in staging, production, and rba until fully validated.
+enable_homepage_proxy = true
```
protsack-stephan: Does it make sense to have this flag at all? From what I can see, we are permanently changing the routes on the React side; therefore, we are only controlling the infrastructure with this flag. If we switch it without changing the UI routing, things will break. Please let me know if I am missing something here.

protsack-stephan (commented):
Good work! Couple small comments.

deployment/terraform/cdn.tf
valencia-max: Yes! That make sense

deployment/environments/terraform-development.tfvars
valencia-max: Yes, we need to change the UI routing with no exceptions. If we switch this to `false`, then the reverse proxy won't be applied but the user won't face any issue since the React app is already handling this case, when hitting "/", they will be redirected to "/map", with no new home page.

deployment/terraform/cdn.tf
```diff
@@ -219,6 +261,41 @@ resource "aws_cloudfront_distribution" "cdn" {
...
+        headers      = ["Accept", "Accept-Encoding", "Accept-Language"]
+
+        cookies {
+          forward = "none"
+        }
+      }
+
+      function_association {
+        event_type   = "viewer-request"
+        function_arn = aws_cloudfront_function.homepage_host_rewrite[0].arn
```
daniellekain: really minor nit: I think this is a uri rewrite, not a host rewrite

deployment/terraform/cdn.tf
```diff
@@ -855,7 +929,7 @@ resource "aws_cloudfront_function" "info_openapparel_redirect" {
   code    = <<-EOT
     function handler(event) {
       var qs = event.request.querystring;
-      var location = "https://info.opensupplyhub.org" + event.request.uri + (qs ? "?" + qs : "");
+      var location = var.craft_cms_origin_domain + event.request.uri + (qs ? "?" + qs : "");
```
daniellekain: I don't think variable syntax will work here

daniellekain: I think we may just want to keep this block (and 948/952) as they are?

deployment/terraform/cdn.tf
```diff
@@ -140,13 +140,39 @@ resource "aws_cloudfront_origin_access_identity" "react" {
...
 }
 
+#
+# Homepage proxy — CloudFront Function to rewrite Host header for Craft CMS origin
+#
+# This is used when enable_homepage_proxy = true (dev only for now).
+# The function sets the Host header to info.opensupplyhub.org so Craft serves
+# the correct site when CloudFront reverse-proxies requests from / to it.
+#
+resource "aws_cloudfront_function" "homepage_host_rewrite" {
```
daniellekain: nit: I think this is a URI change vs a header change (here and below)

deployment/environments/terraform-production.tfvars
```diff
@@ -75,4 +75,8 @@ instance_source = "os_hub"
 
 vpn_ec2_ami = "ami-0940c95b23a1f7cac"
 
+# Homepage proxy — set craft_cms_origin_domain when enable_homepage_proxy is enabled
+enable_homepage_proxy   = true
```
daniellekain, line 79: is this supposed to be true/on? I'm a bit confused on the cutover timeline (and we're also turning on in staging/rba?)

src/react/src/Routes.jsx
```diff
@@ -282,6 +283,11 @@ class Routes extends Component {
                                         return <Route component={Homepage} />;
                                     }}
                                 />
+                                <Route
+                                    exact
+                                    path={mainRoute}
```
daniellekain, line 288: I think we may be dropping the query param here?

daniellekain (commented):
Awesome work! A few questions inline. 

One other finding -- I asked Claude to check and see if there were any other places that we needed to update routing and it flagged: 

`src/react/src/components/Footer/Footer.jsx:17` — <Link to="/" …> → <Link to={mapRoute} …>; no constants import exists, add one.
`src/react/src/components/InitialClaimFlow/ClaimForm/ClaimForm.jsx:399` — history.push('/') → history.push(mapRoute); extend the existing import at line 68.
`src/react/src/components/ClaimFacilityStepper.jsx:335` — href="/" → href={mapRoute}; extend the existing import at line 23.
```

File: src/react/src/__tests__/components/ProductionLocationDialog.test.js [hunk 8]
```diff
@@ -211,7 +211,7 @@ describe('ProductionLocationDialog', () => {
         const searchOSHubButton = getByRole('button', { name: /Search OS Hub/i });
         fireEvent.click(searchOSHubButton);
 
-        expect(mockHistoryPush).toHaveBeenCalledWith('/');
+        expect(mockHistoryPush).toHaveBeenCalledWith('/map');
```
daniellekain: nit: should we use mapRoute like below?

deployment/terraform/cdn.tf
daniellekain: my understanding is that these redirects are here so old open apparel links will redirect to os hub links. We still want info.openapparel/XYZ to redirect to info.opensupplyhub/XYZ

deployment/terraform/cdn.tf
valencia-max: Yes, you are right! At the begining I was doing a different thing, I didn't update comments/descriptions, good catch!

deployment/environments/terraform-production.tfvars
valencia-max: Yes! It should be "true", this will only take effect once it's deployed to prod. For staging it should be "true" as well (since we are mirroring prod), and for RBA no, it must be "false".

But now that I'm thinking about this, let's discuss with the team about if all the envs need to have this configuration, probably we just want to do it for prod, I have created this ticket to follow up this: https://opensupplyhub.atlassian.net/browse/OSDEV-2832

deployment/terraform/cdn.tf
valencia-max: Yes! Good catch

deployment/terraform/cdn.tf
valencia-max: Yes! My bad, I got confused with the other function

src/react/src/Routes.jsx
valencia-max: I'll double check! Not pretty sure

deployment/environments/terraform-production.tfvars
daniellekain: sounds good! We may want to update https://github.com/opensupplyhub/open-supply-hub/pull/1082/changes#diff-2efcbf00967b9a228b5bfc7b42952f421711cfff7e92fca25bff15e7ec39a9f0R973

valencia-max:
> Awesome work! A few questions inline.
> 
> One other finding -- I asked Claude to check and see if there were any other places that we needed to update routing and it flagged:
> 
> `src/react/src/components/Footer/Footer.jsx:17` — <Link to="/" …> → <Link to={mapRoute} …>; no constants import exists, add one. `src/react/src/components/InitialClaimFlow/ClaimForm/ClaimForm.jsx:399` — history.push('/') → history.push(mapRoute); extend the existing import at line 68. `src/react/src/components/ClaimFacilityStepper.jsx:335` — href="/" → href={mapRoute}; extend the existing import at line 23.

Actually it's not require to update these routes since the explicit redirection on "/" force every request to go to "/map", but I think it's better to reference the correct route which is "/map", so I updated them to keep consistency!

src/react/src/Routes.jsx
daniellekain: I tested this by rendering the same `<Redirect to={mapRoute} />` pattern in a small react-router MemoryRouter harness, starting from a URL like `/?contributors=2340&lists=9193&embed=1` — and I was seeing the query string get dropped (it lands on `/map` with no params), which would break saved embed/filter links that point at the root path. Changing it to` render={({ location }) => <Redirect to={{ ...location, pathname: mapRoute }} />}` preserved them 
(`/map?contributors=2340&lists=9193&embed=1`). Let me know if that approach doesn't make sense here!

valencia-max: Yes!! It's working! Good catch 🚀

I print the `location` value and here's a more clear answer why it's working, it's only overwriting the `pathname` to be `/map` and then showing the home page component which is served at that endpoint!
image

valencia-max: This change is only one part, Terraform require more changes since this is only covering when the flag is not active or local development. I modify the current rewrite uri function to redirect to `/map` when query params exist in the URL.

deployment/environments/terraform-production.tfvars
```diff
@@ -75,4 +75,7 @@ instance_source = "os_hub"
 
 vpn_ec2_ami = "ami-0940c95b23a1f7cac"
 
+enable_homepage_proxy   = false
+craft_cms_origin_domain = "info.opensupplyhub.org"
```
vlad-shapik: `craft_cms_origin_domain` should point to the real Craft server on Servd (like `open-supply.staging.servd.dev`), not to `info.opensupplyhub.org`, right?

`info.opensupplyhub.org` is the public marketing URL, and there's already a CDN sitting in front of Craft there. dev/test use the real Servd address, but staging/preprod/prod use the `info.` one, so it's inconsistent. The problem: this cache behavior doesn't pass the visitor's `Host` header, so CloudFront sends the origin's own domain as `Host`. If that's `info.opensupplyhub.org`, the request runs into that domain's redirect rules (and Craft can send links back to it), which pushes the browser off `opensupplyhub.org` - or loops if the two domains point at each other. Pointing straight at the Servd server avoids that.

I'd suggest we enable it in **all environments except RBA** so we actually test it everywhere and keep the config the same/consistent across envs. 

Heads up: turning this on will likely break some of the e2e tests in the related repo (they hit `/` expecting the map), so we'll need to update those too.

deployment/environments/terraform-rba.tfvars
```diff
@@ -83,3 +83,6 @@ vpn_ec2_ami = "ami-0940c95b23a1f7cac"
 db_sync_enabled                   = true
 db_sync_schedule_expression       = "cron(0 7 * * ? *)" # (7:00 AM UTC)
 is_database_private_link_consumer = true
+
+enable_homepage_proxy   = false
+craft_cms_origin_domain = ""
```
vlad-shapik, line 89: Small suggestion, feel free to skip: could we add a validation so `craft_cms_origin_domain` can't be empty when the proxy is on? Right now RBA has `craft_cms_origin_domain = ""`, and if someone flips `enable_homepage_proxy = true` there without setting it, CloudFront gets an origin with an empty `domain_name` and the apply fails (or builds something invalid). A guard would make the misconfiguration fail fast with a clear message instead:

```
variable "craft_cms_origin_domain" {
  type    = string
  default = ""
  validation {
    condition     = !var.enable_homepage_proxy || trimspace(var.craft_cms_origin_domain) != ""
    error_message = "craft_cms_origin_domain must be set when enable_homepage_proxy is true."
  }
}
```
(Note: referencing another variable inside `validation` needs a reasonably recent Terraform; if our version doesn't support it, we can enforce the same thing with a `precondition` on the CloudFront resource instead.)

deployment/terraform/cdn.tf
```diff
@@ -140,13 +140,73 @@ resource "aws_cloudfront_origin_access_identity" "react" {
...
+          statusDescription: 'Found',
+          headers: {
+            location: { value: '/map?' + parts.join('&') }
+          }
+        };
+      }
+
+      request.uri = '/home-page';
+      return request;
+    }
```
vlad-shapik: Small optional suggestion: since this CloudFront Function is getting a bit longer (query-string serialization + redirect + URI rewrite), would it make sense to move the JS into its own file and load it with `file(...)` instead of keeping it inline in the heredoc?

Something like:

`code = file("${path.module}/cloudfront-functions/homepage_uri_rewrite.js")`
We already use `file(...)` elsewhere in Terraform for step functions / task definitions, so this would fit that pattern and make the function easier to review and edit.

vlad-shapik (changes requested):

deployment/environments/terraform-rba.tfvars
valencia-max: It's fine, in order to use this feature we need at least terraform 1.9.0, we are using 1.13.3:
image

vlad-shapik (changes requested):
Could you also add an entry to doc/release/RELEASE-NOTES.md for this change before we merge?

vlad-shapik (dismissed):
LGTM
Amazing work!

deployment/terraform/variables.tf
```diff
@@ -967,6 +967,22 @@ variable "enable_legacy_info_site_redirect" {
   default = false
 }
 
+variable "enable_homepage_proxy" {
+  type        = bool
+  default     = false
+  description = "When true, proxies opensupplyhub.org/ to the Craft CMS homepage at info.opensupplyhub.org. Enable in dev for testing only — do not apply to production without further validation."
```
daniellekain: nit: I think this is going to be for dev and production, right?

daniellekain: nit: technically I think we are going directly to servd?

deployment/terraform/cloudfront-functions/homepage_uri_rewrite.js
```diff
@@ -0,0 +1,39 @@
+function handler(event) {
+    var request = event.request;
+    var qs = request.querystring;
+
+    // Any query params at / indicate a React app request (embedded map,
```
daniellekain: One product-level question about the "any query params → redirect to /map" rule: once the new homepage is live, marketing and campaign links will often carry tracking params (e.g. opensupplyhub.org/?utm_source=newsletter), and as written those will bounce to the map instead of showing the homepage. That's the right behavior for legacy links (the map was /), but I think it will surprise the comms team going forward. Two options I can see: strip a small denylist of tracking params (utm_*, gclid, fbclid) and serve Craft when nothing else remains, or just flag to comms that homepage links need to be param-free. Fine as a follow-up ticket — I mostly want us to make the call consciously before this is the public homepage. Let me know what you think!

deployment/environments/terraform-production.tfvars
daniellekain: Are we tracking the e2e test updates?

doc/release/RELEASE-NOTES.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
+* [OSDEV-2825](https://opensupplyhub.atlassian.net/browse/OSDEV-2825) - Moved the React map application from `/` to `/map`. A `mapRoute` constant was added alongside `mainRoute` in `src/react/src/util/constants.jsx`; `Routes.jsx` now serves the map at `/map` and renders a client-side `<Redirect>` from `/` to `/map` that preserves the incoming location `search` and `state`, so existing map deep links and query-string searches keep working. Every internal link that pointed at `/` now points at `mapRoute`. This redirect is what keeps `/` working in environments that do not run the CloudFront function, where the proxy is enabled CloudFront serves the Craft homepage at `/`, and where it is not React redirects `/` to `/map`.
```
daniellekain: Could we add a line to the Release instructions noting that deploying this release flips the public homepage on opensupplyhub.org (enable_homepage_proxy is now true in production)?

daniellekain: Tiny copy nit: this sentence is a little hard for me to follow: "This redirect is what keeps / working in environments that do not run the CloudFront function, where the proxy is enabled CloudFront serves the Craft homepage at /, and where it is not React redirects / to /map." 

Maybe something like: "Where the proxy is enabled, CloudFront serves the Craft homepage at /; where it is not (e.g. local development), the React app redirects / to /map."

doc/release/RELEASE-NOTES.md
```diff
@@ -29,12 +30,20 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
 * [OSDEV-3114](https://opensupplyhub.atlassian.net/browse/OSDEV-3114) - Fixed intermittent local stack startup failures on `docker compose up -d`. Kafka in ZooKeeper mode could crash with `NodeExistsException` when the broker container was recreated before ZooKeeper cleared the stale `/brokers/ids/1` znode — migrated local Kafka to single-node KRaft in `docker-compose.yml` and removed the `zookeeper` service. The React development server could also fail when its anonymous `node_modules` volume retained stale dependencies; replaced it with named `react_node_modules` so the local dependency cache has an explicit, inspectable lifecycle and can be reset deliberately. Local dev only; AWS uses MSK and static React builds.
 
+### What's new
+* [OSDEV-2825](https://opensupplyhub.atlassian.net/browse/OSDEV-2825) - opensupplyhub.org now opens on the new Open Supply Hub homepage instead of going straight into the map. The map has moved to opensupplyhub.org/map, and every link to it inside the app. Existing map links keep working: a request to `/` that carries search parameters redirects to `/map` with the parameters intact, and once inside the app `/` redirects to `/map`.
```
daniellekain: another copy nit: perhaps " and every link to it inside the app. " --> " and every link to it inside the app now points there"

src/react/src/components/Footer/Footer.jsx
```diff
@@ -9,12 +9,13 @@ import FooterNav from './FooterNav';
 import FooterShapes from './FooterShapes';
 import FooterShapesMobile from './FooterShapesMobile';
 import FooterText from './FooterText';
+import { mapRoute } from '../../util/constants.jsx';
```
daniellekain: optional nit: '../../util/constants.jsx' — every other import of this module drops the extension ('../../util/constants').

daniellekain (dismissed):
Looks great! I had a couple of minor comments and follow-ups, but none are blocking!

deployment/terraform/cloudfront-functions/homepage_uri_rewrite.js
valencia-max: Good flag! @daniellekain 
Actually I was thinking something like this. My first approach was to create a set of possible and valid query params -- the ones that we have for querys and searching, but I think this is kind of hardcored.
As we don't know what are the real fields that the URL could carry, I'd prefer to add here the valid query params that will redirect the user to `/map?queryparams`, such as `q`, `contributors`, `countries`, etc. What do you think?

deployment/terraform/variables.tf
valencia-max: Yes, we are pointing to servd directly.

deployment/terraform/cloudfront-functions/homepage_uri_rewrite.js
valencia-max: After some investigation with Claude, I found these are the possible query params from different sources:

So for the CloudFront function, the distinction is clean: the only params that should NOT trigger a redirect to /map are the well-known tracking params that have nothing to do with the React app:
utm_source, utm_medium, utm_campaign, utm_content, utm_term — standard UTM
gclid — Google Ads click ID
fbclid — Facebook click ID
msclkid — Microsoft Ads click ID
ttclid — TikTok click ID

What do you think adding these ones in the deny list ?

daniellekain: That works for me — I like it better than my denylist idea. The map/search params are a finite set we control, whereas tracking params are open-ended, so allowlisting is the sturdier direction. Two things to consider as you implement:

1. **Source the allowlist from the front end's real filter params.** The React app's URL param names are defined in the frontend constants — worth pulling the list from there (and leaving a comment in the CloudFront function pointing back to that file) so the next person who adds a filter param knows to update both places. That's the main risk with the allowlist: a future `/?newfilter=x` link would quietly land on the homepage instead of the map.

2. **Mixed URLs.** A campaign link could carry both, e.g. `/?utm_source=newsletter&countries=CN`. I'd suggest: if *any* allowlisted param is present, redirect to `/map` and pass the full querystring through (tracking params included, so analytics still sees them). Only serve the homepage when *none* of the map params are present.

Happy for the implementation to land in this PR or a fast-follow, whichever keeps #1082 moving.

daniellekain:
sounds reasonable to me!

deployment/terraform/cloudfront-functions/homepage_uri_rewrite.js
valencia-max: Addressed!

daniellekain (approved):
thank you for making all of these changes!

vlad-shapik (approved):
LGTM

PR #1083, opened 2026-06-22
[OSDEV-2786] Strip trade union-contributed fields from API and downloads

VadimKovalenkoSNF:
## Summary

Hide fields contributed by **Union**-type contributors (`Contributor.contrib_type = 'Union'`) from programmatic API access and bulk downloads, while keeping the facility — and its primary identity fields (name, address, country, location) — fully visible for manual search on the web client and on facility profiles.

This is the **field-level** approach (replacing the earlier whole-facility exclusion): we strip the *fields* a union contributed and "pretend the union never contributed," rather than hiding the whole record. The big win is count parity — no facility ever disappears from the API/downloads, so search counts and export counts stay consistent.

### What is stripped
- Union-contributed **extended fields** and **contributor-supplied partner fields** (both live in `extended_fields`, so one contributor-id filter covers them).
- **Sector** values contributed *solely* by a union (a value also supplied by a non-union contributor or a claim is kept).

### What is NOT touched
- Primary fields: name, address, country, location.
- **System/synthesized partner fields** (MIT living wage, wage indicator) — derived from facility data, not from `extended_fields`, so they're structurally unaffected.
- The contributor list itself (out of scope).
- The web-client search results counter (kept fast/complete).

### Where it applies
| Surface | Web client / browser | API token / non-web-client | Exemption group |
|---|---|---|---|
| `GET /api/facilities-downloads/` (paid export) | **strip** | strip | keep |
| `GET /api/facilities/` (`detail=true`) | keep (manual search stays complete) | strip | keep |
| Facility profile (retrieve) | keep | keep | keep |

Downloads strip for **every** caller because the export is the paid product being protected. The list/search endpoint uses the `Authorization: Token` + `X-OAR-Client-Key`/`Referer` discrimination so manual search on the FE is unchanged.

### Exemption group
New `can_get_union_linked_data` auth group (migration `0214`). Members keep full union data on both surfaces; managed per user in Django admin.

## Why this approach
The previous whole-record exclusion would make a single-contributor union facility visible on the site but missing from the API/downloads (mismatched counts, "ghost" facilities) and over-remove non-union data on shared facilities. Field-level stripping keeps every facility available and removes only the union's contributions. Filtering happens at **serialization time**, so it's dynamic and requires **no reindex**.

## Notes / scope
- Embed-mode downloads are contributor-scoped (iframe widgets, already filtered to the embed contributor) and are left unchanged.
- `Union` already existed as a contributor type; this PR adds the `UNION_CONTRIB_TYPE` constant and the request/permission helpers.

## Test plan
- [x] `api.tests.test_trade_union_field_exclusion` (23 tests): helper unit tests (extended fields, sector), download serializer stripping (extended/partner/sector), system-field independence, permission gate, and download-gate (browser strips, exempt keeps).
- [x] Existing suites green: `test_facility_download_serializer`, `test_facility_download_serializer_embed_mode`, `test_facility_index_details_serializer`, `test_facility_download`.
- [x] `flake8` clean; `manage.py check` clean; migration `0214` applies.
- [ ] Manual: token download of a union-linked facility drops union fields/sector; browser download (non-exempt) drops them too; exempt user keeps them; FE manual search/profile unchanged.

## Release notes
Updated under `Release 2.26.0` in `doc/release/RELEASE-NOTES.md` (migration + behavior). No reindex required.


src/django/api/permissions.py
```diff
@@ -52,6 +54,39 @@ def has_permission(self, request, view):
         return False
 
 
+def has_api_token(request):
```
VadimKovalenkoSNF, line 57: We detect API tokens by parsing the Authorization header (startswith('token ')) rather than checking `request.auth`, because this works on any request object without depending on DRF's auth pipeline having run — which keeps it usable on plain `HttpRequests` and trivially unit-testable with `RequestFactory`. 

It also matches on intent (a token was presented) rather than on successful authentication, so the union-data gate fails safe: anything that announces itself as a programmatic token call gets union data stripped even if the token is invalid or expired.

VadimKovalenkoSNF:
Closed in favor of https://github.com/opensupplyhub/open-supply-hub/pull/1087

PR #1084, opened 2026-06-22
[OSDEV-2881] Implemented selective OpenSearch index and template clearing for deployment workflows

vlad-shapik:
Implements [OSDEV-2881](https://opensupplyhub.atlassian.net/browse/OSDEV-2881)

## Summary

Replace the all-or-nothing `clear-opensearch` boolean with a `clear-opensearch-target` choice so deploys can clear `moderation-events`, `production-locations`, or both without forcing a full reindex when only one index mapping changed.

- Parameterize `clear_opensearch.sh.tpl` via `CLEAR_OPENSEARCH_TARGET` (index + template + EFS lock file per pipeline).
- Update `Deploy to AWS`, `[Release] Deploy`, and `DB - Apply Anonymized DB` workflows.
- **Deploy to AWS:** when **Restore database to original state** is enabled, always clear both indexes regardless of the selected target.
- **DB - Apply Anonymized DB:** always clears both indexes after restoring the anonymized dump (unchanged behavior; no selective option).
- Update `RELEASE-PROTOCOL.md` and 2.26.0 release notes.

Motivated by [OSDEV-2696](https://opensupplyhub.atlassian.net/browse/OSDEV-2696): a moderation-events mapping change only needs ~1 hour to refill, while clearing `production-locations` too would reindex ~2.5M records over 6+ hours. Cleared indexes serve reduced data on related v1 GET endpoints until Logstash finishes refilling.

## Test plan

- [ ] Run `Deploy to AWS` with `clear-opensearch-target: moderation-events` and confirm only moderation-events index, template, and lock file are removed.
- [ ] Run with `clear-opensearch-target: production-locations` and confirm production-locations artifacts are removed only.
- [ ] Run with `clear-opensearch-target: both` and confirm both pipelines are cleared.
- [ ] Run **Deploy to AWS** with **Restore database to original state** enabled and confirm both indexes are cleared regardless of target choice.
- [ ] Run **DB - Apply Anonymized DB** and confirm both indexes are cleared.
- [ ] Verify Logstash repopulates only the cleared index(es) after the workflow completes.

[OSDEV-2881]: https://opensupplyhub.atlassian.net/browse/OSDEV-2881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OSDEV-2696]: https://opensupplyhub.atlassian.net/browse/OSDEV-2696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


VadimKovalenkoSNF (dismissed):

VadimKovalenkoSNF (approved):

valencia-max (approved):

daniellekain:
@vlad-shapik really like the selective-clear idea here — wanted to flag one thing I think might trip it up, in case I'm reading the flow wrong.

From what I can tell, `clear_opensearch.sh` gets rendered with `envsubst` and then scp'd to the Bastion to run there — where `CLEAR_OPENSEARCH_TARGET` isn't set in the environment. So I think the value needs to be baked in at render time by `envsubst`. The template currently has:

```bash
target="${CLEAR_OPENSEARCH_TARGET:-none}"
```

and it looks like `envsubst` only substitutes the plain `$VAR` / `${VAR}` forms — it doesn't seem to touch the `${VAR:-default}` form. If that's right, the rendered script reaching the Bastion would still contain `${CLEAR_OPENSEARCH_TARGET:-none}`, evaluate it at runtime where the var is unset, fall back to `none`, and skip all the `if [ "$target" = … ]` branches — so nothing would get cleared regardless of the selected target.

Here's the quick check I ran:

```
$ printf 'target="${X:-none}"\n' | X=moderation-events envsubst '${X}'
target="${X:-none}"        # :-default form → not substituted
$ printf 'target="${X}"\n'      | X=moderation-events envsubst '${X}'
target="moderation-events" # plain form → substituted
```

If that holds, it'd affect all three paths (selective target, `restore-db`'s "always both", and the anonymized-DB "always both"), and it'd be a quiet one since CI still passes and the deploy reports success.

Possible fix would be to use the plain form so `envsubst` bakes the value in:

```diff
-target="${CLEAR_OPENSEARCH_TARGET:-none}"
+target="${CLEAR_OPENSEARCH_TARGET}"
```

The GH steps already always provide a value (input default `none`, or the `restore-db` if/else, or `both`), so the runtime fallback may not be needed. Does that match your understanding, or am I missing somewhere the target gets passed through to the Bastion?

PR #1086, opened 2026-06-25
[OSDEV-2732] Remove stale editable-fields list from system field error text

protsack-stephan:
## What

Remove the hardcoded list of editable settings ("Only label, unit, source by, base url, display text, active, and availability settings can be edited.") from the validation error raised when a moderator edits a protected attribute (Name, Type, JSON Schema, System Field) of a system-defined partner field in Django admin.

## Why

The hardcoded list duplicated the set of editable partner-field settings and would silently go stale whenever that set changed, showing users an inaccurate message. The error now states only that the protected field cannot be modified for system-defined fields and why, without enumerating editable fields.

## Notes

- Targeted for the next release (2.27.0); release notes updated accordingly.


vlad-shapik (dismissed):
LGTM

VadimKovalenkoSNF (dismissed):

VadimKovalenkoSNF (approved):

valencia-max (approved):

PR #1087, opened 2026-06-25
[OSDEV-2786] Anonymize trade union-linked contributors from paid products

VadimKovalenkoSNF:
## Summary

Implements [OSDEV-2786](https://opensupplyhub.atlassian.net/browse/OSDEV-2786)

Anonymise contributor-linked data in OS Hub's paid products (programmatic API and bulk download) while keeping it visible on the public web app and facility profiles, so contributor identities cannot be harvested at scale.

Anonymisation is opt-in, driven by a per-**contributor** admin checkbox, and applies to any contributor type — not only unions.

### Admin toggle + migration

Adds `anonymise_in_paid_products` (BooleanField, default `False`) to **`Contributor`** (not `User`), exposed in Django admin under the contributor record. Migration `0218`. Opting in is an explicit admin action; contributors are public by default. The flag also covers the `HistoricalContributor` table managed by `simple_history`.

Previously the field lived on `User` and was named `hide_in_paid_products` (default `True`, unions only). Moving it to `Contributor` makes the semantics correct — it's about the contributor's identity, not the admin account — and lets a single admin account manage multiple contributors with different anonymisation settings.

### `ShouldMaskContribution` service

Caches the set of contributors with the flag enabled in the shared `view_cache` (memcached) and matches a contribution by contributor `id`, `admin_id`, or name. When masked, the contributor is relabeled to `"Other"`, its id is dropped, and list metadata is stripped. The `created_from` attribution carries no id, so it is matched by name.

Two performance fixes were also made:
- `_load()` was iterating the `QuerySet` three times (separate passes for `contributor_ids`, `admin_ids`, and `names`), sending three identical `SELECT` statements per cache-miss. A single `list()` materialisation now covers all three.
- For token-authenticated list requests, `get_masked_contributors()` was called twice (once in `__is_anonymised_only`, once in the serializer). The view now resolves the set once in `list()` and passes it to both consumers via the serializer context.

### API (`GET /api/facilities/` list + detail)

- Masks anonymised contributions for programmatic (token-authenticated) requests; web app and facility profiles still show real names.
- Adds `anonymised_only: bool` to list responses. `true` only when every facility in the result is attributed exclusively to anonymised contributors. Drives the front-end download-button state without the frontend needing to inspect contributor types.

### Download (`GET /api/facilities-downloads/`)

Always masks anonymised contributors (web client and API alike), since a CSV/XLSX export is a paid product regardless of caller.

### Search-page download button (frontend)

Disabled with an explanatory tooltip only when `anonymised_only` is `true`. A mixed result or a **Show only shared facilities** search stays downloadable, with anonymised rows relabeled to `"Other"`. The previous union-type detection logic (inspecting `contributor_type`, `UNION_CONTRIBUTOR_TYPE`, `mapContributorTuplesToSelectOptions`) is replaced by this single API signal.

## Caching correctness

- **Vary the detail cache on `Authorization`.**  The detail endpoint (`cache_page`) was keyed by URL only, so a web response (names shown) and an API token response (names masked) collided on the same key. The cache now varies on `Authorization`.
- **Invalidate on flag toggle.** `Contributor.save()` detects when `anonymise_in_paid_products` flips and flushes `view_cache`, clearing both the masked-contributor set and cached detail responses everywhere immediately — no TTL lag.

## Test plan

- [x] `api.tests.test_should_mask_contribution` passes — covers service matching, caching, save-time invalidation, list/detail/download/`created_from` masking, `__is_anonymised_only` helper, download view serializer wiring, and `anonymised_only` key presence in the list response.
- [x] Token API request to a facility with an anonymised contributor returns `"Other"` (no id/list metadata); same request without a token returns the real name.
- [x] CSV/XLSX download relabels anonymised-contributor rows to `"Other"` for both web and API callers.
- [x] Toggling **Anonymise in paid products** in Django admin under a Contributor record takes effect on the next request without a manual cache clear.
- [x] Download button: disabled when `anonymised_only: true`; enabled for a mixed result and for **Show only shared facilities**.

[OSDEV-2786]: https://opensupplyhub.atlassian.net/browse/OSDEV-2786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/models/contributor/contributor.py
```diff
@@ -135,6 +137,17 @@ class Contributor(models.Model):
...
+    anonymise_in_paid_products = models.BooleanField(
+        'Anonymise contributor name in paid products',
+        default=False,
+        help_text=(
+            "When enabled, this contributor's name is anonymised in OS Hub "
+            'paid products - the bulk data download and the programmatic API '
+            '- so the data cannot be attributed to them at scale. Their '
+            'contributions stay visible on the public web app and facility '
+            'profiles.'
+        )
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: The finding is not valid — default=False is intentional design, not a regression.

Why it's correct as-is: The original hide_in_paid_products field on User had default=True, but it was scoped exclusively to Union contributors, so unions were effectively hidden by default without any admin action. When the field was moved to Contributor and broadened to cover any contributor type, changing the default to True would anonymise every contributor in the system on first deploy — including brands, retailers, and all non-union contributors that were never meant to be masked.

The deliberate design choice documented in the PR is opt-in: an admin must explicitly check the box for a contributor. default=False is the only safe default for a flag that now applies universally.

No change needed.

src/django/api/services/should_mask_contribution.py
```diff
@@ -0,0 +1,91 @@
...
+    def _load():
+        rows = list(
+            Contributor.objects
+            .filter(anonymise_in_paid_products=True)
+            .values_list('id', 'admin_id', 'name')
+        )
+        contributor_ids = {row[0] for row in rows}
+        admin_ids = {row[1] for row in rows if row[1] is not None}
+        names = {row[2] for row in rows if row[2]}
+        return contributor_ids, admin_ids, names
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Verdict: Skip — the fix requires a reindex and is out of scope.

Here's what the investigation found:

created_from_info is a denormalized JSON blob populated by the index_created_from_info SQL function (migration 0130). That function selects only contributor_name and contrib_type — no contributor id:

```
json_build_object(
    'should_display_associations', as1.is_active AND as1.is_public,
    'created_at',       afli.created_at,
    'contributor_name', ac.name,
    'contrib_type',     ac.contrib_type
)
```
To match by id instead of name you'd need to:

Add contributor_id to that SQL function (a new migration),
Trigger a full reindex of every FacilityIndex row so the new field propagates into the JSON blobs already stored in the DB, and
Update the serializer and MaskedContributors to use the id.
That's a significant infrastructure operation (potentially millions of rows) that goes well beyond this PR.

The practical risk is also very low:

`masks_name` is used only for `created_from_info`, which appears solely in the facility detail view (not the list).
Over-masking requires two distinct contributors to share the exact same name, one anonymised and one not — an unlikely coincidence.
names is populated from the `same _load()` query, so it already contains only the names of contributors with the flag enabled, not all contributor names.
No code change made.

src/django/api/models/contributor/contributor.py
```diff
@@ -144,6 +157,50 @@ class Contributor(models.Model):
...
+        Drop the caches that decide paid-product anonymisation so an admin
+        toggle of ``anonymise_in_paid_products`` takes effect right away.
+
+        The ``view_cache`` (memcached) holds both the anonymised-contributor
+        set and the per-Authorization facility detail responses, so a single
+        flush clears both for every worker. Without it the change would only
+        surface once they expire (up to
+        ``MEMCACHED_VIEW_CACHE_TIMEOUT_SECONDS``).
+        """
+        caches['view_cache'].clear()
```
protsack-stephan: Is this invalidating the entire cache? If so, it's not a good pattern; we don't want to invalidate other cached keys, as that will slow down the system in general. We need to invalidate the specific key here.

src/django/api/services/should_mask_contribution.py
```diff
@@ -0,0 +1,107 @@
...
+        is only touched for API callers.
+        """
+        if not cls.is_api_request(request):
+            return MaskedContributors()
+        return cls.get_masked_contributors()
+
+    @staticmethod
+    def is_masked(contributor, masked):
+        """Return True when this contributor's identity must be hidden."""
+        return bool(masked) and masked.matches(contributor)
```
protsack-stephan: What's the point of making this a set of static methods? Can't we just create a class that has a `should_mask` or `apply_mask` method instead of calling a set of different methods? If the behavior differs in different scenarios, then we can have two separate classes. Or, we could use a builder in front of the class that would allow us to adapt the functionality to different scenarios.

src/django/api/views/facility/facilities_view_set.py
```diff
@@ -280,6 +246,15 @@ def list(self, request):
...
 
+        # Resolve the masked-contributor set once per request and reuse it in
+        # both __is_anonymised_only (always needs the full set) and the
+        # serializer (empty for web clients, full for API callers), avoiding a
+        # second cache round-trip when serialising.
+        masked = ShouldMaskContribution.get_masked_contributors()
+        context['masked_contributors'] = (
+            ShouldMaskContribution.gate_for_request(request, masked)
+        )
+
```
protsack-stephan: This API endpoint is quite heavy on the database operations, we want to make sure that if we are querying this here that it gets served from cache as much as possible.

src/django/api/views/facility/facilities_view_set.py
```diff
@@ -300,12 +278,40 @@ def list(self, request):
...
+        ``masked`` is the pre-resolved ``MaskedContributors`` from the caller
+        (already fetched from the cache) so this method never hits the cache a
+        second time.
+        """
+        anonymised_ids = list(masked.contributor_ids)
+        if not anonymised_ids:
+            return False
+        if count is None:
+            count = queryset.count()
```
protsack-stephan: Are we doing an additional count here? This is a quite consuming operation, can we avoid doing so?

protsack-stephan (commented):
Good work! Couple of comments from me. Happy to discuss on the call if something is unclear.

src/django/api/views/facility/facilities_view_set.py
VadimKovalenkoSNF: This is already cache-first: `get_masked_contributors()` reads from the shared memcached (view_cache) and only hits the DB on a miss (~once/minute globally, TTL 60s), with immediate invalidation on `Contributor.save()` when the flag changes. It's resolved once per request and reused by both `__is_anonymised_only` and the serializer, so there's no extra DB or cache round-trip on the hot path.

src/django/api/services/should_mask_contribution.py
VadimKovalenkoSNF: Thanks for the comment. I refactored this part and applied separation to make this logic clearer:

1. `ContributorMaskingPolicy` — decides which contributors to hide for a given endpoint (`for_download` = always, `for_facilities_api` = only token callers) and hands back a ready `MaskedContributors`.
2. `MaskedContributors` — the value object that applies the mask per contributor (`should_mask` / `should_mask_name`).

src/django/api/models/contributor/contributor.py
VadimKovalenkoSNF: Good catch, fixed. It was calling `caches['view_cache'].clear()`, which flushed the whole shared cache (the `cache_page` facility-detail responses and everything else in `view_cache`), so every toggle degraded the system as you noted.

It now deletes only the specific key that holds the anonymised-contributor set.

One caveat worth flagging: the list API and bulk download read the masked set live off this single key, so they reflect the toggle immediately. 
The retrieve detail responses are cached by `cache_page` under keys derived from URL + Authorization, which memcached can't enumerate for targeted deletion, **so those self-expire on their own TTL rather than being force-invalidated**. Since toggling the flag is a rare admin action, that bounded staleness on individual facility pages seemed an acceptable trade-off versus flushing the entire cache.

src/django/api/views/facility/facilities_view_set.py
VadimKovalenkoSNF: Good point. I used a single cheap EXISTS first and bails out immediately in the common case (some facility is downloadable → not anonymised-only).

src/django/api/models/contributor/contributor.py
```diff
@@ -144,6 +158,52 @@ class Contributor(models.Model):
         excluded_fields=['uuid', 'origin_source', 'partner_fields']
     )
 
+    def save(self, *args, **kwargs):
```
VadimKovalenkoSNF, line 161: We added the `save()` override on Contributor model for one reason: cache invalidation of the masked-contributor set.

src/django/api/serializers/facility/facility_download_serializer_base.py
```diff
@@ -14,6 +15,14 @@ class FacilityDownloadSerializerBase(Serializer):
 
     row = SerializerMethodField()
 
+    def __init__(self, *args, masked_contributors=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.masked_contributors = (
```
VadimKovalenkoSNF, line 20: Capture the `masked_contributors` argument at construction time and stashes it on the instance (`self.masked_contributors`), so every row/field helper can read it without it being threaded through each call.

src/django/api/models/contributor/contributor.py
```diff
@@ -144,6 +158,52 @@ class Contributor(models.Model):
...
+            except Contributor.DoesNotExist:
+                should_invalidate = True
+
+        super().save(*args, **kwargs)
+
+        if should_invalidate:
+            self.__invalidate_paid_product_masking_cache()
+
+    @staticmethod
+    def __invalidate_paid_product_masking_cache():
```
VadimKovalenkoSNF: Invalidate cache on Contributor checkbox change.

src/django/api/models/contributor/contributor.py
```diff
@@ -144,6 +158,52 @@ class Contributor(models.Model):
...
+
+        Only the single ``MASKED_CONTRIBUTOR_IDS_CACHE_KEY`` entry is deleted -
+        not the whole ``view_cache`` - so unrelated cached responses keep their
+        warm entries. This set is the one shared across every worker and drives
+        the list API and bulk download immediately. The per-facility detail
+        responses cached by ``cache_page`` cannot be enumerated for targeted
+        deletion in memcached, so they fall back to their own short TTL
+        (``MEMCACHED_VIEW_CACHE_TIMEOUT_SECONDS``).
+        """
+        caches['view_cache'].delete(MASKED_CONTRIBUTOR_IDS_CACHE_KEY)
```
protsack-stephan: Does it makes sense to have this as a separate method with a single line of code?

protsack-stephan (dismissed):

src/django/api/models/contributor/contributor.py
VadimKovalenkoSNF: Yes, it is redundant complexity. I removed this wrapper method and explicitly clear the cache.

protsack-stephan (dismissed):

protsack-stephan (dismissed):

src/django/api/serializers/facility/facility_download_serializer_base.py
valencia-max: Good catch!!

src/django/api/serializers/facility/facility_index_details_serializer.py
```diff
@@ -437,7 +457,8 @@ def __serialize_and_sort_partner_fields(
         user_can_see_detail: bool,
         embed_mode_active: bool,
         use_main_created_at: bool,
-        date_field_to_sort: str
+        date_field_to_sort: str,
+        masked
```
valencia-max: nit: What data type should be this "masked" parameter?

src/django/api/serializers/facility/facility_index_serializer.py
```diff
@@ -40,6 +43,22 @@
 logger = logging.getLogger(__name__)
 
 
+def with_masked_contributors(get_field):
```
valencia-max, line 46: What is the advantage of making this decorator over fetching this data in line?

src/django/api/serializers/utils.py
```diff
@@ -53,20 +54,28 @@ def prefer_contributor_name(serializer):
         return False
 
 
-def get_contributor_name(contributor, user_can_see_detail):
+def _is_contribution_masked(contributor, masked):
+    return bool(masked) and masked.should_mask(contributor)
```
valencia-max, line 58: `utils.py` is currently using the same logic and they are doing the same. Do we want to have the same function in two different places?

src/django/api/serializers/facility/facility_index_serializer.py
valencia-max: Forget it, I just read the whole code.

valencia-max (dismissed):
Awesome!! LGTM

src/django/api/serializers/utils.py
VadimKovalenkoSNF: Fixed

src/django/api/serializers/facility/facility_index_details_serializer.py
VadimKovalenkoSNF: Added type here.

vlad-shapik (approved):
LGTM

protsack-stephan (approved):

PR #1092, opened 2026-06-29
[OSDEV-2679] Show latest edit date in claimed section

daniellekain:
## Summary

Fixes [OSDEV-2679](https://opensupplyhub.atlassian.net/browse/OSDEV-2679): the claimed **"Operational Details Submitted by Management"** section showed the *original claim date* instead of the *latest edit date*, even after a claimant updated their profile.

This is **Level 1** (claim-level date) from the [ticket analysis](https://opensupplyhub.atlassian.net/browse/OSDEV-2679?focusedCommentId=32372) — the whole section now shows one "last edited" date. (Per-field dates = Level 2, a separate larger effort.)

## Root cause

The profile page reads `claim_info` from a denormalized JSON blob on `FacilityIndex`. The blob never contained `updated_at` (only `created_at`), so the frontend could only ever show the original claim date.

**Correction from review (Stephan was right):** the original diagnosis also claimed claim edits never reindexed the blob. That was wrong — the existing DB trigger (`facility_claim_post_update_insert_indexing_trigger` → `perform_facility_claim_indexing`) refreshes `claim_info` on every `api_facilityclaim` UPDATE, covering all edit paths including claim reassignment during merges. Verified empirically (ORM-only save refreshes the blob; the integration test passes with no view-side reindex). The blob refreshed all along — it just never contained the timestamp.

No data loss — the claim's `updated_at` and full history were always saved; this is display/indexing only.

## Changes

- **`0218_add_updated_at_to_index_claim_info.py` + `0218_index_claim_info.sql`** — add `updated_at` to the `index_claim_info()` SQL function (follows the `0200`/`0203` pattern).
- **`claimDataSelectors.js`** — prefer `updated_at` for the displayed date (fallback `approved_at`/`created_at`).
- **RELEASE-NOTES.md** — 2.27.0 entries (migration, bugfix, reindex note).

## Deploy / reindex

Existing approved claims pick up `updated_at` only after their blob is recomputed. Handled by a **one-time targeted backfill**: a `claim_info` field group added to `backfill_facility_index` (reusing the OSDEV-2390 backfill infrastructure, per review), scoped to already-claimed rows (`claim_info IS NOT NULL`). Note `reindex_database` (a Postgres `REINDEX`) does **not** recompute `claim_info` — the earlier description was wrong about that. New edits refresh automatically via the existing DB trigger.

## Testing

- **Backend (13/13 pass):** added `test_updating_claim_profile_refreshes_indexed_claim_info` — PUTs a claim profile edit, then asserts the `FacilityIndex.claim_info` blob contains a non-null `updated_at` (proves the `0218` SQL function is live, since the test DB applies migrations) **and** the post-edit description (proves the new `index_facilities_new` call rebuilt the blob rather than serving it stale). Covers all three backend pieces in one test.
- **Frontend (19/19 pass):** added a `claimDataSelectors` test pinning the new precedence — `claimedAt` prefers `updated_at`, with `approved_at`/`created_at` as fallbacks.
- flake8 clean.

## Note on caching

The facility-details GET is view-cached (`MEMCACHED_VIEW_CACHE_TIMEOUT_SECONDS`, default 10 min), so after an edit the profile may display the previous date until the cache entry expires. Pre-existing behavior for all claim-edit data — not introduced or changed here.

[OSDEV-2679]: https://opensupplyhub.atlassian.net/browse/OSDEV-2679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


daniellekain:
### Verification note — SQL function tested directly against the DB

I couldn't run a full `migrate` locally (the dev image predates main's `dj-rest-auth`/allauth bump from OSDEV-2820, so Django won't boot until rebuilt). Since migration `0218`'s only action is running `0218_index_claim_info.sql`, I applied that SQL directly to local Postgres and called the function against a real approved claim:

```
CREATE FUNCTION                          -- applied cleanly, no syntax errors

 blob_created_at                  | blob_updated_at
----------------------------------+----------------------------------
 2025-07-05T06:49:29.58022+00:00  | 2025-07-24T15:06:38.982565+00:00
```

`index_claim_info()` now returns **both** `created_at` and `updated_at`, and they differ — confirming the function surfaces the claim's latest-edit date distinct from the original creation date. ✅

**Still wants a run on a rebuilt image before un-drafting:**
- [x] SQL function returns `updated_at` (verified directly)
- [x] `migrate` end-to-end records `0218`
- [x] Edit an approved claim → confirm reindex fires and the claimed section shows the new date
- [x] Frontend renders `updated_at`


edit: : migrate applied 0218 cleanly on the rebuilt dev image (showmigrations ✓, function present in pg_proc), and a live ORM edit of approved claim 1 showed the trigger refreshing the blob — updated_at moved to the edit time, created_at unchanged, description propagated. FE covered by the updated selector test in green CI.

src/django/api/views/facility/facility_claim_view_set.py
```diff
@@ -507,6 +507,11 @@ def get_claimed_details(self, request, pk=None):
                     )
                     claim.facility.save()
 
+            # Refresh the facility's denormalized index so the claimed
+            # section reflects this edit (e.g. the updated_at date) rather
+            # than serving the stale, pre-edit claim_info blob. See OSDEV-2679.
+            index_facilities_new([claim.facility_id])
```
protsack-stephan: If I remember correctly we did this using database triggers in the past. Can we follow the same approach here instead of calling the function in the view?

protsack-stephan (commented):
Can we maybe re-use the indexing infrastructure that Vlad has created here? Or is it not fitting the use case for this specific problem?

daniellekain:
> Can we maybe re-use the indexing infrastructure that Vlad has created here?

Good call — done in 09e477c7, with one caveat. There are two distinct reindex moments in this PR:

**1. One-time bulk refresh of existing claims (deploy) — now uses Vlad's `backfill_facility_index`.** Since `0218` changed `index_claim_info()`, every already-approved claim's stored `claim_info` blob needs recomputing once. I added a `claim_info` field group to `specs.py` and wired it into the existing 2.27.0 `post_deployment` backfill (`fields='contributors,claim_info'`). It's filtered to `claim_info IS NOT NULL`, so it only touches claimed rows (~3.6k locally, not the full table). This also fixes a real gap: the previous plan leaned on `reindex_database`, but that's a Postgres `REINDEX` (physical indexes) and never recomputes the denormalized `claim_info` data — so existing claims would never have picked up `updated_at`.

**2. Per-edit refresh of a single facility (request path) — kept as `index_facilities_new([claim.facility_id])`.** Vlad's backfill can't target one facility; its workers hash-partition the whole `api_facilityindex` table and each spawns a full Django subprocess. For a synchronous single-facility refresh, `index_facilities_new` → `index_facilities_by([id])` (one row) is the established idiom (same as `revoke_claim`, `facility_match_temp`, and the merge path). So that one isn't a fit for the backfill package.

valencia-max (dismissed):
LGTM!

VadimKovalenkoSNF (dismissed):
Approved. If you haven't already, could you also test this with the facility merging and promotion workflows to confirm that the fix covers all relevant use cases?

src/django/api/views/facility/facility_claim_view_set.py
daniellekain: You're right, and thanks for the pointer — this taught me something. The `facility_claim_post_update_insert_indexing_trigger` (from the 0170 triggers migration) already fires on every `api_facilityclaim` UPDATE and calls `perform_facility_claim_indexing`, which recomputes `claim_info` among the claim-derived columns. I verified empirically: an ORM-only `claim.save()` (no view involved) refreshes the blob, and the integration test passes with the view-side call removed — the test DB's trigger does the work.

So the view call was redundant and it's gone (b02c1e57). It also means my original "claim edits never reindexed" diagnosis was wrong — the blob refreshed all along; it just never *contained* `updated_at`. The real fix was always just the SQL function + the frontend preference. PR description and release notes corrected.

daniellekain:
@protsack-stephan re: reusing the indexing infrastructure — yes, adopted for the piece where it fits. The per-edit refresh doesn't need it (the DB trigger covers that, per your inline comment), but the **one-time backfill of existing blobs** is exactly its use case: added a `claim_info` field group to `backfill_facility_index` (09e477c7), scoped to already-claimed rows (`claim_info IS NOT NULL`), so the release backfills the claimed subset in one targeted pass instead of a full reindex. Also fixed a wrong claim in the release notes while in there: `reindex_database` is a Postgres `REINDEX` and does *not* recompute `claim_info`, so it was never going to cover this.

@VadimKovalenkoSNF re: merges and promotions — traced both. **Merges are covered:** the merge flow reassigns each claim with `claim.facility = target; claim.save()`, and each save is an `api_facilityclaim` UPDATE, so the DB trigger fires per claim with the survivor's facility id and refreshes the survivor's `claim_info` (including the status flips to REVOKED/DENIED when the target already has an approved claim — those are saves too). The losing facility's index row is deleted with the facility. **Promotion doesn't touch claim rows**, so there's nothing to refresh there — the claimed section is unaffected. The empirical trigger verification in the thread above covers the mechanism these paths ride on.

VadimKovalenkoSNF (approved):
Approved

protsack-stephan (approved):

PR #1093, opened 2026-06-29
[OSDEV-2920] Align moderation-event PATCH os_id with GET

daniellekain:
## Summary

Implements **Option 2** from [OSDEV-2920](https://opensupplyhub.atlassian.net/browse/OSDEV-2920) (align behavior): the `PATCH /v1/moderation-events/{moderation_id}/` response now matches the GET endpoints' `os_id` / `os_id_snapshot` shape.

## Problem

The GET endpoints (OpenSearch-backed) serve `os_id` as `COALESCE(NULLIF(os_id_snapshot, ''), os_id)` — the **snapshot takes precedence**, with the live `os` FK as the fallback — and expose `os_id_snapshot` as its own field. `PATCH` (Postgres via `ModerationEventUpdateSerializer`) returned only the live FK and omitted the snapshot — so the same resource had two different shapes, and the shared `moderation_event` docs schema couldn't be accurate for both.

## Change

`ModerationEventUpdateSerializer` now mirrors GET:
- `os_id` → serves `os_id_snapshot` when set (it wins even over a differing live FK — the merge case), falling back to the live `os` FK only when the snapshot is unset. Preserves the originally-approved OS ID after a facility delete/merge.
- adds `os_id_snapshot` as its own field (null when unset, matching GET's `NULLIF(os_id_snapshot, '')`).

## Null presentation (verified)

The *values* now align between GET and PATCH, but the null *presentation* differs by design: the v1 GET path strips null values from responses (`OpenSearchService.__remove_null_values`), so unset fields are **omitted** on GET but returned as explicit **`null`** on PATCH (standard DRF). Documented in the shared schema (api-docs#70).

## Testing

- `api.tests.test_moderation_events_update` — **9/9 pass locally**, including three added by this PR:
  - `os_id` falls back to the snapshot when the live FK is null;
  - the snapshot **wins over a differing live FK** (mirrors GET's coalesce order);
  - `os_id` / `os_id_snapshot` are explicit nulls when neither a facility nor a snapshot exists.
- flake8 clean.
- GET-side behavior cross-checked against the `sync_moderation_events.sql` coalesce and a local OpenSearch index spot-check (unset fields absent from GET responses). The moderation-events GET has no `_source` allowlist, so no response-mapping change is needed.

## Related PR — docs

Schema update (add `os_id_snapshot`, correct the `os_id` precedence description, note null-vs-omitted) is in **opensupplyhub/open-supply-hub-api-docs#70**. With Option 2 the single shared `moderation_event` schema is valid for both GET and PATCH (no schema split needed). **That docs PR should merge _after_ this one**, since it documents behavior this PR introduces on PATCH.

## Note

Behavior change to a public API response (additive: adds a field, and `os_id` now serves the snapshot rather than going null post-deletion). Right after a PATCH the facility still exists, so `os_id == snapshot` — the divergence only matters post-deletion/merge. Consistency, not correctness.

[OSDEV-2920]: https://opensupplyhub.atlassian.net/browse/OSDEV-2920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


valencia-max (approved):
Cool! 🤖 🚀

protsack-stephan (approved):

VadimKovalenkoSNF (approved):

PR #1094, opened 2026-06-30
[OSDEV-1149] Add is_closed to production-locations endpoint

daniellekain:
## Summary

Adds an **`is_closed`** boolean to `GET /api/v1/production-locations/` (and the underlying OpenSearch index), per [OSDEV-1149](https://opensupplyhub.atlassian.net/browse/OSDEV-1149) AC#1. Part of the OpenSearch transition epic (OSDEV-1575).

## Why it was needed

The v1 response is a pass-through of the OpenSearch `_source`. The `production_locations` index already had closure **timestamps** (`opened_at` / `closed_at`) but **no `is_closed` boolean** — and closed-ness can't be reliably derived from the dates (a location can be closed then reopened). The authoritative value already exists as `Facility.is_closed` (already exposed by the legacy `/api/facilities/`), so this is a projection, not new closure logic.

## Changes

- **`sync_production_locations.sql`** — add `af.is_closed` (`af` = `api_facility`) to the Logstash sync query, placed with the canonical facility columns (plain scalar pass-through, no ruby transform).
- **`production_locations.json`** — map `is_closed` as `boolean`.
- **`production_locations_response.py`** — add `is_closed` to the `ProductionLocationsResponseMapping` base list. The v1 endpoints filter the OpenSearch `_source` through this allowlist, so indexing alone isn't enough — a field missing here is silently dropped from every response.
- **`test_production_locations_get.py`** — `test_is_closed_included_in_response_mappings` pins the field into both mapping lists (the test that would have caught the allowlist gap).
- **RELEASE-NOTES.md** — 2.27.0 Code/API entry + reindex note.

**No Django migration** — production-locations is fed by a raw Logstash JDBC query, not a Postgres `index_*()` function.

## Null semantics

`Facility.is_closed` is nullable, and the v1 API strips null values from responses (`OpenSearchService.__remove_null_values`). So locations whose closure status was never recorded **omit** `is_closed` rather than returning `null` — consumers should treat a missing key as "closure status not recorded."

## Testing

- **Unit:** `api.tests.test_production_locations_get` (11 tests incl. the new mapping test) and `test_production_locations_query_builder` pass; flake8 clean.
- **Verified end-to-end locally:** rebuilt Logstash with the new SQL, re-synced, and confirmed:
  - the OpenSearch doc carries `is_closed: true` for a closed facility, with the mapping typed `boolean`;
  - `GET /api/v1/production-locations/` (search) returns `is_closed: true` — AC#1;
  - `GET /api/v1/production-locations/{os_id}/` returns `is_closed: true`;
  - a facility with no recorded closure status omits the field.

## Deploy

**Release step required: a full production-locations OpenSearch reindex (~4 hours).** Existing locations pick up `is_closed` only after the index is rebuilt (clear + full Logstash re-sync) — the JDBC input tracks `updated_at`, so adding a column doesn't refresh old rows by itself. Schedule the rebuild within the release window and verify afterwards that `is_closed` appears on a known location. New/updated locations get the field on the next incremental sync regardless. Called out in RELEASE-NOTES.md release instructions. No Logstash pipeline `.conf` change needed — its `remove_field` list only strips `*_value` intermediates.

## Follow-ups (not in this PR)

- **Docs ([OSDEV-2951](https://opensupplyhub.atlassian.net/browse/OSDEV-2951)):** [api-docs#71](https://github.com/opensupplyhub/open-supply-hub-api-docs/pull/71) documents `is_closed` in the response schema (merge after this PR); the RFC schema page entry is drafted and will be folded in after merge (AC#2).
- **FE:** show "Closed" in search results. No dedicated ticket exists yet — this lands with the search-page integration of the v1 endpoint ([OSDEV-1202](https://opensupplyhub.atlassian.net/browse/OSDEV-1202) non-embed / [OSDEV-1597](https://opensupplyhub.atlassian.net/browse/OSDEV-1597) embed), which is where the search page starts consuming this response.

[OSDEV-1149]: https://opensupplyhub.atlassian.net/browse/OSDEV-1149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OSDEV-2951]: https://opensupplyhub.atlassian.net/browse/OSDEV-2951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


protsack-stephan (dismissed):
This will require re-indexing of open search during the release, that takes around 4 hours. And we need to make sure we do that during the release process, just FYI.

daniellekain:
@protsack-stephan good call — captured this as an explicit release step in `RELEASE-NOTES.md` (release instructions) and in the PR's Deploy section: full index clear + re-sync, **~4h**, scheduled within the release window, with a post-rebuild verification check (`is_closed` present on a known location). Also merged main and resolved the release-notes conflicts.

valencia-max (approved):
Amazing! 🚀

VadimKovalenkoSNF (approved):
Looks clear, approved.

protsack-stephan (approved):

PR #1099, opened 2026-07-02
[OSDEV-2953] Fix flaky test_geocoding_no_results blocking all PR checks

daniellekain:
## What

Mock the Google Geocoding API response in `FacilitySubmitTest.test_geocoding_no_results` using the existing `geocoding_no_results` fixture from `test_data.py`, so the test no longer makes a live geocoding call.

## Why

The Django unittest job is currently failing on `main` and blocking the checks on every open PR (e.g. #1093, #1094 — see [this run](https://github.com/opensupplyhub/open-supply-hub/actions/runs/28564390000), where both the feature-branch and base-branch Django jobs failed with the same error):

```
FAIL: test_geocoding_no_results (api.tests.test_facility_submit.FacilitySubmitTest.test_geocoding_no_results)
AssertionError: 'ERROR_MATCHING' != 'GEOCODED_NO_RESULTS'
```

The test submitted a real address to the live Google Geocoding API and asserted that Google returns `ZERO_RESULTS` for it. Google's geocoder now resolves that address, so the list item is geocoded successfully, proceeds to matching, and ends in `ERROR_MATCHING` in the test environment. Mocking `api.geocoding.requests.get` makes the test deterministic and independent of external geocoder behavior — it also drops its runtime from ~26s to ~1.5s since no live HTTP request (with retries) is made.

Verified locally: the test passes with the mock in place.

## Notes

Other tests in this suite (e.g. `test_exact_matches_with_create_false`, ~51s in CI) also appear to make live geocoding calls and could break the same way in the future; mocking those is left as a follow-up.


valencia-max (approved):
Awesome work! Can you please create a ticket to track this task? And I think if you are only fixing a flaky test, it doesn't need to be tracked in release notes... But happy to hear your thoughts!

protsack-stephan (approved):

VadimKovalenkoSNF (approved):

PR #1106, opened 2026-07-06
[OSDEV-2964] Prioritize claimant data as the highlighted source on the production location page, served from the v1 API

jamesjohnson4395:
## Problem

The new claims form no longer collects facility name and address — claimants are now expected to submit that data through SLC or list contributions from their own contributor account, while still editing optional fields (sector, number of workers, etc.) through the claim forms. But the promotion logic that picks the highlighted data source only recognized **FacilityClaim fields** as claimant data: a claimant's own SLC/list contributions received no special treatment and competed on recency with every other contributor. In practice this meant a claimed facility could submit its correct name via SLC and never see it highlighted on its own page.

Separately, the page sourced its highlighted values from the legacy `/api/facilities/{os_id}` endpoint, which the team wants to move away from.

This PR is a first attempt at solving these problems, fully using the new Claude Fable model, [as suggested by Bus](https://opensupplyhub.slack.com/archives/C03T93DQKDZ/p1783344496374179?thread_ts=1782897345.395539&cid=C03T93DQKDZ).

## What this does

**New per-field priority order** (computed in the OpenSearch sync, served by `GET /api/v1/production-locations/{os_id}`):

1. **Claimant data** — approved-claim fields *or* any contribution from the claimant's contributor account (SLC/list items); most recent wins within the tier
2. Moderator-promoted data (`created_from` + `promote_match`)
3. Data from a verified contributor
4. Most recent data
5. Canonical facility value (name/address only)

Recency is compared by `created_at` (original contribution date) rather than `updated_at`, because claim extended fields are re-saved wholesale on every claim edit — using `updated_at` would let a stale old-claims-form name outrank a newer claimant SLC after any unrelated dashboard save.

### By layer

- **Logstash** (`sync_production_locations.sql`): each promoted field (name, address, sector, parent_company, product_type, location_type, processing_type, number_of_workers, local_name, rba/duns/lei IDs) is now selected by a single ranked `LATERAL` subquery instead of nested `COALESCE`s, returning the value together with a new `*_source` attribution object (`contributor_id`, `contributor_name`, `contributed_at`, `is_claimed_data`, `is_promoted`, `is_verified`). Also fixes a data-quality gap: list-item candidates now require an active match (`AUTOMATIC`/`CONFIRMED`/`MERGED`, `is_active`) from an active, public source — the old "most recent" tier included deactivated and private contributions. Pipeline conf and index template updated for the new fields.
- **Django**: the twelve `*_source` objects are added to `PRODUCTION_LOCATION_BY_OS_ID` so single-location v1 lookups return them (search/list responses unchanged).
- **React**: the production location page now fetches the v1 document alongside the legacy facilities API response. Highlighted values, attribution (contributor, date), and the Claimed/Crowdsourced chip come from the v1 fields; the contribution drawers keep listing the other data sources from the legacy `extended_fields`, excluding the highlighted entry. If a document lacks the new attribution (not yet re-synced), the field falls back to the previous legacy-only behavior, and a stale-document guard prevents mismatched data while navigating between locations. The page title is intentionally unchanged.

This is the hybrid architecture described in the promotion-logic Confluence doc (highlighted source from v1/OpenSearch, other sources from the facilities API) — a deliberate first step off the legacy endpoint rather than a full migration, since the OpenSearch document doesn't yet hold per-field contribution lists, claim info, or activity reports.

## Shortcomings / follow-ups

- OpenSearch syncs on a schedule, so the highlighted value can briefly lag the drawer after a new contribution or claim approval; the frontend fallback covers unsynced documents.
- `isic_4` and `parent_company_os_id` aren't in the v1 document yet and keep legacy highlighting.
- The claimed-details PUT endpoint still accepts `facility_name_english`/`facility_address` (deliberately out of scope).
- The Confluence promotion-logic page needs updating after this ships.

## Testing

- New SQL validated against a local database, including synthetic claim/SLC scenarios: claimant SLC outranks other contributors; a newer claimant SLC name beats a stale old-claims-form name; within-claimant recency picks the newer of claim vs SLC values.
- Full local Logstash resync verified: all documents indexed with `*_source` attribution and returned by the live v1 endpoint.
- 104 React tests and 39 Django tests pass, including new suites for the v1 helpers, field configs, map address, and response mapping.

## Deployment

The `production-locations` index template changed and documents must be regenerated: run `[Release] Deploy` with the OpenSearch clear option set to `production-locations` (~6h refill; the page degrades gracefully to legacy behavior per location until its document is re-synced). See the 2.28.0 release-notes entry.


doc/release/RELEASE-NOTES.md
```diff
@@ -3,6 +3,33 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html). The format is based on the `RELEASE-NOTES-TEMPLATE.md` file.
 
+## Release 2.28.0
```
daniellekain, line 6: note: we pushed release 2.27 back to next week (July 17), so this can all go in release 2.27 (and the date of the release can shift back). I updated the release date in another PR of mine but not sure which will ship first

src/logstash/sql/sync_production_locations.sql
```diff
@@ -671,8 +237,903 @@ FROM
...
+      END,
+      c.contributed_at DESC
+    LIMIT
+      1
+  ) sector_ranked ON TRUE
+  LEFT JOIN LATERAL (
+    SELECT
+      c.value::TEXT AS value_json,
+      json_build_object(
+        'contributor_id', c.contributor_admin_id,
```
daniellekain, line 491: in the *_source objects, the field named contributor_id is actually filled with ac.admin_id — the contributor's user account id, not the contributor id.

Everywhere else in the codebase these are kept strictly apart: the moderation-events v1 index and the indexed contributor objects both use contributor_id = api_contributor.id, and the user-account id is always called user_id — and it's permission-gated, never served on a public endpoint. As written this both mislabels the value (consumers will build links/lookups on the wrong id) and publishes an id we deliberately gate elsewhere.

In each of the 12 LATERAL blocks we need to change change ac.admin_id AS contributor_admin_id → ac.id AS contributor_id (and the json_build_object key stays contributor_id)

src/logstash/sql/sync_production_locations.sql
```diff
@@ -1,41 +1,23 @@
...
+-- with its attribution (emitted as a *_source JSON column). Tier order:
+--   1. claimant data: values from the approved claim OR contributed by the
+--      claimant's contributor account (SLC/list items); most recent wins
+--   2. moderator-promoted data (created_from item with a promote_match action)
+--   3. data from a verified contributor
+--   4. most recent data
+--   5. canonical facility value (name/address only)
+-- Candidates are compared by created_at (original contribution date) because
+-- claim extended fields are re-saved wholesale on every claim edit, which
+-- makes updated_at unreliable for per-field recency.
```
daniellekain, line 12: I think one edge-case that this creates is that claim edits never win recency. 

ex/
2024: facility gets claimed, claimant enters number_of_workers = 100 → claim extended field row created, created_at = 2024
2025: claimant submits an SLC with number_of_workers = 150 → new row, created_at = 2025
2026: claimant corrects their dashboard to number_of_workers = 200 → the 2024 row is updated in place, created_at still 2024

Both rows are tier 1 (claimant data), so recency decides — and 2025 beats 2024. The page shows 150, and no amount of dashboard editing can ever fix it. Since the claims  dashboard is still the main editing surface for exactly these optional fields, this will bite real claimants.

I think we can either rank by most recent of created_at or updated_by; or note and document this as a trade-off

src/logstash/sql/sync_production_locations.sql
```diff
@@ -1,41 +1,23 @@
+-- Each promoted field is selected by a LATERAL subquery that ranks all
+-- eligible candidates in a single pass and returns the winning value together
+-- with its attribution (emitted as a *_source JSON column). Tier order:
+--   1. claimant data: values from the approved claim OR contributed by the
+--      claimant's contributor account (SLC/list items); most recent wins
+--   2. moderator-promoted data (created_from item with a promote_match action)
+--   3. data from a verified contributor
```
daniellekain, line 7: quick question about verified contributors--

I think the promotion-logic spike concluded the verified toggle is outdated.

now, any contributor flagged verified years ago now beats every unverified contribution forever, regardless of recency.

I just wanted to double check that this was intentional, and was wondering if we plan to start to set is_verified again?

src/logstash/sql/sync_production_locations.sql
```diff
@@ -1,41 +1,23 @@
...
+--   4. most recent data
+--   5. canonical facility value (name/address only)
+-- Candidates are compared by created_at (original contribution date) because
+-- claim extended fields are re-saved wholesale on every claim edit, which
+-- makes updated_at unreliable for per-field recency.
+-- List-item candidates are only eligible while they have an active match
```
daniellekain, line 13: With the new eligibility rules (active match, active + public source) are a genuine data-quality fix — but they mean some facilities' displayed name/address/etc. will be different after the reindex, because values that used to win under "most recent, no questions asked" are no longer eligible. 

Before the release, we may want to run the new SQL against a copy of prod and diff the promoted values against the current index for a sample (or all) facilities — so we know how many locations change and can spot-check that the changes look like corrections, not regressions. It also gives support a heads-up in case contributors ask why their facility's displayed name changed. Would be great to attach the count + a few examples to the PR/ticket.

src/react/src/components/ProductionLocation/ProductionLocationDetailsContainer/ProductionLocationDetailsContainer.jsx
```diff
@@ -62,6 +68,10 @@ function ProductionLocationDetailsContainer({
     // eslint-disable-next-line react-hooks/exhaustive-deps
     useEffect(() => {
         fetchFacility(normalizedOsID, contributors);
+        // The v1 document provides the highlighted (promoted) value and its
+        // attribution for each General Information field; the legacy fetch
+        // above still provides the full contribution lists for the drawers.
+        fetchProductionLocation(normalizedOsID);
```
daniellekain, line 74: Navigation race worth guarding: go from location A to location B — nothing cancels A's in-flight v1 request, and `completeFetchSingleProductionLocation` stores whatever response lands last. If A's response is slower than B's entire fetch, A's doc ends up in the store on B's page; the os_id guard then (correctly) refuses it, so B silently drops to legacy-only attribution until a refresh, with no visible error.

**Fix:** cancel the in-flight request when the osID changes (AbortController in the effect cleanup), or key the completion to the *requested* osID and drop stale ones. Note: compare against the requested id, not the response's `os_id` — the response carries the canonical id, so an alias-visited page would wrongly reject a correct response. And worth one test for the mismatch path — the new tests all inject matching docs, so the guard never actually fires in CI.

src/react/src/components/ProductionLocation/ProductionLocationDetailsGeneralFields/utils.js
```diff
@@ -66,6 +75,43 @@ const getOrderedFieldConfigs = includeAdditionalIdentifiers => {
         key: FIELD_CONFIG.name.key,
         getDataPointProps: data => {
             if (!data) return null;
+            const v1Highlighted = getV1HighlightedField(v1Data, 'name');
+            if (v1Highlighted) {
+                const highlightedComparable = normalizeDisplayValue(
```
daniellekain, line 80: The drawers exclude the highlighted entry by comparing normalized values, which misfires two ways:

1. If *two different contributors* submitted the same value, both get excluded — the second contributor's entry disappears from "other sources" entirely (and the match is case-insensitive, so `FACILITY A` from another org vanishes when `Facility A` is highlighted).
2. For `parent_company`, the legacy comparable is built differently (`contributor_name || name || raw_value`), so the exclusion can *fail* to match and the highlighted entry shows up twice.

**Fix:** the `*_source` object already carries `contributor_id` — exclude by contributor + value instead of value alone. (Ties into the earlier comment about which id that field holds.)

src/react/src/components/ProductionLocation/ProductionLocationDetailsGeneralFields/ProductionLocationDetailsGeneralFields.jsx
```diff
@@ -47,9 +48,16 @@ const ProductionLocationDetailsGeneralFields = ({
     const showAdditionalIdentifiers =
         !featureFlagsFetching &&
         activeFeatureFlags.includes(SHOW_ADDITIONAL_IDENTIFIERS);
+    // Guard against a stale v1 document while navigating between locations:
+    // the highlighted values must describe the same location as the legacy
+    // facilities-API payload that fills the contribution drawers.
+    const v1DataForLocation = useMemo(() => {
```
daniellekain, line 54: No loading gate on the v1 doc — the page renders legacy attribution first, then values/chip/contributor visibly swap when this lands. Worth a `fetching` check (or holding the chip until settled) to stop the flicker.

src/react/src/components/ProductionLocation/ProductionLocationDetailsMap/ProductionLocationDetailsMap.jsx
```diff
@@ -60,12 +60,20 @@ import getSelectedDrawerField from '../utils';
...
     singleFacilityData,
+    v1Data,
     gridColorRamp,
     history: { push },
     match: { params: { osID } = {} } = {},
 }) => {
     const mapRef = useRef(null);
 
+    // Guard against a stale v1 document while navigating between locations.
+    const v1DataForLocation =
```
daniellekain, line 71: This is the same os_id guard as GeneralFields (lines 54–57) copy-pasted — worth extracting to a shared helper so the two can't drift, and giving the mismatch path a test (the new tests only inject matching docs, so the guard is never exercised).

src/react/src/components/ProductionLocation/ProductionLocationDetailsGeneralFields/utils.js
```diff
@@ -66,6 +75,43 @@ const getOrderedFieldConfigs = includeAdditionalIdentifiers => {
...
+                    statusLabel: getStatusLabel(v1Highlighted.isClaimedData),
+                    contributorName: v1Highlighted.sourceName,
+                    userId: promotedContribution.userId,
+                    date: v1Highlighted.date,
+                    drawerData: { promotedContribution, contributions },
+                };
+            }
             const coreName = get(data, 'properties.name', '');
             const createdFrom = formatAttribution(
                 get(data, 'properties.created_from.created_at', ''),
```
daniellekain, line 117: In v1 mode this synthetic `created_from` entry never makes it into the drawer, so the original creator's contribution can vanish from "other sources" when the core name isn't also in `extended_fields.name`. Probably worth keeping the legacy entry in the v1 branch too.

src/react/src/components/ProductionLocation/ProductionLocationDetailsContainer/ProductionLocationDetailsContainer.jsx
```diff
@@ -25,6 +25,10 @@ import {
     fetchFacilities,
     resetSingleFacility,
 } from '../../../actions/facilities';
+import {
+    fetchProductionLocationByOsId,
```
daniellekain, line 29: The details page now shares `singleProductionLocation` state with the contribute/claim flows (SearchByOsIdResult, ProductionLocationInfo, ClaimForm all read it). Works today because each fetch resets the slot, but it's fragile coupling — a dedicated slice (or storing `{osId, doc}`) is cheap insurance against two consumers fighting over one slot.

src/react/src/components/ProductionLocation/ProductionLocationDetailsGeneralFields/utils.js
```diff
@@ -54,7 +56,14 @@ const formatActivityReports = data => {
     return items;
 };
 
-const getOrderedFieldConfigs = includeAdditionalIdentifiers => {
+const toV1PromotedContribution = (highlighted, displayValue) => ({
```
daniellekain, line 59: `is_promoted` / `is_verified` come back in every `*_source` object but nothing on the FE reads them. Fine if they're for a future chip — drop them from the response if not, so the payload stays honest.

daniellekain (commented):
Really strong PR — the ranked LATERAL design (value + attribution decided together, so they can never disagree), the honest trade-offs section, and dodging the v1 `_source` allowlist trap (with a pinning test!) all stand out.

One coordination heads-up: **this will conflict with #1094**, which edits the same region of `sync_production_locations.sql` (it adds an `af.is_closed` select right where the old `AS name,` block was rewritten), plus the same response-mapping, index-template, and test files. #1094 is small and slated for 2.27, so easiest is to let it merge first and rebase this on top — just make sure the `af.is_closed` line survives into the rewritten SELECT. Also note both PRs need a full production-locations reindex (2.27 ~4h, this one ~6h) — if 2.27 slips, worth folding them into a single rebuild.

ColumbusLeonardOSHub:
Can we please add a screenshot and/or recording to this? It'd be helpful to put it in context as I'm picking up to review.

daniellekain:
Heads-up: #1094 just merged, so this branch now conflicts as predicted — main's `sync_production_locations.sql` has `af.is_closed` in the region the ranked-subquery rewrite replaced, plus small overlaps in the response mapping, index template, and `test_production_locations_get.py`. When you rebase, make sure `af.is_closed` survives into the rewritten SELECT (it's a plain scalar pass-through, no ruby filter) and `"is_closed"` stays in `PRODUCTION_LOCATIONS`. Also: since 2.27 now runs a full production-locations reindex on July 17, if this PR lands in 2.28 it'll need its own — the shared-rebuild option from the review is worth keeping in mind.

PR #1117, opened 2026-07-10
[OSDEV-3017] URL-encode download-log path so multi-filter searches are recorded

daniellekain:
## What

URL-encode the search path in `makeLogDownloadUrl` so download logging records the **full** search query instead of just its first filter.

## Why

`makeLogDownloadUrl` (`src/react/src/util/util.js`) interpolated the raw browser path — `window.location.pathname + search + hash`, which contains its own `&`/`?`/`=` — directly into the log-download request URL:

```
/api/log-download/?path=${path}&record_count=${recordCount}
```

So when Django read `request.query_params.get('path')` (`api/views/log_download.py`), it had already split the query string on the first `&` and stored only the substring before it. Every multi-filter download was silently truncated to its first filter.

Confirmed against a local anonymized prod dump: **66,101** download records, **0** containing `combine_contributors`, and **at most one** `&` in any stored path (99.8% have none). This made the "Show only shared facilities" (`combine_contributors`) feature — and any multi-filter download — impossible to measure, which is what surfaced it in **OSDEV-2891**.

The bug has existed since the code was imported from open-apparel-registry, so it is long-standing rather than a recent regression.

## Change

- `encodeURIComponent(path)` so the whole path is one opaque query-string value that round-trips intact.
- Added frontend unit tests for `makeLogDownloadUrl` in `src/react/src/__tests__/utils.tests.js`, including a many-`&` case asserting the full path (with `combine_contributors=AND`) survives.
- No backend change — the server already stored whatever `path` value it received.

## Scope / caveats

- Fixes data captured **from deploy forward only**; the ~66k existing rows remain truncated and are not recoverable.
- No downstream consumer parses `DownloadLog.path` assuming the truncated shape (checked).

Refs: OSDEV-3017 · surfaced by OSDEV-2891


valencia-max (approved):
Good catch!

valencia-max:
Just a question: Do we want to fix this in the backend? Is there a case when a user tries to download data without using the OSHub UI?

daniellekain:
> Just a question: Do we want to fix this in the backend? Is there a case when a user tries to download data without using the OSHub UI?

that's a really good point!

I dug into this a bit, and it looks like this table is only tracking UI downloads (it tracks via /api/log-download/), so we're good with regards to this bug! 

I don't know if it was an intentional decision to only track UI downloads, or if it was an oversight. API users hit GET /api/facilities-downloads/  directly (instead of via api/log-download), so they bypass this logging. 

I think logging server-side at FacilitiesDownloadService.log_request() would be the more complete answer since every download flows through it, but that's a bigger change than this fix. Could be a good candidate for the backlog -- what do you think?

VadimKovalenkoSNF (approved):
Thank you for noticing this bug, approved!

protsack-stephan (approved):

PR #1118, opened 2026-07-10
[OSDEV-2949] Enforce partner field JSON Schema as API output whitelist

VadimKovalenkoSNF:
Fix [OSDEV-2949](https://opensupplyhub.atlassian.net/browse/OSDEV-2949)

## What

Enforce a partner field's JSON Schema as an output whitelist on the two JSON API endpoints, so contributor-supplied keys that the schema does not declare are never returned.

## Why

A partner uploaded `rsc_grievance_mechanism` with an extra `internal_ID` key (used to map their IDs to OS IDs during list contribution) and explicitly asked that it not be shared externally. It is currently visible in:
- `GET /api/facilities/{os_id}/`
- `GET /api/v1/production-locations/{os_id}`

Root cause is that the schema was never enforced on output. Contribution-time validation allows extra keys (no schema sets `additionalProperties: false`, which is the JSON Schema default), storage keeps the whole nested object under `value.raw_values` verbatim, and both endpoints emit it unchanged. The CSV/Excel download and the front end were already safe because they render only schema-declared fields — the JSON API was the one boundary that didn't apply that rule.

Rather than add an ingestion-time stripping service (destructive, needs a backfill, and ambiguous when a client actually wants an extra field), this applies the rule the rest of the system already follows — the schema defines what is public — at the serialization boundary. This fixes existing and future records at read time, is field-agnostic, and the intentional way to expose a field stays "add it to the schema in admin".

## Changes

- **`partner_field_helper.filter_value_to_schema()`** — recursively keeps only keys declared in the schema's `properties`. Handles nested objects, array `items`, and (defensively) list values under a `type`-less object schema. Returns the value unchanged when there is no usable object schema, so fields without a schema are never stripped.
- **`FacilityIndexExtendedFieldListSerializer`** and **`ProductionLocationsViewSet`** — apply the filter to `raw_values`. Redundant `deepcopy` calls were replaced with shallow copies since the helper already returns fresh structures (a small perf win on the v1 endpoint).

## Coverage note

Object fields with a `{type: object}` schema cannot store list `raw_values` (contribution-time validation rejects a list against `type: object`), so the list path mainly guards `type`-less schemas and data that bypassed validation.

## Test plan

- [x] Unit tests for `filter_value_to_schema` (incl. the exact RSC grievance schema + `internal_ID`, nested objects, list branches).
- [x] Integration test: `internal_ID` stripped from `GET /api/v1/production-locations/{os_id}`.
- [x] Integration test: undeclared keys stripped from `GET /api/facilities/{os_id}/` partner fields.
- [x] `flake8` clean on all changed files.

[OSDEV-2949]: https://opensupplyhub.atlassian.net/browse/OSDEV-2949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


valencia-max (approved):
LGTM!

VadimKovalenkoSNF:
Temporary set this to Draft in favor of https://github.com/opensupplyhub/open-supply-hub/pull/1131

VadimKovalenkoSNF:
Closed. Appropriate solution is here: https://github.com/opensupplyhub/open-supply-hub/pull/1131

PR #1120, opened 2026-07-13
[OSDEV-2928] Provision ContriBot infrastructure in Terraform

protsack-stephan:
## Summary

Provision ContriBot AWS infrastructure in Terraform and add the facility list validation library that future Lambda handlers will use.

**Infrastructure (Terraform)**
- Create four empty Secrets Manager stores for runtime credentials (OS Hub API token, Monday, Slack, Google Drive).
- Add an on-demand DynamoDB state table keyed by `list_id` to track processed facility lists across scheduled runs.
- Deploy three placeholder Lambda handlers (`fetch_lists`, `process_list`, `notify`) with IAM for DynamoDB, Secrets Manager, and S3 read access.
- Add a Step Functions workflow (fetch → Map over process → notify) with per-list error handling so a failed `process_list` still routes through `notify` and other Map iterations continue.
- Schedule the workflow via EventBridge (default: every 5 minutes).
- Build Lambda zip artifacts during deploy instead of tracking them in git.

**Validation library (`src/contribot/lib/`)**
- Add `ContriBot` in [`lib/contribot.py`](src/contribot/lib/contribot.py) to validate contributor Excel workbooks: table- and column-level checks, optional auto-fixes, duplicate detection, and annotated output workbooks (Summary, Findings, Similarities, Fixes sheets).
- Include supporting modules (`known_countries.py`, `utils.py`) and 81 pytest unit tests.
- Document architecture, configuration, and local test instructions in [`src/contribot/README.md`](src/contribot/README.md).

Lambda handlers still return stub data and do not call the validation library, Open Supply Hub API, Google Drive, Slack, or Monday yet. Secret values must be populated manually in AWS after deploy.

Ticket: [OSDEV-2928](https://opensupplyhub.atlassian.net/browse/OSDEV-2928)

## Test plan

- [x] Run `terraform validate` in `deployment/terraform/`
- [x] Deploy to a non-production environment via workflow_dispatch
- [x] Confirm Secrets Manager stores, DynamoDB table, Lambdas, Step Functions state machine, and EventBridge rule are created
- [x] Manually start the ContriBot state machine and verify all three Lambda steps succeed
- [x] Trigger a `process_list` failure in the Map state and confirm execution continues to `notify` for that iteration
- [x] Confirm the EventBridge rule triggers scheduled executions
- [x] Run validation unit tests: `cd src/contribot/lib && python -m pytest tests/test_contribot.py`

[OSDEV-2928]: https://opensupplyhub.atlassian.net/browse/OSDEV-2928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/contribot/lib/contribot.py
```diff
@@ -0,0 +1,2046 @@
...
+                    column = cols[cnt]
+                    if c == "Keep":
+                        ws[f"{column}{row}"] = False
+                    elif c == "Value":
+                        if str(r["cell"]).lower() != "nan":
+                            ws[f"{column}{row}"] = self.wb[self.sourcesheet][
+                                r["cell"]
+                            ].value
+                    else:
+                        if c == "cell" and str(r[c]).lower != "nan":
```
VadimKovalenkoSNF, line 359: Logic bug 1/3: `.lower` method compared instead of called. Needs `lower()`

src/contribot/lib/contribot.py
```diff
@@ -0,0 +1,2046 @@
...
```
VadimKovalenkoSNF, line 512: Logic bug 3/3: `.lower` method compared instead of called. Needs `lower()`

src/contribot/lib/contribot.py
```diff
@@ -0,0 +1,2046 @@
...
```
VadimKovalenkoSNF, line 415: Logic bug 2/3: `.lower` method compared instead of called. Needs `lower()`

src/contribot/lib/contribot.py
```diff
@@ -0,0 +1,2046 @@
...
+            self.wb._sheets[summary_sheet_no],
+            self.wb._sheets[findings_sheet_no],
+            self.wb._sheets[similarities_sheet_no],
+            self.wb._sheets[0],
+            self.wb._sheets[fixes_sheet_no],
+        ] + [self.wb._sheets[r] for r in residual[1:]]
+
+        self.df = pd.read_excel(filename)
+        if len(self.df) > 10000:
+            self._add_diagnosis(code="T0016", num_lines=len(self.df))
```
VadimKovalenkoSNF, line 103: Can you check: 
`__init__` calls `self._add_diagnosis(code="T0016", ...)`, but `self.diagnostics_table/self.diagnostics_column` and `self._get_config()` don't run until lines 120–130. Construction will raise on exactly the large-file case the check is meant to handle. Move the state init and `_get_config()` above the row-count check.

src/contribot/lib/contribot.py
```diff
@@ -0,0 +1,2046 @@
...
+import pandas as pd
+import rapidfuzz
+from jinja2.nativetypes import NativeEnvironment
+from jinja2.nodes import TemplateData
+from openpyxl.comments import Comment
+from openpyxl.formatting.rule import ColorScale, FormatObject, Rule
+from openpyxl.styles import Alignment, Color, Font
+from openpyxl.utils.dataframe import dataframe_to_rows
+from sqlalchemy import create_engine
+from unidecode import unidecode
```
VadimKovalenkoSNF, line 34: `contribot.py`  imports `cleanco`, `rapidfuzz`, `jinja2`, `unidecode`, and `sqlalchemy` -  none of which are in `requirements.txt` or `requirements-dev.txt`

src/contribot/lib/contribot.py
```diff
@@ -0,0 +1,2046 @@
...
+        df["num_lines"] = self.summary["num_lines"]
+        df["num_errors"] = self.summary["num_errors"]
+
+        try:
+            if engine:
+                with engine.connect() as conn:
+                    df.to_sql(table_name, con=conn, index=False, if_exists="append")
+            else:
+                with create_engine(
+                    "mysql+mysqlconnector://opensupplyhub:secret@localhost/opensupplyhub"
```
VadimKovalenkoSNF, line 229: Please, check this suggestion provided by the Cursor:

> This PR is the remediation, yet the library it connection string and writes diagnostics straight to a DB, wrapped in a bare except: pass that silently swallows failures (and falls back to an ephemeral local SQLite in a Lambda). Even though the stub handlers don't call save() yet, this is the thing the migration is supposed to remove.

VadimKovalenkoSNF (changes requested):
Nice work! Please take a look at the review comments I added with the help of AI. Let me know if anything is unclear or if you'd like to discuss any of the suggestions.

src/contribot/lib/contribot.py
protsack-stephan: Fixing Contribot is out of scope for this specific ticket; for now, we just want to get things working. This is definitely something we should look into in the future.

src/contribot/lib/contribot.py
protsack-stephan: Fixing Contribot is out of scope for this specific ticket; for now, we just want to get things working. This is definitely something we should look into in the future.

src/contribot/lib/contribot.py
protsack-stephan: Fixing Contribot is out of scope for this specific ticket; for now, we just want to get things working. This is definitely something we should look into in the future.

src/contribot/lib/contribot.py
protsack-stephan: Fixing Contribot is out of scope for this specific ticket; for now, we just want to get things working. This is definitely something we should look into in the future.

src/contribot/lib/contribot.py
protsack-stephan: Fixing Contribot is out of scope for this specific ticket; for now, we just want to get things working. This is definitely something we should look into in the future.

src/contribot/lib/contribot.py
protsack-stephan: Good point, need to make sure those are installed.

protsack-stephan: Updated.

VadimKovalenkoSNF (dismissed):
Approved, good work!

doc/release/RELEASE-NOTES.md
```diff
@@ -3,6 +3,22 @@ All notable changes to this project will be documented in this file.
...
+
+## Introduction
+* Product name: Open Supply Hub
+* Release date: August 7, 2026
+
+### Architecture/Environment changes
+* [OSDEV-2928](https://opensupplyhub.atlassian.net/browse/OSDEV-2928) - Provisioned ContriBot AWS infrastructure in Terraform: four empty Secrets Manager stores for runtime credentials, an on-demand DynamoDB state table keyed by `list_id`, three placeholder Lambda functions (`fetch_lists`, `process_list`, `notify`), a Step Functions workflow (fetch → Map over process → notify), and an EventBridge schedule (default every 5 minutes). Handlers are stubs only; secret values must be populated manually in AWS before real processing can run. Lambda dependency bundling for production handler code is not yet wired into the deploy pipeline.
+
+### Release instructions
+* Ensure that the following commands are included in the `post_deployment` command:
```
vlad-shapik, line 16: Could I ask you to remove the 2.27 related command from post_deployment command?

protsack-stephan: Updated.

vlad-shapik: I meant cleaning up this command code and removing the commands that are no longer necessary for the 2.28 release.
Screenshot 2026-07-17 at 15 15 09

vlad-shapik: Could you please remove backfill_parallel_worker_count and BACKFILL_PARALLEL_BY_ENVIRONMENT as well?

madm2n: Yeah, good point. Updated.

vlad-shapik (approved):
LGTM!

deployment/terraform/contribot_lambda.tf
```diff
@@ -0,0 +1,183 @@
...
+
+  tags = merge(local.default_tags, {
+    Name = "funcContribotFetchLists"
+  })
+}
+
+resource "aws_lambda_function" "contribot_process_list" {
+  filename         = "${path.module}/lambda-functions/contribot_process_list/contribot_process_list.zip"
+  source_code_hash = filebase64sha256("${path.module}/lambda-functions/contribot_process_list/contribot_process_list.zip")
+  function_name    = "func${local.short}ContribotProcessList"
```
valencia-max, line 138: Vanta is flagging some issues related with the creation of these functions without a metric alarm. We can create the alarm manually or using Terraform for this -- I'd prefer to use Terraform, since last time I created the alarms for the rest of functions manually but the best approach would be to do it with Terraform.
You can see the whole explanation here: https://app.vanta.com/c/opensupplyhub.org/tests/serverless-function-error-rate-monitored-aws?tab=results

deployment/terraform/contribot.tf
```diff
@@ -0,0 +1,75 @@
...
+}
+
+#
+# ContriBot DynamoDB state table
+#
+# Tracks processed facility lists by list ID. Lambdas read this table via
+# CONTRIBOT_STATE_TABLE_NAME to skip already-handled lists and resume after failures.
+#
+
+resource "aws_dynamodb_table" "contribot_state" {
```
valencia-max, line 54: Vanta is flagging this:
- This test verifies whether each AWS DynamoDB table has a configured CloudWatch alarm for monitoring the ConsumedReadCapacityUnits metric.
- This test verifies whether each AWS DynamoDB table has a configured CloudWatch alarm for monitoring the ConsumedWriteCapacityUnits metric.

You can take a look here:
- https://app.vanta.com/c/opensupplyhub.org/tests/nosql-database-monitored-and-alarmed-config-read?tab=results
- https://app.vanta.com/c/opensupplyhub.org/tests/nosql-database-monitored-and-alarmed-config-write?tab=results

Vanta says that this test is not relevant for us, but it worth to double check.

src/contribot/fetch_lists/handler.py
```diff
@@ -0,0 +1,7 @@
+def handler(event, context):
+    return {
+        "lists": [
+            {"list_id": "dummy-1"},
```
valencia-max, line 4: These dummy ids are intentional?

src/contribot/requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+openpyxl>=3.1
+google-api-python-client>=2.100
+google-auth>=2.23
+cleanco>=2.2
+rapidfuzz>=3.0
+Jinja2>=3.1
+SQLAlchemy>=2.0
+Unidecode>=1.3
+tqdm>=4.66
+mysql-connector-python>=8.0
```
valencia-max, line 14: Based or our previous/ongoing work on vulnerabilities:

All 14 packages are clean at their **latest** versions — a fresh install today has no known vulnerabilities.

However, 4 of the `>=` floors permit vulnerable versions, so a cached/constrained resolution could still install a bad one:

| Package | Floor allows | Known vulns | Suggested floor |
|---|---|---|---|
| `requests` | 2.31.0 | 3 (PYSEC-2026-1872/1873/2275; fixes ≤2.33.0) | `>=2.33` |
| `Jinja2` | 3.1.0 | 5 (sandbox escape/XSS; fixes ≤3.1.6) | `>=3.1.6` |
| `tqdm` | 4.66.0 | 1 (PYSEC-2026-1976; fixed 4.66.3) | `>=4.66.3` |
| `mysql-connector-python` | 8.0.x | 1 (PYSEC-2026-1685; fixed 9.1.0) | `>=9.1` |

Suggested change:

```diff
-requests>=2.31
+requests>=2.33
-Jinja2>=3.1
+Jinja2>=3.1.6
-tqdm>=4.66
+tqdm>=4.66.3
-mysql-connector-python>=8.0
+mysql-connector-python>=9.1
```

Notes:
- `mysql-connector-python` 8→9 is a major version bump — API largely compatible, but worth a quick test pass against the MySQL workflows.
- Longer term: floor-style requirements drift silently; consider pinning exact versions (e.g. `pip-compile`) so vulnerability scans are reproducible.

valencia-max (approved):
Left some commnets, overall looks good to me!

deployment/terraform/contribot.tf
protsack-stephan: We don't really need this at this point, I can think about this further on in the implementation.

src/contribot/requirements.txt
protsack-stephan: I'm going to deal with this in the follow up pr, those dependencies need cleanup for sure. Thanks for flagging.

src/contribot/fetch_lists/handler.py
protsack-stephan: Yes, this is only infra setup at this point.

deployment/terraform/contribot_lambda.tf
protsack-stephan: Let me take a look in the follow up PR, thanks for flagging!

PR #1121, opened 2026-07-13
[OSDEV-2786] Optimize anonymise_in_paid_products migration

protsack-stephan:
## Summary

Optimizes migration `0219_add_contributor_anonymise_in_paid_products.py`, a follow-up to the OSDEV-2786 feature (#1087), so it applies quickly on large tables.

- **Drop the full-table backfill.** The original ran `ADD COLUMN` (nullable) → `UPDATE ... SET false` → `ALTER COLUMN SET NOT NULL` for both `api_contributor` and `api_historicalcontributor`. The `UPDATE` rewrites every row and generates a lot of WAL. This is replaced with a single `ADD COLUMN ... NOT NULL DEFAULT false` per table, which on PostgreSQL 11+ is a metadata-only operation (no table rewrite).
- **Carry the DB default forward, explicitly.** The `DEFAULT false` now stays on the column, and the field declares `db_default=False` — both on `Contributor.anonymise_in_paid_products` and in the migration's `state_operations` — so Django's recorded state matches the actual DB schema (no schema-vs-state divergence, no spurious `makemigrations` diff).

`state_operations` and `reverse_sql` behavior are otherwise unchanged.

## Notes

- Relies on PostgreSQL 11+ for the metadata-only `ADD COLUMN ... DEFAULT` behavior (the case in all our environments).
- Requires Django 5.0+ for `db_default` (repo is on 5.2).

## Test plan

- [ ] `docker compose exec django python manage.py migrate` applies cleanly
- [ ] `docker compose exec django python manage.py makemigrations --check` reports no changes
- [ ] Column `anonymise_in_paid_products` exists on `api_contributor` / `api_historicalcontributor` as `NOT NULL` with `DEFAULT false`
- [ ] Reverse migration drops the columns

### DIFF
--- a/doc/release/RELEASE-NOTES.md
+++ b/doc/release/RELEASE-NOTES.md
@@ -14,7 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 #### Migrations
 * `0217_add_contribution_dates_to_index_contributors.py` - Updates the `index_contributors()` SQL function to include `list_uploaded_at` (from `FacilityList.created_at`) and `last_contributed_at` (from `FacilityListItem.updated_at`) in the indexed contributor JSON.
 * `0218_add_updated_at_to_index_claim_info.py` - Updates the `index_claim_info()` SQL function to include the claim's `updated_at` timestamp in the indexed `claim_info` JSON, so the claimed section can display the latest edit date instead of the original claim date. See OSDEV-2679.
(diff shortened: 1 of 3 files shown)


VadimKovalenkoSNF (approved):
Good catch, approved!

mValencia123 (approved):

valencia-max (approved):
LGTM! (again)

PR #1126, opened 2026-07-14
[OSDEV-2652] Allow contributors to deactivate uploaded lists

VadimKovalenkoSNF:
Implements [OSDEV-2652](https://opensupplyhub.atlassian.net/browse/OSDEV-2652)

Allow contributors to deactivate their own approved facility lists through the legacy API.

Contributors need a self-service way to remove all associations from an uploaded list without deleting facilities or requiring OS Hub staff intervention. The existing per-facility `POST /api/facilities/{id}/dissociate/` operation remains unchanged; this PR adds only whole-list deactivation while the RESTful v1 design is considered separately.

## What changed

- Add `POST /api/facility-lists/{list_id}/deactivate/` to the legacy API and Swagger documentation.
- Require an authenticated, confirmed contributor who owns the requested list; missing and non-owned lists return the same `404` to avoid leaking ownership.
- Allow only approved, active lists. Pending lists return `400`, and repeated deactivation returns `404`.
- Reuse the existing rejection behavior without sending a rejection email: set `FacilityList.status` to `REJECTED`, clear `replaces` when present, and deactivate the associated `Source`.
- Preserve all facilities and matches while anonymizing the contributor across affected facility profiles.
- Keep unrelated facility-list admin operations hidden from Swagger and add no v1 route.

## Validation

- Added coverage for successful deactivation, profile anonymization, ownership isolation, pending/inactive/orphan lists, authentication, confirmed email, durability, admin-rejection parity, removed v1 routing, and Swagger schema.
- 14 focused list-deactivation tests pass.
- 46 list and existing facility-dissociation regression tests pass.
- Flake8 and diff checks pass.

## Release notes

Updated the current release entry. No migration or manual reindex is required; the existing `api_source` trigger propagates the change.

[OSDEV-2652](https://opensupplyhub.atlassian.net/browse/OSDEV-2652)

[OSDEV-2652]: https://opensupplyhub.atlassian.net/browse/OSDEV-2652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/tests/test_facility_lists_deactivate.py
```diff
@@ -0,0 +1,295 @@
...
+        self.other_password = 'example456'
+        self.other_user, self.other_contributor = self.__create_contributor(
+            self.other_email, self.other_password, 'Contributor Two'
+        )
+        self.other_list, self.other_source = self.__create_list(
+            self.other_contributor, 'Other Supplier List'
+        )
+
+        self.superuser_email = 'superadmin@example.com'
+        self.superuser_password = 'superpass'
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: It's just a test suite.

valencia-max (dismissed):
Awesome work!

protsack-stephan:
@VadimKovalenkoSNF For now, we should probably add those endpoints to the old API. We need to think about how this will fit into the RESTful v1 API, which isn't obvious to me at the moment. However, if we document those in our old API docs and keep them in the old API, we can buy ourselves some time to think about this.

src/django/api/services/facility_list_deactivation_service.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Returns True if the list was deactivated, False if its source was already
+    inactive.
+    """
+    source = facility_list.source
+
+    if not source.is_active:
+        return False
+
+    facility_list.status_change_reason = 'Deactivated by the contributor.'
```
valencia-max, line 30: nitpick: We could add this to a constants file

src/django/api/services/facility_list_deactivation_service.py
```diff
@@ -0,0 +1,44 @@
...
+    facility_list.status_change_by = user
+    facility_list.status = FacilityList.REJECTED
+    facility_list.save()
+
+    if facility_list.replaces_id:
+        facility_list.replaces = None
```
valencia-max, line 36: The definition of this field in the data model says:
`If not null this list is an updated version of the list specified by this field.`

As we are deactivating the current list, and this is an updated version of the linked list in `replaces_id`, should we also deactivate that list?

valencia-max (dismissed):

src/django/api/services/facility_list_deactivation_service.py
VadimKovalenkoSNF: The replaced list is already inactive: approving the replacement deactivates its source. We should not reject it because that would rewrite its historical moderation status. This flow only deactivates the current list and clears replaces, matching the existing admin reject behavior.

src/django/api/services/facility_list_deactivation_service.py
VadimKovalenkoSNF: Since this message is only used by this service, I’d prefer to keep it local. I can extract it if we expect it to be reused or treated as shared domain vocabulary.

doc/release/RELEASE-NOTES.md
```diff
@@ -45,6 +55,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Release instructions
 * Ensure that the following commands are included in the `post_deployment` command:
     * `migrate`
+* No reindex is required: list deactivation propagates through the existing `api_source` indexing trigger.
```
vlad-shapik: Could you clarify what this is referring to? I understand it from the PR context, but once it's merged, it will be unclear while reading it without the PR history. It also seems to break the list in the middle.
Also, I'm wondering if we really need this note. If nothing requires reindexing, we probably don't have to mention it in the instructions. If something did require reindexing, it would be added anyway, and then this info note would make sense to have - otherwise it might just bloat this section a bit. What do you think?

vlad-shapik (changes requested):

doc/release/RELEASE-NOTES.md
VadimKovalenkoSNF: Thanks, I removed this redundant instruction.

vlad-shapik (approved):
LGTM

valencia-max (approved):

PR #1130, opened 2026-07-15
[OSDEV-2928] Extend admin facility lists endpoint for list-moderation workflows

protsack-stephan:
## Summary

Extend `GET /api/admin-facility-lists/` so superusers can iterate facility lists in id order and see contributor contact details without extra queries per row.

- Add optional `ordering` query param (`created_at`, `-created_at`, `id`, `-id`; default remains `created_at` ascending).
- Add optional `id__gt` cursor filter for id-based pagination (`?ordering=id&id__gt=100`).
- Return `contributor_name` and `contributor_email` on each list (from the list's source contributor and admin user).
- Join `source`, `contributor`, and `admin` via `select_related` to avoid N+1 lookups for the new fields.

Existing clients are unchanged when the new query params are omitted.

## Test plan

- [x] `docker compose exec django python manage.py test api.tests.test_dashboard_list`
- [ ] Confirm `GET /api/admin-facility-lists/?ordering=id&id__gt=0` returns lists in ascending id order with contributor fields populated
- [ ] Confirm omitting new params preserves previous dashboard behavior


VadimKovalenkoSNF (approved):
Looks clear, approved.

daniellekain (approved):

PR #1131, opened 2026-07-15
[OSDEV-2949] Strip exposed internal_ID from rsc_grievance_mechanism partner field

VadimKovalenkoSNF:
## Summary

Fixes [OSDEV-2949](https://opensupplyhub.atlassian.net/browse/OSDEV-2949)

`GET /api/facilities/{os_id}/` and `GET /api/v1/production-locations/{os_id}/` were returning an `internal_ID` value on the `rsc_grievance_mechanism` partner field. That key was ingested from the partner's list contribution (used only to map their internal IDs to OS IDs) but is **not** part of the partner field's JSON Schema, and the partner explicitly asked us not to expose it.

Root cause: the partner field JSON Schemas either omitted `additionalProperties` or placed it *inside* `properties`, where JSON Schema treats it as a property definition rather than the keyword — so extra keys like `internal_ID` were never rejected and got stored in `ExtendedField.value['raw_values']`.

This PR fixes the already-stored data. The schema guarding (adding `additionalProperties: false` at the correct level across **all** object-type partner fields) is applied in Django admin and tracked in the release notes.

## Changes

- **New management command** `remove_rsc_grievance_mechanism_nested_internal_ids`:
  - Removes the nested `internal_ID` from every `rsc_grievance_mechanism` `ExtendedField.value['raw_values']` (~1.7k rows).
  - Reindexes **only** the affected production locations via `index_facilities_by` (targeted; guarded so it can never fall through to a full `index_facilities` reindex).
  - Uses `bulk_update` on `value` only, so `updated_at` is not bumped and no new history rows are created — the data looks as if the key was never contributed.
  - Idempotent; supports `--dry-run`, `--batch-size`, and `--no-reindex`.
- **`post_deployment.py`**: wired the command in as a one-time 2.28.0 cleanup (marked temporary; to be removed after the release is deployed everywhere).
- **Release notes**: added a `2.28.0` entry (Bugfix, Code/API changes, Release instructions), noting that all object-type partner field schemas are being guarded in admin.

## Test plan

- [x] `python manage.py remove_rsc_grievance_mechanism_nested_internal_ids --dry-run` reports the expected row/location counts without writing.
- [x] After a real run, `GET /api/facilities/{os_id}/` and `GET /api/v1/production-locations/{os_id}/` no longer include `internal_ID` for `rsc_grievance_mechanism`.
- [x] Re-running the command is a no-op ("nothing to do").
- [x] Verify only the affected `api_facilityindex` rows were touched (no full reindex).
- [ ] Confirm timing on Staging before Production.

[OSDEV-2949]: https://opensupplyhub.atlassian.net/browse/OSDEV-2949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/management/commands/remove_rsc_grievance_mechanism_nested_internal_ids.py
```diff
@@ -0,0 +1,206 @@
...
+            dest='reindex',
+            default=True,
+            help=(
+                'Skip reindexing the affected locations (the heaviest step). '
+                'The ExtendedField rows are still cleaned, but the API keeps '
+                'serving the stale index until a reindex runs.'
+            )
+        )
+
+    def handle(self, *args, **options):
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Finding: valid — fixed. `handle()` is now decorated with `@transaction.atomic`, and the per-batch with `transaction.atomic()` was removed from `flush_batch`. All `batch bulk_updates` and the `final _reindex` call now run inside the single outer transaction, so any failure rolls back every database change (both the `ExtendedField` edits and the `api_facilityindex reindex`).

src/django/api/management/commands/post_deployment.py
```diff
@@ -48,3 +48,9 @@ def handle(self, *args, **options):
         # attribution for facilities promoted before the OSDEV-2197 fix. Remove
         # after the release has been deployed everywhere.
         call_command('reindex_promoted_locations')
```
protsack-stephan: Can you please remove both of those commands, those are no longer needed.

src/django/api/management/commands/remove_rsc_grievance_mechanism_nested_internal_ids.py
```diff
@@ -0,0 +1,206 @@
...
+            .order_by('id')
+        )
+
+        total = base_qs.count()
+        if total == 0:
+            self.stdout.write(self.style.SUCCESS(
+                f"No '{self.FIELD_NAME}' records contain a nested "
+                f"'{self.INTERNAL_ID_KEY}'; nothing to do."
+            ))
+            return
```
protsack-stephan: This will slow down the command; instead of performing a count, we could just run the query and check whether any rows are returned.

src/django/api/management/commands/remove_rsc_grievance_mechanism_nested_internal_ids.py
```diff
@@ -0,0 +1,206 @@
...
+                    f"Progress: scanned={stats['scanned']}/{total} "
+                    f"updated={stats['updated']} "
+                    f"would_update={stats['would_update']} "
+                    f"skipped={stats['skipped']}"
+                )
+                last_log = now
+
+        flush_batch()
+
+        self._reindex(affected_facility_ids, dry_run, reindex)
```
protsack-stephan: Do we need to trigger the re-indexing manually? I am not sure, but it looks like since we are not disabling the triggers after updating the data, those IDs will be re-indexed automatically.

protsack-stephan (commented):
Good work!

src/django/api/management/commands/post_deployment.py
valencia-max: Yes, it's no longer needed. I created this ticket to remove this line after deployment: https://opensupplyhub.atlassian.net/browse/OSDEV-2954

src/django/api/management/commands/remove_rsc_grievance_mechanism_nested_internal_ids.py
valencia-max: The trigger doesn't execute "index_facility" direclty, it's calling a different function that it's only updating the extended fields and another fields, I think we don't need to call reindex direclty, since we are just modifying a field. But maybe for another field it would require to executed manually.

valencia-max (dismissed):

src/django/api/management/commands/post_deployment.py
valencia-max: @VadimKovalenkoSN, if you remove this line here, feel free to assing yourself the above ticket, otherwise I could remove the line and track it that ticket.

src/django/api/management/commands/remove_rsc_grievance_mechanism_nested_internal_ids.py
VadimKovalenkoSNF: Very good catch. I confirm that we don't need to perform a re-index explicitly here. I've added a test `test_bulk_update_trigger_refreshes_facility_index` to prove this.

src/django/api/management/commands/remove_rsc_grievance_mechanism_nested_internal_ids.py
VadimKovalenkoSNF: Good point, fixed.

src/django/api/management/commands/post_deployment.py
VadimKovalenkoSNF: I removed those stale commands.

Also, if there's already a dedicated ticket for this work, It would be great to make those changes in a separate PR to keep the scope focused and align with our development process.

VadimKovalenkoSNF: Also, I removed those commands in one of a previous [commit](https://github.com/opensupplyhub/open-supply-hub/pull/1131/files#diff-c1e5bb839c9e6cf9d6eed00b270250b6519481663e438ba1f3e2d03d94654fb4), but it looks like they were silently resurrected during a merge from `main` >_<

vlad-shapik:
@VadimKovalenkoSNF 
Could you please handle the MR conflicts?

protsack-stephan (dismissed):

valencia-max (dismissed):

vlad-shapik (dismissed):

protsack-stephan (approved):

vlad-shapik (approved):

PR #1135, opened 2026-07-16
[OSDEV-3093] Mark approved claimants' contributions as claim data (Claimed chip fix)

daniellekain:
## What

Make the "Claimed" data attribution (`is_from_claim` on extended fields) recognize data the approved claimant contributed through **any** channel — SLC, list upload — not only fields created directly on the `FacilityClaim`.

## Why

Second week-1 quick win from the [Promotion Logic Q3 Report](https://opensupplyhub.atlassian.net/wiki/spaces/SD/pages/1363410945/Promotion+Logic+Q3+Report+Options+Recommendation): a claimant who submits data via SLC or list upload sees it chipped "Crowdsourced" on their own facility page, because the chip only recognized `FacilityClaim`-sourced fields (`facility_list_item_id IS NULL`). This mislabels the owner's own data and undercuts the value of claiming.

## How

- `FacilityIndexSerializer.get_extended_fields` resolves `approved_claim.contributor_id` (already present in the indexed JSON) and passes it into the `FacilityIndexExtendedFieldListSerializer` context. The details serializer inherits this path, so the production location page and the legacy `/api/facilities/{id}/` API both benefit.
- `_get_is_from_claim` returns true when the field has no list item (as before) **or** when its contributor matches the approved claimant. No claim in context → previous behavior exactly.
- No migration or reindex needed — both ids are already in the indexed data.

### Behavior notes (beyond the chip)

- **This also changes value promotion, not just the label.** `is_from_claim` is a sort key in `_sort_order`/`_sort_order_excluding_date`, so the claimant's contributions now rank above other contributors' unverified values — the top/displayed value of an extended field can change. This is the intended direction of the promotion-logic plan (claimant data wins), and an integration test pins it down.
- **Masked contributions are never attributed to the claim.** If the claimant's contribution is masked (anonymized), marking it "Claimed" would link it to the publicly named claimant by inference, so `_get_is_from_claim` checks `is_contribution_masked` first.
- **Partner fields keep the previous behavior** — `__serialize_and_sort_partner_fields` builds its own serializer context without a claimant id. Partner fields are system-sourced, so the chip does not apply there.
- **Out of scope: v1 and downloads.** `/api/v1/production-locations/` is served from a separate OpenSearch-backed pipeline (`sync_production_locations.sql`), whose `COALESCE` chains have no notion of the approved claimant's contributor — this PR does not touch it. Data downloads' `claim_*` columns and generic extended-field columns are likewise unaffected. Both are real product-scope questions, tracked as follow-ups: [OSDEV-3115](https://opensupplyhub.atlassian.net/browse/OSDEV-3115) (v1 API), [OSDEV-3116](https://opensupplyhub.atlassian.net/browse/OSDEV-3116) (downloads).

## Testing

- Unit tests: claim-created field, claimant list contribution, other contributor, unclaimed facility, masked claimant contribution.
- Integration test through `FacilityIndexDetailsSerializer`: verifies the `claimant_contributor_id` wiring from `facility.approved_claim` and that the claimant's older contribution is promoted above another contributor's newer value.
- `test_facility_index_extended_field_list_serializer`, `test_facility_index_details_serializer` pass.

Ticket: [OSDEV-3093](https://opensupplyhub.atlassian.net/browse/OSDEV-3093)

[OSDEV-3093]: https://opensupplyhub.atlassian.net/browse/OSDEV-3093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OSDEV-3115]: https://opensupplyhub.atlassian.net/browse/OSDEV-3115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


doc/release/RELEASE-NOTES.md
```diff
@@ -26,7 +27,6 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 * Ensure that the following commands are included in the `post_deployment` command:
     * `migrate`
 
-
 ## Release 2.27.0
```
VadimKovalenkoSNF, line 40: Please, make sure there are 2 empty lines between releases here.

VadimKovalenkoSNF (changes requested):

VadimKovalenkoSNF (dismissed):

vlad-shapik:
@daniellekain

LGTM

One thing I'd like to confirm before approving, on the **"FE and API both benefit"** claim:

- ✅ **FE is fully covered** — both the legacy facility page and the new production-location page (`ProductionLocationDetailsContainer` → `fetchSingleFacility` → `/api/facilities/{osId}/`) go through `FacilityIndexDetailsSerializer`, so they inherit the patched `get_extended_fields`.
- ⚠️ **v1 `/api/v1/production-locations/` is not touched by this change.** That endpoint is served from OpenSearch, and its extended-field values come from a separate SQL path (`src/logstash/sql/sync_production_locations.sql`) that picks one top value per field via `COALESCE` (claim-created → promoted match → latest non-claim). It doesn't promote a claimant's SLC/list-upload contributions. So after this PR the FE/`/api/facilities/` promotes claimant data, while v1 keeps the old ordering — a behavioral divergence between the two APIs.

So my only question is scope: does OSDEV-3093 / the Promotion Logic Q3 plan intend "claimant data wins" to apply to the v1 public API too?
- If **yes**, we'd need a follow-up on the `COALESCE` chains in `sync_production_locations.sql` (+ reindex).
- If **no** (this is strictly the FE Claimed chip), then this is complete as-is — I'd just suggest tightening the PR wording so "API" clearly refers to the legacy facilities API, and ideally filing the v1 follow-up so the divergence is intentional and documented.

vlad-shapik:
@daniellekain

Also checked **data downloads** — one more consistency point to flag alongside the v1 one.

But there's an expectation mismatch worth deciding on: after this change, a claimant's SLC/list contribution shows as **"Claimed"** on the location profile. A user who then downloads the data would reasonably expect that same value under the `claim_*` columns — but it won't be there. `claim_*` will keep showing only the claim-form value, while the SLC/list value sits in the generic column mixed in with crowdsourced entries. So the profile says "Claimed" but the download attributes it as just another contribution.

valencia-max (dismissed):
🔥

daniellekain:
@vlad-shapik 

confirmed both of those issues -- thanks for the careful review!
 
re scope: I think the legacy fix only most closely aligns with the intent of the "quick win" mentioned in the Promotion Logic doc and the ticket.

I clarified the description so it says explicitly that this is a legacy API only change. I also filed two follow-up tickets so the divergence you found is tracked rather than silently inconsistent:
 
  - OSDEV-3115 (https://opensupplyhub.atlassian.net/browse/OSDEV-3115) (v1 API)
  - OSDEV-3116 (https://opensupplyhub.atlassian.net/browse/OSDEV-3116) (downloads)

I don't like that we are introducing inconsistency, but 3115 is substantially more work than this ticket. So I'm leaning towards shipping this and waiting on https://opensupplyhub.atlassian.net/browse/OSDEV-3115. lmk what you think!

valencia-max (dismissed):

vlad-shapik (approved):
LGTM

doc/release/RELEASE-NOTES.md
```diff
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
...hub.atlassian.net/browse/OSDEV-2949) - Fixed the `internal_ID` field being exposed in the API for the `rsc_grievance_mechanism` partner field. The key had been ingested from the partner's list contribution (used only to map their IDs to OS IDs) but is not part of the partner field's JSON Schema, and the partner explicitly asked us not to share it. Two parts: (1) all object-type partner field JSON Schemas were guarded in Django admin with `additionalProperties: false` placed as a sibling of `properties` (at the root and, where applicable, inside each nested object) — not just `rsc_grievance_mechanism` — so redundant keys like `internal_ID` are rejected on future contributions across every partner field; previously `additionalProperties` had been either absent or mistakenly nested *inside* `properties`, where JSON Schema treats it as a property definition rather than the keyword, so it had no effect. (2) The `remove_rsc_grievance_mechanism_nested_internal_ids` command strips `internal_ID` from the ~1.7k already-stored `rsc_grievance_mechanism` values, with the existing database trigger refreshing the affected facility index data, so the data looks as if the key was never contributed.
```
vlad-shapik: It reintroduces this text. There is Bugfix section already

vlad-shapik (changes requested):

doc/release/RELEASE-NOTES.md
daniellekain: thank you!

valencia-max (approved):

vlad-shapik (approved):
LGTM

PR #1140, opened 2026-07-20
[OSDEV-2417] Migrate self-hosted runner jobs to AWS CodeBuild

vlad-shapik:
Replace the self-hosted MacBook runner with an ephemeral GitHub Actions runner on AWS CodeBuild for the long-running DB jobs.

## Why

The three jobs using `runs-on: self-hosted` — **DB - Save Anonymized DB**, **DB - Apply Anonymized DB**, and the `restore_database` job of **Deploy to AWS** — run on a MacBook that must stay online 24/7 and has had repeated outages, making releases unpredictable ([OSDEV-2417](https://opensupplyhub.atlassian.net/browse/OSDEV-2417)). The jobs only need Linux compute with Docker, ~50+ GB of disk, and SSH to a public bastion; runs take 85–150 minutes. With CodeBuild's native runner integration, each queued job gets a fresh build container that registers as a just-in-time runner and is destroyed afterwards — nothing runs (or bills) between jobs (~a few $/month). GitHub-hosted runners were rejected: ~14 GB of free disk cannot hold the anonymization restore.

## What

- New `deployment/terraform/codebuild_github_runner` module: CodeBuild project `osh-github-actions-runner` (`BUILD_GENERAL1_LARGE`, Ubuntu 24.04 `standard:8.0`, privileged mode for Docker-in-Docker, 300-min timeout), a `WORKFLOW_JOB_QUEUED` webhook (what makes it act as a runner), minimal IAM role, and a log group. Enabled only in Test (`codebuild_github_runner_enabled`); the single runner serves all target environments because per-environment credentials come from GitHub environment secrets.
- The three jobs now use `runs-on: codebuild-osh-github-actions-runner-${{ github.run_id }}-${{ github.run_attempt }}`. The DB access pattern (SSH tunnel through the target environment's bastion) is untouched.
- AWS provider bumped `~> 5.8.0` → `~> 5.91` (minimum for the webhook filter and `CODECONNECTIONS` credentials).
- Release notes entry under 2.28.0.

Bastion reachability from the runner's AWS egress IPs was verified for every environment; no security group changes were needed.

## Status

Validated end to end in Development (shares an AWS account with Test): Terraform apply plus a full **DB - Save Anonymized DB** run against production data on the CodeBuild runner. The temporary Development enablement has been reverted; Test is the runner's permanent home.

Compute sizing is settled: the production dump run measured ~14.25 GB memory utilization (page-cache-heavy local restore/anonymization) and ~1.7 vCPU peak, so `BUILD_GENERAL1_LARGE` (16 GB) stays and a downgrade to `BUILD_GENERAL1_MEDIUM` (8 GB) is ruled out.

## Before applying to Test

- Merge [ci-deployment#66](https://github.com/opensupplyhub/ci-deployment/pull/66) (CodeConnections connection ARN for Test). A validation rule fails the plan with a clear message if the ARN is missing.
- If the temporary Development runner still exists, apply Terraform to Development first (destroys it), then to Test — the project name and account-level source credential are singletons in the shared account.

## Remaining test plan

- [x] Validate in Development: Terraform apply + **DB - Save Anonymized DB** on the CodeBuild runner
- [x] Confirm compute sizing from run metrics — keep `BUILD_GENERAL1_LARGE` (memory-bound, ~14.25 GB observed)
- [ ] Apply Terraform to Development (cleanup), then to Test
- [ ] Run **DB - Apply Anonymized DB** (`deploy-env: Test`) and **Deploy to AWS** to Test with `restore-db: true`
- [ ] After burn-in, remove the MacBook from the repo's runner list and decommission it

Note: the MacBook serialized all jobs; CodeBuild runs jobs in parallel. Per team decision no `concurrency` guards were added — avoid simultaneous dump/restore runs against the same environment.

[OSDEV-2417]: https://opensupplyhub.atlassian.net/browse/OSDEV-2417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployment/terraform/codebuild_github_runner/main.tf
```diff
@@ -0,0 +1,133 @@
...
+  project_name = aws_codebuild_project.github_runner.name
+  build_type   = "BUILD"
+
+  filter_group {
+    filter {
+      type    = "EVENT"
+      pattern = "WORKFLOW_JOB_QUEUED"
+    }
+  }
+}
```
protsack-stephan, line 133: Is this project created per-environment? Do I understand correctly that, in order to create a new environment (e.g., pre-prod), we would need to do so from a different set of runners? What would happen if we needed to create another new environment from scratch? My main question is whether it makes sense to keep this code here, or if we should move it to a separate repository or pipeline process.

deployment/terraform/codebuild_github_runner/README.md
```diff
@@ -0,0 +1,91 @@
+# CodeBuild GitHub Actions Runner
+
+This module provisions an AWS CodeBuild project that acts as an **ephemeral
+GitHub Actions runner** for the long-running database jobs:
+
+- **DB - Save Anonymized DB** (`.github/workflows/db_save_anonymized.yml`)
+- **DB - Apply Anonymized DB** (`.github/workflows/db_apply_anonimized.yml`)
+- **Deploy to AWS** → `restore_database` job (`.github/workflows/deploy_to_aws.yml`)
```
protsack-stephan, line 8: Why are we not running other workflows on those runners? It would be great to remove the dependency on external runners completely.

deployment/terraform/codebuild_github_runner/main.tf
vlad-shapik: No — a single project, created once in the Test account, serving jobs for *all* environments. The runner is pure compute: each job gets its target environment's credentials from GitHub environment secrets and reaches the DB via that environment's public bastion. Already validated running the dump against Production from this one runner.

New environments need **no** new runners.

deployment/terraform/codebuild_github_runner/README.md
vlad-shapik: Other workflows run on GitHub-hosted runners, which are free/unlimited for this public repo and haven't been a reliability problem. Moving all CI to CodeBuild would turn that into paid AWS minutes (tens of $/month at our CI volume vs ~$10/month for just the DB jobs).

These three jobs moved because free runners physically can't run them (~50 GB disk, ~14 GB RAM measured vs 14 GB SSD on `ubuntu-latest`).

protsack-stephan (approved):

valencia-max (approved):

PR #1145, opened 2026-07-23
[OSDEV-3114] Migrate local Kafka to KRaft and fix React dev server startup

VadimKovalenkoSNF:
## Summary

Fixes [OSDEV-3114](https://opensupplyhub.atlassian.net/browse/OSDEV-3114).

- Migrate local `docker-compose.yml` Kafka from ZooKeeper mode to single-node KRaft, removing the intermittent `NodeExistsException` on `docker compose up -d`.
- Replace the anonymous `node_modules` volume with named `react_node_modules` volume.
- Add the OSDEV-3114 bugfix entry to the Release 2.28.0 release notes.

## Context

Local developers were hitting two unrelated startup failures:

1. **Kafka** — recreating the broker container could leave a stale ZooKeeper broker znode (`/brokers/ids/1`), causing Kafka to exit before becoming healthy.
2. **React** — the local anonymous `node_modules` Docker volume contained an old dependency tree with nested PostCSS 8.2.6, so `craco start` failed under Node 22 with `ERR_PACKAGE_PATH_NOT_EXPORTED` on `postcss/lib/tokenize`.

The named volume does not automatically update stale dependencies. It makes the local dependency cache explicit and easy to inspect or reset deliberately, for example with `docker volume rm open-supply-hub_react_node_modules` before rebuilding the React service.

These changes affect local docker-compose only. AWS deploy continues to use MSK for Kafka and static React builds for the frontend.

## Test plan

- [x] `docker compose up -d kafka` — Kafka starts healthy without `NodeExistsException` after container recreation
- [x] `docker compose run --rm --no-deps react yarn install` — no `yarn.lock` changes; nested `postcss-safe-parser` PostCSS is 8.5.16
- [x] `docker compose up -d react` — dev server compiles successfully; `http://localhost:6543/` returns HTTP 200
- [x] `docker compose exec -T kafka kafka-topics.sh --bootstrap-server kafka:9092 --list` — `dedupe.hub.topic` exists
- [x] `docker compose up -d api-app` and run `python3 manage.py batch_process -a match -l <list_id>` — requires sufficient Docker memory for the dedupe-hub gazetteer build

[OSDEV-3114]: https://opensupplyhub.atlassian.net/browse/OSDEV-3114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vlad-shapik:
@VadimKovalenkoSNF Quick question on the `postcss-safe-parser/postcss` resolution change in `src/react/package.json`: did you check whether `yarn.lock` needs to be recomputed?

From what I can tell the lockfile already resolves `postcss@^8.1.0` (what `postcss-safe-parser` depends on) to `8.5.16`, so the tree may already be correct and this could be a no-op. But could you run `yarn install` in `src/react` and check the result?

VadimKovalenkoSNF:
@vlad-shapik good point!  I re-ran `yarn install` and confirmed that `yarn.lock` already resolves `postcss@^8.1.0` to `8.5.16`; it does not change after install. 
I also verified that the existing Yarn v1 resolution (`**/postcss-safe-parser/postcss`) correctly installs postcss-safe-parser’s nested PostCSS at `8.5.16`.

The resolution syntax change was unnecessary, so I reverted it and updated the release note. The React failure came from my local anonymous `node_modules` Docker volume, which contained a stale dependency tree with PostCSS `8.2.6` - not from the committed dependency configuration.

I would still like to keep the named `react_node_modules` volume. It does not automatically refresh stale dependencies, but it makes the volume explicit, stable, and easy to inspect or reset deliberately when dependencies change:

```
docker compose down
docker volume rm open-supply-hub_react_node_modules
docker compose up -d --build react
```

With the anonymous volume, Docker assigns an opaque name and may leave old volumes behind after container recreation, making this recovery path less discoverable.

vlad-shapik (approved):
LGTM

daniellekain (approved):

PR #1150, opened 2026-07-24
[OSDEV-2977] Remove closing date from claim form and API responses

VadimKovalenkoSNF:
## Summary

Fix [OSDEV-2977](https://opensupplyhub.atlassian.net/browse/OSDEV-2977)

Remove the claim closing date from the Free GHG Emissions Estimates form, claimed profile UI, legacy facility API, v1 production locations (`closed_at`), and facility CSV/XLSX downloads (`claim_closing_date`).

This PR removes closing-date data end to end, not just from user-facing surfaces:

- **PostgreSQL:** migration `0221` stops indexing `closing_date` in `claim_info` / `approved_claim`; migration `0222` drops `closing_date` from `FacilityClaim` and `HistoricalFacilityClaim`, permanently deleting stored values on deploy (stakeholder-approved).
- **OpenSearch / Logstash:** remove `closed_at` from the production-locations index mapping, JDBC sync SQL, and pipeline filters.
- **Application:** stop accepting or exposing the field in serializers, views, React forms/profiles, and Django admin.

A targeted `claim_info`/`approved_claim` backfill runs via `post_deployment` so existing indexed JSON is recomputed without `closing_date`.

Opening date and the separate facility closure workflow (`Facility.is_closed`) are unchanged.

## Test plan

- [x] Django: `api.tests.test_facility_claim`, `api.tests.test_facility_create_claim_serializer`, `api.tests.test_facility_download`, `api.tests.test_index_facilities`, `api.tests.test_facility_index_backfill_specs`
- [x] React: `utils.freeEmissionsEstimate.tests.js`, `FacilityDetailsClaimedInfo.test.js`, `ClaimDataContainer.test.js`, `ClaimedFacilitiesDetails.test.jsx`
- [x] v1 integration: `src/tests/v1/test_production_locations.py`
- [x] Manual: submit/update a claim and confirm closing date is absent from the form and profile
- [x] Manual: verify legacy API and v1 production-location detail responses omit `closing_date` / `closed_at`
- [x] Manual: verify CSV/XLSX download no longer includes `claim_closing_date`

## Release notes

Release notes updated in `doc/release/RELEASE-NOTES.md`. Deploy requires:

1. `backfill_facility_index --fields claim_info,approved_claim` (via `post_deployment`)

## Platform dependency review

Ran `.claude/hooks/check-platform-dependencies.py --manual` before push. Triggered areas (RBA, dual FE surfaces, OpenSearch/Logstash, migration indexation) were reviewed. Migration `0221` updates SQL functions only; migration `0222` drops the claim column (metadata-only `DROP COLUMN` on PostgreSQL).

[OSDEV-2977]: https://opensupplyhub.atlassian.net/browse/OSDEV-2977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/migrations/0221_remove_closing_date_from_claim_index_functions.py
```diff
@@ -0,0 +1,36 @@
...
+        '0220_remove_closing_date_from_index_claim_info.sql',
+        '0220_remove_closing_date_from_index_approved_claim.sql',
+    ])
+
+
+def revert_indexing_functions(apps, schema_editor):
+    helper.run_sql_files([
+        '0218_index_claim_info.sql',
+        '0207_index_approved_claim.sql',
+    ])
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: **Skipped — no code change needed.**

I checked the migration chain and SQL files against the finding. The current revert already restores the correct pre-0220 state:

| Function | Pre-0220 SQL file | Last migration that changed it |
|---|---|---|
| `index_claim_info` | `0218_index_claim_info.sql` | 0218 (adds `updated_at`) |
| `index_approved_claim` | `0207_index_approved_claim.sql` | 0207 (energy/throughput fields) |

Migration **0219** only adds `anonymise_in_paid_products` on `api_contributor` — it does not touch either index function. There are no `0219_*index*.sql` files in `src/django/sqls/`.

**Validation:** `0220` differs from each pre-0220 file only by removing `closing_date`:

```bash
diff 0218_index_claim_info.sql ↔ 0220_remove_closing_date_from_index_claim_info.sql  # only closing_date removed
diff 0207_index_approved_claim.sql ↔ 0220_remove_closing_date_from_index_approved_claim.sql  # only closing_date removed
```

So `0218` and `0207` are not stale revert targets — they are exactly the function definitions live at dependency `0219`, before `0220` runs. Using different filenames would require new SQL files that duplicate existing ones, with no behavioral difference.

The review note likely conflates migration numbers (0218/0219) with `index_approved_claim` still being at its 0207 SQL revision, which is expected since 0218–0219 never updated that function.

src/django/api/management/commands/post_deployment.py
```diff
@@ -1,6 +1,22 @@
...
+    'Test': 10,
+    'Staging': 10,
+    'Preprod': 10,
+    'Production': 10,
+    'Rba': 10,
+}
+
+
+def backfill_parallel_worker_count() -> int:
+    return BACKFILL_PARALLEL_BY_ENVIRONMENT[settings.ENVIRONMENT]
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Done, verify the fix.

daniellekain (changes requested):
I wanted to confirm -- are we tracking the sheet export that Francesca's requested?

daniellekain (commented):
Really thorough work getting this removed end-to-end. The mapping, sync SQL, and backfill coverage all look great!

One question on the deploy plan: do we actually need the `clear-opensearch-target=production-locations` rebuild for this change? Since `closed_at` is removed from the v1 response mapping and the query builder, the API can't return or filter on it once the new code deploys so the stale values left in the index are unreachable from outside. The rebuild would only be scrubbing invisible leftovers on ~100 documents, at the cost of ~4 hours of incomplete search results and v1 detail 404s, and it'd be the only thing putting a rebuild into 2.28.

Suggestion: keep the mapping/Logstash changes (they'll take effect automatically at whatever rebuild happens next, e.g. the one #1106 will need), but drop the mandatory rebuild step from the release notes. If we want the leftover values physically purged from the index right away, a targeted `update_by_query` removing the field would do it in seconds with no downtime. Whenever we need to rebuild for any other reason, this will get cleaned up. 

Happy to be wrong if there's a reason the rebuild is needed that I'm missing!

VadimKovalenkoSNF:
Hi @daniellekain ! Thank you for the review.
1. I've added the data backup requirement to the ticket as Acceptance Criterion 7. The day before the release, I'll export the historical data from Production and attach it to the ticket. I’ve also added the SQL scripts to the ticket - please take a look.
2. That's a very valid point, thank you. Indeed, the 4-hour rebuild is disproportionate for this change. I dropped it from `2.28` release instructions. Once the new code is live, `closed_at` is removed from the v1 response mapping and query builder, so stale values in the index aren't reachable through the API anyway.

src/django/api/migrations/0221_remove_closing_date_from_claim_index_functions.py
```diff
@@ -0,0 +1,36 @@
...
+
+def update_indexing_functions(apps, schema_editor):
+    """
+    Remove closing_date from index_claim_info and index_approved_claim JSON
+    output so claim closing dates are no longer exposed via API or downloads.
+    See OSDEV-2977.
+    """
+    helper.run_sql_files([
+        '0221_remove_closing_date_from_index_claim_info.sql',
+        '0221_remove_closing_date_from_index_approved_claim.sql',
```
vlad-shapik: Please follow the naming convention of the other SQL files, like 0221_index_claim_info.sql. Let's stick to the already defined naming pattern.

src/django/api/management/commands/post_deployment.py
```diff
@@ -35,6 +43,15 @@ def handle(self, *args, **options):
         # refreshes the affected FacilityIndex rows. Remove after the release
         # has been deployed everywhere.
         call_command('remove_rsc_grievance_mechanism_nested_internal_ids')
+        # Temporary for OSDEV-2977 — recompute claim_info and approved_claim
+        # without closing_date. Remove after the release has been deployed
+        # everywhere.
+        call_command(
+            'backfill_facility_index',
+            fields='claim_info,approved_claim',
```
vlad-shapik: Can it be added to the existing call of the `backfill_facility_index`?

doc/release/RELEASE-NOTES.md
```diff
@@ -46,6 +52,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 * Ensure that the following commands are included in the `post_deployment` command:
     * `migrate`
     * `remove_rsc_grievance_mechanism_nested_internal_ids` — one-time cleanup (OSDEV-2949) that strips the nested `internal_ID` from `rsc_grievance_mechanism` values; remove from `post_deployment` after this release has been deployed everywhere.
+    * `backfill_facility_index --fields claim_info,approved_claim --parallel 10 --batch-size 10000` — one-time backfill (OSDEV-2977) that recomputes indexed claim JSON without `closing_date`; remove from `post_deployment` after this release has been deployed everywhere.
```
vlad-shapik: Can this note be combined with lines that start with:

- `backfill_facility_index --fields processing_type` — one-time ...
- The `backfill_facility_index --fields processing_type` reindex ...

src/logstash/indexes/production_locations.json
```diff
@@ -223,10 +223,6 @@
           "type": "date",
           "format": "yyyy"
         },
-        "closed_at": {
```
vlad-shapik, line 226: I see that the production-locations OpenSearch index has been changed. I saw the comment from Danielle and your discussion. Can you verify locally that Logstash won't fail or throw errors after grabbing data via SQL to update the index once it picks up the new mapping (I believe it will pick it up since ECS will restart the task with new Docker image)? I just want you to reproduce the production case where OpenSearch does not have the new mapping but Logstash already knows about it. FYI, the clear_opensearch script clears the mapping from OpenSearch to let Logstash re-upload the new one.

cc @daniellekain

vlad-shapik (changes requested):
1. Please make sure the PR description is not stale - for example, the migration numbers.
2. MonthYearPicker.test.js still uses "Closing Date" / "Enter the closing date" as fixture labels. Cosmetic only.
3. I have left a few comments below.

src/logstash/indexes/production_locations.json
VadimKovalenkoSNF: @vlad-shapik @daniellekain 
Verified locally against the production-like scenario.

1. On main, I populated the existing production-locations index with `2,514,223` documents. Its old mapping included `closed_at`, with 133 documents containing values.
2. I switched to this branch, applied migrations `0220–0222`, and ran the targeted FacilityIndex backfill.
3. I rebuilt/restarted only Logstash with the new code, without clearing or recreating OpenSearch.

The new Logstash pipeline successfully indexed at least 300,000 documents into the unchanged old index. OpenSearch reported `index_failed: 0`, and there were no SQL, mapping, or pipeline errors. Both services remained operational.

I also confirmed that a document still containing `closed_at` in OpenSearch no longer exposes it through the v1 API.

Locally, recreating Logstash reset its container-local JDBC tracking state and triggered a broader replay. Production stores that state on EFS, so the ECS task replacement should continue incrementally.

Conclusion: `clear_opensearch` and the resulting full rebuild are not required for this change. The old mapping can safely retain the unused field until a future rebuild. Also, I will track this once we deploy this change to dev and test envs.

src/django/api/migrations/0221_remove_closing_date_from_claim_index_functions.py
VadimKovalenkoSNF: Done.

src/django/api/management/commands/post_deployment.py
VadimKovalenkoSNF: Good point. Updated.

doc/release/RELEASE-NOTES.md
VadimKovalenkoSNF: Done.

VadimKovalenkoSNF:
@vlad-shapik 

> Please make sure the PR description is not stale - for example, the migration numbers.

Updated

> MonthYearPicker.test.js still uses "Closing Date" / "Enter the closing date" as fixture labels. Cosmetic only.

While working on this test, I discovered that the `MonthYearPicker` component was only used for the closing date input. Since it's no longer needed, I removed the component along with its associated tests.

vlad-shapik (approved):
LGTM 💯 

It would be great if you could remove the requirement to clear OpenSearch from the PR description. I believe it will be OK to remove, since, based on the research, there is no need for clearing - at least for this change.

daniellekain (approved):

PR #1151, opened 2026-07-24
[OSDEV-1034] Align processing type search index with profile display

vlad-shapik:
## What
[OSDEV-1034](https://opensupplyhub.atlassian.net/browse/OSDEV-1034)
Fix the `/api/facilities/` **Processing Type** filter so it returns facilities whose processing type is shown on the location profile. Concretely, correct how the searchable `FacilityIndex.processing_type` column is populated.

## Why

Filtering by a processing type such as **Final Product Assembly** (e.g. Sector = Apparel + Processing Type = Final Product Assembly) returned **no results**, even though the value is clearly displayed under Processing Type on the location page.

Root cause: the taxonomy matcher classifies some values (like "Final Product Assembly") as a **facility type** even when they are contributed as a processing type. The old `index_processing_type()` gated `matched_values` entries on the field-type tag (`raw ? 'PROCESSING_TYPE'`), so these entries were dropped from the search index. Meanwhile the location profile displays the processing-type slot (`matched_values[3]`) regardless of that tag. The result was a display-vs-search mismatch: shown on the profile, invisible to the filter.

## How

- **`src/django/sqls/0220_index_processing_type.sql`** — new `index_processing_type()` that keeps every entry whose processing-type slot (`raw->>3`) is populated, mirroring the facility-type index (`raw->>2`) and the frontend display logic (rather than gating on the field-type tag).
- **`0220_fix_processing_type_search_index.py`** — migration applying the corrected function (reverts to `0130` on rollback). Its docstring also documents the intentionally out-of-scope case below.
- **`facility_index_backfill/specs.py`** — add a `processing_type` FacilityIndex backfill field group so existing rows can be re-indexed.
- **`post_deployment.py`** — temporarily run the `processing_type` backfill on deploy so existing facilities are corrected (flagged for removal after the release rolls out everywhere).
- **Tests** — regression tests covering a value tagged as a facility type and one tagged as a processing type; tests for the new backfill spec.
- **Release notes** — entry under the upcoming release.

## Verified

Reproduced end-to-end locally (contribute + approve a facility with Processing Type = Final Product Assembly): before the change `processing_type=Final Product Assembly` returned 0 results and the `api_facilityindex.processing_type` column was empty; after the migration + backfill the facility is returned.

## Scope note

This PR is intentionally scoped to the **processing type** filter (per the bug description). Related but distinct filter issues — sector over-matching (OSDEV-992, the "Electronics" comment), product type normalization (OSDEV-1390), facility type (OSDEV-887), parent company (OSDEV-1218) — share the same umbrella (OSDEV-1094 spike) but are not addressed here.

**Out of scope (by design):** raw processing-type text that failed taxonomy matching entirely (null processing-type slot) is still not indexed. Such values have never been searchable and cannot be queried from the search page anyway — the Processing Type filter is a non-creatable dropdown of known taxonomy values, and the API filter maps every input through `get_facility_and_processing_type`. They only surface as raw fallback text on the profile. Making them searchable is tracked as a follow-up under [OSDEV-3127](https://opensupplyhub.atlassian.net/browse/OSDEV-3127) (OSDEV-1094 umbrella).

Ticket: https://opensupplyhub.atlassian.net/browse/OSDEV-1034

## Test plan

- [ ] CI passes (Django tests)
- [ ] Migration 0220 applies cleanly and rolls back to 0130
- [ ] After deploy, filtering by a processing type shown on a profile (e.g. Final Product Assembly) returns the facility
- [ ] Remove the temporary backfill from `post_deployment.py` once the release is deployed everywhere

[OSDEV-1034]: https://opensupplyhub.atlassian.net/browse/OSDEV-1034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OSDEV-3127]: https://opensupplyhub.atlassian.net/browse/OSDEV-3127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


daniellekain (dismissed):
This looks great! 

One thing that comes to mind: in the case where an Apparel value fails taxonomy matching entirely, the profile still displays the raw contributed text (constants.jsx falls back to rawValues[i] when val[3] is null), but there's nothing in the processing-type slot to index so we still won't find the result when filtering, which looks almost like another instance of this bug when in fact it's a separate case (unmatched raw text has never been searchable). 

Does that sound correct? If so, I wonder if we should add a short note to the 0220 migration docstring and file a follow-up ticket (under the OSDEV-1094 umbrella) so we don't re-open this bug by mistake?

Feel free to disregard if this is already tracked or seems unecessary!

vlad-shapik:
> This looks great!
> 
> One thing that comes to mind: in the case where an Apparel value fails taxonomy matching entirely, the profile still displays the raw contributed text (constants.jsx falls back to rawValues[i] when val[3] is null), but there's nothing in the processing-type slot to index so we still won't find the result when filtering, which looks almost like another instance of this bug when in fact it's a separate case (unmatched raw text has never been searchable).
> 
> Does that sound correct? If so, I wonder if we should add a short note to the 0220 migration docstring and file a follow-up ticket (under the OSDEV-1094 umbrella) so we don't re-open this bug by mistake?
> 
> Feel free to disregard if this is already tracked or seems unecessary!

Hello @daniellekain,

Great catch - you're right, and it's a separate case. OSDEV-1034 was about matched values dropped by the tag gate; this is text that never matched the taxonomy, so there's no slot to index - the old and new functions both skip it, so it's not a regression.

Unmatched raw text can't be queried from search anyway (the filter is a non-creatable dropdown of known taxonomy values, and the API maps every input through the matcher), so it only shows as raw text on the profile. The real risk is a mistaken re-open, as you noted.

I've added a note to the 0220 docstring and filed [a follow-up ticket](https://opensupplyhub.atlassian.net/browse/OSDEV-3127) connected to OSDEV-1094 for making unmatched raw contributions searchable. Thanks!

daniellekain (approved):
Looks good to me! Thanks for making these changes!

valencia-max (approved):

PR #1152, opened 2026-07-24
[OSDEV-1148] Add contribution data to the v1 production-locations endpoint

daniellekain:
## What

Add contribution-related data to the v1 production-locations pipeline and endpoint, per the [endpoint schema RFC](https://opensupplyhub.atlassian.net/wiki/spaces/SD/pages/419659778/RFC+-+Production+Locations+Endpoint+Schema): `GET /api/v1/production-locations/` now returns `contributors` (`[{id, type}]`), `number_of_contributors`, and `lists` (`[{id, contributor_id, name}]`), and supports filtering by `contributor_id` / `contributor_type` (repeatable) and `sort_by=number_of_contributors`.

## Why

The v1 API had no contributor data at all — a gap flagged in the original Transition-to-OpenSearch RFC's preliminary analysis and a blocker for the search-page migration chain (OSDEV-1202, OSDEV-1596/1597, OSDEV-1808), since the search UI filters by contributor.

## How

- **Index** (`src/logstash/indexes/production_locations.json`): plain object mappings (consistent with `country`/`number_of_workers`, not `nested` — both filters are single-field, and the legacy index has identical cross-field semantics): `contributors {id: integer, type: text+keyword}`, `number_of_contributors: integer`, `lists {id: integer, contributor_id: integer, name: text+keyword}`.
- **Sync** (`sync_production_locations.sql` + `.conf` + two new ruby passthroughs): three correlated subqueries following the file's existing pattern. **Visibility mirrors the legacy FacilityIndex contributor queries** (`src/django/sqls/0130_index_contributors_id.sql`, `0130_index_contributors_count.sql`, `0130_index_lists.sql`): only `afm.is_active AND afm.status IN ('AUTOMATIC','CONFIRMED','MERGED') AND source.is_active AND source.is_public` — no non-public contributor id ever lands in the index. The count matches the legacy public-only count exactly.
- **Query plumbing**: `contributor_id` → terms on `contributors.id` (raw integer), `contributor_type` → terms on `contributors.type.keyword`, `number_of_contributors` added to the raw-sort special case (previously `claimed_at` only); serializer validates `contributor_id` as a list of integers (garbage → 400) and accepts the new `sort_by` choice; all three fields added to the base response mapping (detail inherits).

## Rollout

- The sync is **incremental** on `facility.updated_at` — existing documents gain the fields only after a **one-time full resync** of the `production-locations` index.
- **Ongoing limitation ([OSDEV-3131](https://opensupplyhub.atlassian.net/browse/OSDEV-3131)):** the incremental sync only sees facilities whose `facility.updated_at` changed. A list upload matched to an *existing* facility doesn't touch that column (dedupe-hub refreshes only the legacy Postgres index via `index_facilities_by`), and neither does making a source private/inactive. So list-based contribution changes reach `contributors`/`number_of_contributors`/`lists` only at the next full resync — including the removal direction, where a now-private source's contributor id/type and list name stay visible in v1 until then. The legacy index updates immediately via triggers; moderation-event (v1 SLC) contributions do touch `facility.updated_at` and flow through. Fixing this properly means extending the sync's tracking criteria (e.g. a computed `GREATEST` of facility/match/source `updated_at`), which needs supporting indexes and perf validation — out of scope here.
- Safe deploy order: mapping + SQL/conf first, resync, then rely on the API (sorting on an unmapped field errors; filters return misleadingly-empty results until resync completes). Since everything ships in one release, the operative step is: **schedule the full resync at deploy**.
- Sequencing note: OSDEV-3115 also modifies this SQL file and also needs a full resync — if both land in the same window, one resync covers both (cross-comments on both tickets).
- Known, safe divergence: legacy shows anonymized contrib *types* for non-public contributors; v1's `contributors.type` includes public contributors only (ids sit beside types), so `contributor_type` filtering misses locations whose only such contributor is non-public. Stricter than legacy, not a leak. Also, `contributors` is a plain object array (not `nested`), so combining `contributor_id` and `contributor_type` in one request can match a location where the id and type belong to *different* contributors — the same flattened-array semantics as the legacy index; worth a sentence in the api-docs follow-up.

## Testing

- Unit: query-builder terms/sort tests (raw `contributors.id`, `contributors.type.keyword`, raw `number_of_contributors` sort), `_source` list extension, serializer param validation (valid lists, invalid `contributor_id` → 400). 52 tests pass; flake8 clean.
- Integration (`src/tests/v1/test_production_locations.py`): fields present in list + detail responses; filter by single/multiple `contributor_id`; filter by `contributor_type`; invalid `contributor_id` → 400; sort asc/desc including legacy documents without the new fields (no error, missing sort last).
- Follow-up ([OSDEV-3130](https://opensupplyhub.atlassian.net/browse/OSDEV-3130), separate repo): update [open-supply-hub-api-docs](https://github.com/opensupplyhub/open-supply-hub-api-docs) with the new fields/params.

Ticket: [OSDEV-1148](https://opensupplyhub.atlassian.net/browse/OSDEV-1148)

[OSDEV-1148]: https://opensupplyhub.atlassian.net/browse/OSDEV-1148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OSDEV-3131]: https://opensupplyhub.atlassian.net/browse/OSDEV-3131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


doc/release/RELEASE-NOTES.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
...ssian.net/browse/OSDEV-1148) - The `GET /api/v1/production-locations/` endpoint now returns contribution-related data per the endpoint schema RFC: `contributors` (array of `{id, type}` for publicly-visible contributors), `number_of_contributors` (count of distinct public contributors), and `lists` (array of `{id, contributor_id, name}`). New query capabilities: filter by `contributor_id` and `contributor_type` (both repeatable), and `sort_by=number_of_contributors`. Only contributors from active, public sources with active matches are exposed — the same visibility rules as the legacy facilities index. Existing documents gain the fields only after a full resync of the `production-locations` OpenSearch index (the logstash sync is incremental); until then the new filters return only recently-updated locations. Note the incremental sync keys on `facility.updated_at` only, so contribution changes that do not touch the facility row (a list upload matched to an existing location, or a source later made private/inactive) are reflected in these fields only at the next full resync — a known v1-pipeline limitation tracked in [OSDEV-3131](https://opensupplyhub.atlassian.net/browse/OSDEV-3131).
```
vlad-shapik: I believe it will be released in the next release, 2.29.0?

vlad-shapik (changes requested):

doc/release/RELEASE-NOTES.md
daniellekain: ah yes thanks! I had this PR as a draft for a while

daniellekain: should be fixed now

doc/release/RELEASE-NOTES.md
```diff
@@ -46,6 +46,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
```
valencia-max: Just to be in the same page. It says that _Only contributors from active, public sources with active matches are exposed — the same visibility rules as the legacy facilities index_. The work that James did related for anonymizing SLC should be also excluded here?

doc/release/RELEASE-NOTES.md
```diff
@@ -141,6 +144,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
+* [OSDEV-1148] The `production-locations` index template gains new mappings (`contributors`, `number_of_contributors`, `lists`), and template changes only apply to a newly created index. When deploying, run the `[Release] Deploy` workflow with `clear-opensearch-target` set to `production-locations` so the index is dropped and rebuilt by the Logstash `sync_production_locations` pipeline, which also backfills the new fields for all existing locations (the incremental sync alone would not). If OSDEV-3115 lands in the same release, one rebuild covers both changes.
```
valencia-max: Can you please move this to the release v2.29.0 block? Currently it's under v2.28.0 :)

valencia-max (dismissed):
LGTM!

Just worry about cleaning open search index, but, as we need a full re-index here, I was thinking how we could take advantage of this process in this PR https://github.com/opensupplyhub/open-supply-hub/pull/1196 which require to reindex some facilitites. Let me know if you have an idea!

daniellekain:
@valencia-max 

"Just worry about cleaning open search index, but, as we need a full re-index here, I was thinking how we could take advantage of this process in this PR https://github.com/opensupplyhub/open-supply-hub/pull/1196 which require to reindex some facilitites. Let me know if you have an idea!"

That's a great idea! Your backfill only changes extended-field rows, and the incremental sync never re-reads those unless the facility record itself changes, so v1 would never pick up your corrections on its own. The full drop-and-rebuild we're already doing for OSDEV-1148 re-indexes everything from scratch, so it captures your fix for free. Just add your backfill to the post_deployment commands (which run before the rebuild in the pipeline) and one rebuild covers both PRs.

src/logstash/sql/sync_production_locations.sql
```diff
@@ -666,7 +666,64 @@ SELECT
...
+    SELECT
+      json_agg(DISTINCT jsonb_build_object(
+        'id', afl.id,
+        'contributor_id', as1.contributor_id,
+        'name', afl.name
+      ))::TEXT
+    FROM
+      api_facilitymatch afm
+      JOIN api_facilitylistitem afli2 ON afli2.id = afm.facility_list_item_id
+      JOIN api_source as1 ON as1.id = afli2.source_id
```
protsack-stephan, line 717: Those joins can be quite expensive, and I'm not entirely sure how this will affect replication latency. Since those are two of the largest tables in the database, can we benchmark this just in case? We could run it through the query planner with EXPLAIN, maybe even on a production connection, so we can be sure it's safe to deploy.

src/logstash/sql/sync_production_locations.sql
```diff
@@ -666,7 +666,64 @@ SELECT
...
+  -- only ACTIVE, PUBLIC, NON-ANONYMIZED sources with active matches are
+  -- publicly exposable. `NOT is_anonymized` keeps anonymized SLC sources
+  -- (public, but identity-masked) out of the contributor data.
+  (
+    SELECT
+      json_agg(DISTINCT jsonb_build_object('id', ac.id, 'type', ac.contrib_type))::TEXT
+    FROM
+      api_facilitymatch afm
+      JOIN api_facilitylistitem afli2 ON afli2.id = afm.facility_list_item_id
+      JOIN api_source as1 ON as1.id = afli2.source_id
```
protsack-stephan, line 682: Similar to the comment below, this query might be quite expensive. It might be worth ensuring this won't break anything.

protsack-stephan (commented):
Great work! I have a couple of comments. I also see that we are introducing filters for `contributor_type` and `contributor_id`. What about lists? Should we add some filters there as well?

vlad-shapik (dismissed):

src/logstash/sql/sync_production_locations.sql
daniellekain: Benchmarked on the anonymized prod-scale dump (2.5M facilities, 4.5M matches, 5.6M list items). All three subqueries plan as pure index scans — `api_facilitymatch`'s `facility_id` btree, then PK lookups on listitem/source/contributor/list — no seq scans anywhere. Worst-case facility in the dataset (775 active matches): 235ms cold for all three subqueries combined. Realistic batch of the 5,000 most-recently-updated facilities: 37.6s cold, 2.5s warm → ~0.5ms/facility steady-state on laptop Docker. Implications: incremental sync batches are negligible; a full rebuild adds on the order of 20–90 minutes across 2.5M facilities depending on cache, and follows the same access shape as the existing per-facility subqueries in this file. One caveat: the local dump predates migration 0225, so the `NOT is_anonymized` predicate was dropped from the bench — it is a boolean filter on already-fetched rows and does not change the plan. Happy to run the same EXPLAIN on a production connection before deploy for belt-and-braces.

src/logstash/sql/sync_production_locations.sql
daniellekain: Benchmarked together with the thread below — same answer applies: pure index scans, worst case 235ms, ~0.5ms/facility warm. Full numbers there.

doc/release/RELEASE-NOTES.md
daniellekain: Yes — exactly the case the fix covers. All three contributor subqueries carry `AND NOT as1.is_anonymized`, mirroring `should_display_associations` from `0225_index_contributors.sql`, so anonymized SLC sources (public but identity-masked) are excluded from `contributors`, `number_of_contributors`, and `lists` entirely. The release-notes entry spells this out too.

daniellekain:
Re: the lists-filter question — good call. The data is there (`lists` carries `id`, `contributor_id`, `name`), so a list-id filter is cheap to add later. We kept the query params to the RFC scope for this PR (`contributor_id`, `contributor_type`, `sort_by=number_of_contributors`); I'd put a lists filter in a follow-up ticket alongside the other v1 parity items rather than grow this change.

src/logstash/sql/sync_production_locations.sql
protsack-stephan: Thanks for checking!

src/logstash/sql/sync_production_locations.sql
protsack-stephan: Thanks for checking!

protsack-stephan (approved):
Great work!

PR #1156, opened 2026-07-28
[OSDEV-2980] Add duplicate check into the SLC submission process

jamesjohnson4395:
Contributors sometimes re-submit the same production location multiple times via SLC. We reject these duplicates during moderation, as the OpenSearch production-locations index takes around 15 minutes before including a just-approved submission in the list of potential matches. As we move to immediate auto-approval for SLC submissions, we want to add a check for duplicate submissions into the SLC form so that we do not auto-approve duplicate submissions and accidentally create multiple duplicate OSIDs. Actual duplicates should wait at least 30 minutes before re-submission — up to 15 minutes for the (externally hosted) auto-approval automation to approve the pending moderation event, plus up to another 15 minutes for the Logstash pipeline to reindex it.

A new DuplicateSubmissionProcessor in the location contribution pipeline queries ModerationEvent directly (not OpenSearch, which has the lag above) for the same contributor's non-rejected SLC/CREATE submissions from the last 30 minutes. A match requires exact country plus ≥0.9 name/address similarity (difflib.SequenceMatcher) — deliberately conservative on all three fields, since a single-field fuzzy match has caused false positives elsewhere in this codebase (ContributorManager.filter_by_name). Additionally, if numbers (i.e. street number, unit number, etc.) differ anywhere between submissions, they are not considered duplicates. Rejected submissions are excluded so resubmitting corrected data after a rejection isn't blocked.

On a match, POST /api/v1/production-locations/ returns 409 with a duplicate_of payload instead of creating a new moderation event; duplicate_override: true in the body bypasses it. The SLC form shows a "Possible Duplicate Submission" dialog on 409 and resubmits with the override flag if the contributor confirms.

**Known bug**: This change is currently scoped to CREATE SLC requests only, not PATCH/UPDATE. PATCH requests can sometimes still be approved as a new OSID by moderators, when the submitted patch data doesn't actually match its target OSID. But, cases where this happens alongside duplicate submissions should be few, so we can defer this edge case for now in favor of not increasing the risk of false positives.

No migrations or OpenSearch/Logstash changes; standard deploy.


src/django/api/moderation_event_actions/creation/location_contribution/processors/duplicate_submission_processor.py
```diff
@@ -0,0 +1,150 @@
...
+    Rejected submissions are excluded, since a rejected event won't become
+    a real facility and resubmitting after a rejection is legitimate.
+    '''
+
+    def process(
+            self,
+            event_dto: CreateModerationEventDTO) -> CreateModerationEventDTO:
+        if event_dto.request_type != ModerationEvent.RequestType.CREATE.value:
+            return super().process(event_dto)
+
```
VadimKovalenkoSNF, line 70: This processor is intended to apply only to SLC submissions, but it currently checks every CREATE request. Since **SourceProcessor** has already populated `event_dto.source`, could we also return early when the source is not `ModerationEvent.Source.SLC.value`? Otherwise an API submission can be rejected because it matches a recent SLC submission from the same contributor. Please add a regression test for an API CREATE following an SLC CREATE.

src/django/api/moderation_event_actions/creation/location_contribution/processors/duplicate_submission_processor.py
```diff
@@ -0,0 +1,150 @@
...
+        if event_dto.raw_data.get('duplicate_override'):
+            return super().process(event_dto)
+
```
VadimKovalenkoSNF, line 81: Could `duplicate_override` be moved out of raw_data and accepted as a validated query parameter instead? It is request-processing metadata rather than production-location data. Keeping it in the body causes it to be persisted in `ModerationEvent.raw_data` and requires special removal before calling ContriCleaner. I suggest validating `?duplicate_override=true`, adding a boolean field to CreateModerationEventDTO, and checking that field here. Strict validation is important because the current truthiness check allows values such as "false" or 1 to bypass duplicate detection.

src/django/api/moderation_event_actions/creation/location_contribution/processors/duplicate_submission_processor.py
```diff
@@ -0,0 +1,150 @@
...
+        return super().process(event_dto)
+
+    @staticmethod
+    def __find_recent_duplicate(event_dto: CreateModerationEventDTO):
+        cutoff = timezone.now() - timedelta(
+            minutes=DUPLICATE_CHECK_WINDOW_MINUTES
+        )
+        recent_events = ModerationEvent.objects.filter(
+            contributor=event_dto.contributor,
```
VadimKovalenkoSNF, line 149: There is a race between this lookup and creation of the moderation event: two concurrent identical requests can both find no recent event and then both create one. `transaction.atomic` alone does not serialize that sequence. Could we add contributor-scoped locking or another serialization mechanism around duplicate-check-plus-create? Locking only matching moderation events would not cover the case where no event exists yet.

src/react/src/actions/contributeProductionLocation.js
```diff
@@ -86,16 +92,23 @@ export function createProductionLocation(contribData) {
     };
 }
 
-export function updateProductionLocation(contribData, osID) {
+export function updateProductionLocation(
+    contribData,
+    osID,
+    duplicateOverride = false,
```
VadimKovalenkoSNF: The backend explicitly excludes UPDATE requests from duplicate detection, so passing `duplicateOverride` through `updateProductionLocation` appears to be dead behavior. Could we remove the override parameter and body mutation from the PATCH path to keep the contract scoped to CREATE?

src/react/src/components/Contribute/ProductionLocationInfo.jsx
```diff
@@ -1214,6 +1227,16 @@ const ProductionLocationInfo = ({
...
             />
+            <ContributionWarningDialog
+                open={showDuplicateSubmissionDialog}
+                onClose={() => setShowDuplicateSubmissionDialog(false)}
+                onSubmitAnyway={() => {
+                    setShowDuplicateSubmissionDialog(false);
+                    handleProductionLocation(contributionForm.values, true);
+                }}
+                title="Possible Duplicate Submission"
+                message="You recently submitted a very similar production location. Please wait at least 30 minutes before re-submitting information for the same production location, as doing so could create unwanted duplicates. If this is a different, new location, go back and double-check the name, address, and country. Otherwise, click 'Submit anyway' to confirm this submission."
```
VadimKovalenkoSNF: The message hardcodes 30 minutes separately from `DUPLICATE_CHECK_WINDOW_MINUTES`. Could the backend include the wait duration in the duplicate response so the UI renders the authoritative value? Otherwise these can silently diverge.

VadimKovalenkoSNF (dismissed):
Hi, I left a few comments using Cursor assistance. Please, take a look.

daniellekain:
One ask from the monitoring side: could we add two log lines so we can measure this feature once it ships? Neither side of the flow leaves a database trace — a blocked duplicate creates no ModerationEvent, and with `duplicate_override` now a query parameter (good change) the override isn't recorded on the event either — so logs are the only signal for how often the check fires and how users respond.

```python
import logging

logger = logging.getLogger(__name__)

# where the 409 is set:
logger.info(
    'Duplicate SLC submission flagged: contributor=%s duplicate_of=%s',
    event_dto.contributor.id,
    duplicate.uuid,
)

# where duplicate_override short-circuits the check:
logger.info(
    'Duplicate check bypassed via duplicate_override: contributor=%s',
    event_dto.contributor.id,
)
```

With both lines, CloudWatch Logs Insights gives us the full funnel — warnings shown vs. overridden vs. abandoned (users who went back and edited) — and distinct contributors affected. Structured application logs also stay reliable if access-log formats change or another 409 is ever added to this endpoint.

I want to make sure we can easily track (and broadcast!) the impact of this change :).

daniellekain:
The lock placement works and is nicely simple — one check, under the lock, and the second racing request never even reaches the geocoder. Two hardening suggestions:

1. **Make the atomic dependency self-enforcing.** As the docstring notes, this relies on `@transaction.atomic` on `ProductionLocations.create`. If that decorator is ever removed — or a future path (e.g. the SLC automation work) calls `ModerationEventCreator` for SLC CREATEs outside a transaction — `pg_advisory_xact_lock` silently becomes a no-op: autocommit releases it at statement end, no error, and the tests stay green since they mock the connection. One guard line closes that trap:

   ```python
   if not connection.in_atomic_block:
       raise RuntimeError(
           'The duplicate-check advisory lock requires an open '
           'transaction (transaction.atomic) to have any effect.'
       )
   ```

2. **Namespace the lock key.** `pg_advisory_xact_lock(contributor.pk)` uses the single-key form, so any future advisory lock elsewhere in the app keyed on a bare integer pk could collide with contributor ids. The two-key form is free insurance: `pg_advisory_xact_lock(2980, contributor.pk)` (any constant unique to this feature). No other advisory locks exist in the codebase today, so this is future-proofing, not a live bug.

Separately: nice addition with the geocoder timeout. Note that it also applies to list uploads and claims (both previously unbounded), since `geocode_address` is shared — a net improvement, but worth a mention in the release notes bullet as a deliberate cross-cutting change.

daniellekain (dismissed):
Approving as my remaining comments are non-blocking. Happy to re-review/approve as needed!

daniellekain (dismissed):
Amazing work!

ColumbusLeonardOSHub (approved):

daniellekain (approved):

PR #1158, opened 2026-07-28
[OSDEV-2865] Show specific API error when deleting a claimed facility

sidd-OSH:
Display the specific API error message on the Delete Facility dashboard page when deletion fails.

Previously, any delete failure showed the generic "An error prevented deleting that facility". When a data moderator tries to delete a facility with an approved claim, the backend returns `{"detail":"Facilities with approved claims cannot be deleted"}` but that detail was not shown on the frontend and replaced with the hardcoded string.

The error detail was already being extracted, the component just wasn't rendering it. This change renders the actual error array instead of the hardcoded fallback.

https://opensupplyhub.atlassian.net/browse/OSDEV-2865 

Screenshot 2026-07-26 at 8 14 24 AM

Screenshot 2026-07-26 at 8 20 45 AM


src/react/src/components/DashboardDeleteFacilityControls.jsx
```diff
@@ -76,7 +76,7 @@ export default function DashboardDeleteFacilityControls({
             {deleting && <CircularProgress />}
             {error && (
                 <span style={{ color: 'red' }}>
-                    An error prevented deleting that facility
+                    {error.join(' ')}
```
valencia-max: Nitpick: Instead of joining this using space, could be ", "

daniellekain (commented):
great work! Would you mind including a before/after screenshot?

valencia-max (dismissed):
Thanks for fixing this! 🤖

VadimKovalenkoSNF (changes requested):
Could you please add a test file for `DashboardDeleteFacilityControls.jsx` and verify this scenario there as well? We keep FE tests in `src/react/src/__tests__/components/` folder.

valencia-max (approved):

daniellekain (approved):

vlad-shapik (approved):
LGTM

VadimKovalenkoSNF (approved):

PR #1159, opened 2026-07-29
[OSDEV-2384] Fix empty mixed-case fields in embedded-map downloads

VadimKovalenkoSNF:
Resolve mixed-case LIST custom fields in embedded-map CSV/XLSX downloads.

## Summary

Fix [OSDEV-2384](https://opensupplyhub.atlassian.net/browse/OSDEV-2384): configured embed fields such as `program_1_Name` were empty in `GET /api/facilities-downloads/` exports even when the embedded map showed values.

LIST uploads were looked up through `get_raw_json()`, which lowercases header names at runtime. Embed config keeps the original case, so mixed-case columns never matched.

- Resolve LIST custom fields by original header position: prefer an exact-case match, then fall back to case-insensitive matching.
- Cache per-header column indexes for the duration of a download.
- Leave standard extended-field columns, ownership scoping, and value joining unchanged.

## Test plan

- [x] `docker exec open-supply-hub-django-1 sh -c "python3 manage.py test api.tests.test_facility_download_serializer_embed_mode"`
- [x] `docker exec open-supply-hub-django-1 sh -c "python3 manage.py test api.tests.test_facility_download.FacilityDownloadTest"`
- [x] Flake8 on modified Python files
- [ ] Verify the RISE embed download on Test with contributor `7715` (e.g. facility `BD2021337FH4VRF`) — `program_1_Name` populated where the map shows a value

[OSDEV-2384]: https://opensupplyhub.atlassian.net/browse/OSDEV-2384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/serializers/facility/facility_download_serializer_embed_mode.py
```diff
@@ -36,7 +37,7 @@ def get_row(self, facility: FacilityIndexNewManager) -> List[str]:
         return [
             *self.get_common_row(facility),
             *self.get_contributor_custom_fields(facility),
-            *self.get_extended_fields(self.get_extended_fields_raw(facility)),
+            *self.get_extended_fields(facility.extended_fields),
```
valencia-max: Just for curiosity:
Here you're passing the entire list of `extended_fields` associated with the `facility` as a parameter. That list may include `extended_fields` from other contributors, so its size will be larger than the one generated by `self.get_extended_fields_raw`, since that function only returned the fields belonging to the current contributor.

Besides fixing the bug, is this behaviour intentional? Do we want to download these values (in embed mode) even if they are not direct contributions from the current contributor?

valencia-max (dismissed):
LGTM!

src/django/api/serializers/facility/facility_download_serializer_embed_mode.py
vlad-shapik: @VadimKovalenkoSNF 

>  That list may include extended_fields from other contributors, so its size will be larger than the one generated by self.get_extended_fields_raw

I believe it is a valid point: they must only be able to download from the embed map the values they see, since we intentionally don't show contributions from other contributors on the embedded map in order to promote only the owners' data.

VadimKovalenkoSNF: Thank you for flagging this!
The review uncovered several issues requiring further work:

1. Embed downloads may include data from other contributors.
2. Multiple values may be exported although embed mode displays only one.
3. Hidden embed fields may appear in downloads.
4. Sector exports aggregate all contributors instead of matching the embed view.
5. Sharing selection logic with the facility API may change its existing response contract and introduce regressions.

I’m moving the ticket back to In Progress to align download behavior with the embed UI while preserving existing API behavior.

valencia-max: I don't have any problem changing the current behavior, I mean, if we need data from other contributors to not download empty rows it's fine, maybe we could get some input from product to address this with no more changes. Let me know what you think @VadimKovalenkoSNF @vlad-shapik

src/django/api/serializers/facility/facility_download_serializer_embed_mode.py
```diff
@@ -68,19 +81,40 @@ def get_extended_fields_raw(self, facility: FacilityIndexNewManager):
...
-            if field["column_name"]
+            column_name
+            for column_name in EmbedField.objects.filter(
+                embed_config_id=contributor.embed_config_id,
+                visible=True,
+            ).order_by("order").values_list("column_name", flat=True)
+            if column_name
         ]
+
+    def _get_list_field_indexes(self, list_header: str):
```
VadimKovalenkoSNF, line 102: LIST custom fields are resolved against the uploaded list header by position: exact name match **first**, then a case-insensitive (casefold) **fallback**. 

The download column name always comes from the embed config; **only the value lookup is case-tolerant**. 

Example: embed config `program_1_name` + list header `program_1_Name` → column `program_1_name` is populated from that list cell (and vice versa). If both an exact and a differently-cased header exist, the exact match wins.

src/django/api/serializers/facility/facility_download_serializer_embed_mode.py
VadimKovalenkoSNF: I'd rather not expand scope here. This ticket is specifically about empty custom fields caused by mixed-case LIST header matching (e.g. `program_1_Name`). Including other contributors’ extended-field data is a separate product decision about what embed downloads should contain, so I’d keep that in a follow-up ticket if product wants it. Happy to open one once we have that input.

vladsha-dev (approved):

src/django/api/management/commands/post_deployment.py
```diff
@@ -37,20 +37,3 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         call_command('migrate')
-        # Temporary for 2.28.0 (OSDEV-2949) — one-time cleanup that strips the
-        # nested 'internal_ID' (not part of the partner field JSON Schema) from
-        # 'rsc_grievance_mechanism' values. The ExtendedField database trigger
-        # refreshes the affected FacilityIndex rows. Remove after the release
```
vlad-shapik, line 43: Could you remove the temp code above as well? Or maybe move it into a utils file inside the relevant reindexing module.

vlad-shapik (dismissed):
LGTM

I've left optional comment.

src/django/api/management/commands/post_deployment.py
VadimKovalenkoSNF: I've moved this to the `src/django/api/facility_index_backfill/utils.py`. Also, I updated `README.md`, pls check.

valencia-max (approved):

VadimKovalenkoSNF:
@coderabbitai resolve

vlad-shapik (approved):

PR #1165, opened 2026-07-30
[OSDEV-2980] Sample: close duplicate-check race with per-contributor advisory lock

daniellekain:
Sample implementation for the concurrency discussion on #1156 (Vadim's race-condition comment) — offered for cherry-picking or as a review aid, not as a competing PR. Base is the #1156 branch, so the diff shows only the delta on top of James's work.

## What

Two concurrent identical submissions can both pass the duplicate check, because neither moderation event exists yet when the other one looks, and both then create events. This closes that race with double-checked locking:

- **`duplicate_check.py` (new)**: the detection logic extracted verbatim from `DuplicateSubmissionProcessor` into a shared module (`applies_to` / `find_recent_duplicate` / `set_duplicate_error`), so the two call sites below cannot drift apart.
- **`DuplicateSubmissionProcessor`** keeps the early check — fast user-facing 409 before the geocoder call — now documented as advisory-only.
- **`ModerationEventCreator`** re-checks authoritatively inside `transaction.atomic()` while holding `pg_advisory_xact_lock(2980, contributor_id)`, immediately before the `ModerationEvent` insert. The second of two racing requests waits a few milliseconds for the lock, then its re-check sees the first request's committed row and returns the same 409 payload. The lock is per-contributor (no cross-contributor contention), held only for one indexed query + one insert (no external calls inside the critical section), and self-releases at transaction end.

The shared `applies_to` gate also checks `event_dto.source == SLC` — which doubles as the fix for the other review thread (API-source CREATEs being rejected against recent SLC submissions), since the incoming-side gate now mirrors the candidate query's `source=SLC` filter.

Deliberately **not** addressed here: the `duplicate_override` validation question (truthiness vs strict boolean / body vs query param) — `applies_to` centralizes that check so whatever is decided lands in one place.

## Tests

- `test_creator_recheck_catches_duplicate_missed_by_processor` — simulates the race by bypassing the processor-level check, asserting the creator-level re-check returns the 409 with the correct `duplicate_of` payload.
- `test_api_source_submission_is_not_checked_against_slc_submission` — regression test for the source scoping.
- Full affected suites green: `test_duplicate_submission_processor` (18) + `test_location_contribution_strategy` + `test_partner_data_file_upload` (49) via docker compose; flake8 clean.

A true-concurrency test (threads + `TransactionTestCase`) was deliberately skipped as flake-prone; the race path is instead covered deterministically by the bypass test above.

### DIFF
--- a/src/django/api/moderation_event_actions/creation/location_contribution/duplicate_check.py
+++ b/src/django/api/moderation_event_actions/creation/location_contribution/duplicate_check.py
@@ -1,0 +1,151 @@
+'''
+Shared duplicate-detection logic for SLC CREATE submissions.
+
+Used by two call sites that must never drift apart:
+- DuplicateSubmissionProcessor runs the fast, user-facing check early in
+  the contribution chain, before the geocoding call, so a duplicate fails
+  quickly and cheaply. That check is advisory only: two concurrent
+  identical requests can both pass it, because neither event exists yet
+  when the other one looks.
+- ModerationEventCreator runs the authoritative re-check under a
+  per-contributor advisory lock immediately before inserting the
+  ModerationEvent row, which closes that race.
+'''
+
+import re
+from datetime import timedelta
+from difflib import SequenceMatcher
+
+from django.utils import timezone
+from rest_framework import status
+
+from api.constants import APIV1LocationContributionErrorMessages
+from api.models.moderation_event import ModerationEvent
+
+# How far back to look for a recent submission by the same contributor.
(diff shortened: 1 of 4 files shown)


daniellekain:
Superseded — #1156 now closes the race with an in-processor advisory lock held for the request transaction, which is simpler than the double-check shape sampled here (single check point, and the second racing request never reaches the geocoder). The SLC source-gate part also landed in #1156 directly. Closing.

PR #1014, opened 2026-05-16
[OSDEV-2724] Fix facility list table header displaying row number instead of blank

daniellekain:
## Summary

The header row was assigned `rowIndex="1"`, causing it to display a row number. Data rows were consequently numbered starting from 2. The header should show no row number and data rows should start at 1. The total row count displayed on the page was always correct and is unaffected by this change.

## Before

Screenshot 2026-05-15 at 9 09 36 PM

## After

Screenshot 2026-05-15 at 9 08 05 PM

## Test plan

- [x] Navigate to `/dashboard/lists` and open a facility list
- [x] Confirm the header row has a blank first column (no row number)
- [x] Confirm data rows are numbered starting from 1

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1167, opened 2026-07-31
[OSDEV-2724] Restore uploaded-file row numbering on facility list pages

daniellekain:
Revert the OSDEV-2724 display changes so facility list pages number rows exactly as the contributor's uploaded file does: the header row shows "1" and `uploadedFileRowIndexOffset` returns to 2.

## Why

The platform stores `row_index` 0-based, while the uploaded file counts the header as row 1 and the first data row as row 2. The display offset of 2 existed precisely to bridge that — its comment read "necessary to match row indices in the uploaded files."

#1014 changed the page to a prettier scheme (blank header, data rows 1..N). But row numbers on this page aren't cosmetic: ContriBot validation reports and the data team's contributor feedback emails cite *file* row numbers, and moderators cross-reference those against this page when removing malformed rows before approving a list. With the shifted numbering, every report and email pointed one row away from what the page showed — surfaced while reconciling a moderation list against its ContriBot report.

## What

- Header row displays "1" again, mirroring the file's header row (and how ContriBot counts).
- `uploadedFileRowIndexOffset` back to 2, so data rows display 2..N+1 — matching the file exactly.
- The constant now documents the invariant: the list page and ContriBot number rows the same way, and any renumbering must change both together or not at all.

## Notes for review

- This intentionally restores the pre-#1014 look. The original cosmetic complaint (a list "starting at 2" looks odd) is legitimate, but the fix is a coordinated renumbering of the list page **and** ContriBot together — deferred to a follow-up; changing either side alone recreates this off-by-one.
- Frontend display only; `row_index` storage untouched. No tests reference the constant.
- Release notes updated (2.28.0, Bugfix).


VadimKovalenkoSNF (approved):
Approved.

valencia-max (approved):
LGTM!

PR #1170, opened 2026-08-03
[OSDEV-3170] Mark claimant name, address and sector contributions as claim data

daniellekain:
## What

Extend the "Claimed" data attribution to **name, address and sector**, so an
approved claimant's own contributions are labeled as claim data wherever they
appear on the production location page.

## Why

[OSDEV-3093](https://opensupplyhub.atlassian.net/browse/OSDEV-3093) (2.28)
taught `is_from_claim` to recognize claimant data submitted through any
channel — but only for values served by the extended-field serializer
(workers, parent company, product/location types, native-language name).
Three field types take different paths and were missed:

- **name / address** — contributed values are appended into the response from
  `FacilityIndex.facility_names` / `facility_addresses` by
  `create_name_field_from_facility_name` /
  `create_address_field_from_facility_address`, which never set the flag.
- **sector** — served by `get_sector` → `format_sectors`, which sets the flag
  purely by whether an entry came from the claim details.

The visible result is a self-contradicting row: the claimant's own submitted
name, credited to the claimant's own organization, chipped "Crowdsourced".
Reproduced on Test after the 2.28 deploy — e.g.
[US2024279C8R9VJ](https://test.os-hub.net/production-locations/US2024279C8R9VJ)
(A.V. Thomas Produce), where Number of Workers and Parent Company correctly
read "Claimed" while Name and Industry/Sectors do not.

Note the gap exists *because* name/address were removed from the claims form:
claimants are pointed at SLC instead, and the chip judged those two fields by
channel alone. The [Promotion Logic Q3 Report](https://opensupplyhub.atlassian.net/wiki/spaces/SD/pages/1363410945/Promotion+Logic+Q3+Report+Options+Recommendation)
scopes the desired behavior as "any data point contributed by the approved
claimant shows as Claimed, regardless of channel".

## How

All three call sites now resolve `approved_claim.contributor_id` and mark an
entry via a shared helper, `is_contribution_from_claimant`, which keeps
OSDEV-3093's guard: **a masked (anonymized) contribution is never attributed
to the claim**, since the claimant is publicly named and the badge would undo
the masking by inference.

No frontend change is needed — `formatExtendedField` already maps
`is_from_claim` through, and the sector row already reads it.

## Labeling only — what does and does not move

This was scoped deliberately after tracing the ordering effects, since
`is_from_claim` also participates in sorting (findings are on the ticket):

- **No displayed value changes anywhere.** The name and address rows render
  the canonical `properties.name` / `properties.address`, computed by the
  index promotion SQL, which this PR does not touch.
- Marking a claimant's name/address entry **does** raise it within its list —
  that is precisely what makes the row credit the claimant instead of another
  contributor who submitted the same value. `verified_count` still outranks
  the claim flag, so moderator-verified values keep winning.
- **The sector list order is untouched** (`claim_sectors + item_sectors` is
  unchanged): entries are marked in place, so the displayed sector never
  moves.

Out of scope, tracked under OSDEV-392: promoting a claimant's value above
other contributors'. Concretely, on a facility where a non-claimant
contributed a sector *more recently* than the claimant, that contributor's
sector stays first and still reads "Crowdsourced" — the test below pins this
so it can't be "fixed" by accident. Also out of scope: v1 API
([OSDEV-3115](https://opensupplyhub.atlassian.net/browse/OSDEV-3115)) and
downloads ([OSDEV-3116](https://opensupplyhub.atlassian.net/browse/OSDEV-3116)).

## API surfaces affected

**Legacy `/api/facilities/`** — both `retrieve` and `list` expose
`extended_fields` and `sector`, so responses change:

- `extended_fields.name` entries gain an `is_from_claim` key that was not
  previously present (address and sector entries already carried it; only its
  value can now be `true`). Additive, so no consumer breaks.
- `extended_fields.name` / `address` arrays can re-order: a claimant entry
  rises above a `created_from` entry holding the same value, so a consumer
  reading `[0]` may see a different contributor credited for an identical
  value.
- `sector` array order is unchanged; only flag values differ.

**Caching — expect a lag after deploy.** `list` is wrapped in
`cache_view_response('facilities_list')` and `retrieve` additionally in
`cache_page_by_auth_tier(MEMCACHED_VIEW_CACHE_TIMEOUT_SECONDS, ...)`
(10 minutes by default). Cache keys vary on the visibility token, not on
claim state, so warm entries keep serving the old attribution until they age
out. If QA checks a profile immediately after deploy and still sees
"Crowdsourced", suspect a warm cache before suspecting the fix.

**v1 `/api/v1/production-locations/` — unaffected.** It is served from
OpenSearch via `sync_production_locations.sql`, which does not use these
serializers and has no notion of the approved claimant's contributor. Until
[OSDEV-3115](https://opensupplyhub.atlassian.net/browse/OSDEV-3115) lands, v1
and the production location page will disagree about attribution for
claimant-contributed name/address/sector. Downloads likewise
([OSDEV-3116](https://opensupplyhub.atlassian.net/browse/OSDEV-3116)).

## Testing

`api.tests.test_facility_index_details_serializer` — 36 tests pass, including
three new ones:

- claimant name and address contributions are marked, other contributors' are not;
- the claimant's sector is marked **without reordering** — the fixture's newer
  non-claimant sector stays first (asserted explicitly, with a guard on the
  fixture's premise);
- the helper's guards: masked → not claim data, wrong contributor → not claim
  data, no claim → not claim data.

`api.tests.test_facility_index_extended_field_list_serializer` and
`api.tests.test_sector_api` also pass; `flake8` clean.

## Release notes

Added under **2.28** to try to land before the code freeze — if this misses
it, the entry needs to move to a new 2.29 section.

Ticket: [OSDEV-3170](https://opensupplyhub.atlassian.net/browse/OSDEV-3170)

[OSDEV-3093]: https://opensupplyhub.atlassian.net/browse/OSDEV-3093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OSDEV-3115]: https://opensupplyhub.atlassian.net/browse/OSDEV-3115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/serializers/facility/utils.py
```diff
@@ -173,14 +173,38 @@ def get_efs_associated_with_contributor(
...
+    A masked (anonymized) contribution is never attributed to the claim: the
+    claimant is publicly named, so the attribution would undo the masking by
+    inference.
+    """
+    if claimant_contributor_id is None:
+        return False
+
+    contributor = contributor or {}
+
+    if is_contribution_masked(contributor, masked_ids):
```
valencia-max: If the contribution is by an anonymized contributor and it's from claimant, I think we should keep it as contribution from claimant, just hidding the contribution name. Let me know what you think

daniellekain:
Agreed — done in b31e764. And you've found something wider than this PR: the same guard exists in OSDEV-3093, which is where I copied it from, so I've corrected both.

Two things convinced me:

- `get_claim_info` masks the claimant's name with the same policy, so when the claimant is a masked contributor no name appears anywhere on the profile — including the claimed banner. The label therefore discloses no identity; it only says the (unnamed) claimant supplied the value.
- `_get_is_from_claim`'s first branch already returns `True` for claim-form fields *before* any masking check, so a masked claimant's claim-form data is labelled "Claimed" today. Keeping the guard meant the same claimant's data was labelled two different ways depending on which channel it arrived through — exactly the inconsistency this ticket exists to remove.

So `is_contribution_from_claimant` no longer takes the masked set, and `_get_is_from_claim` now delegates to it — extended fields, name, address and sector all decide this the same way. Identity hiding is untouched: names and ids still go through `get_contributor_name_from_facilityindex` / `get_contributor_id_from_facilityindex` with the masked set, and the updated test asserts both at once (attribution kept, `contributor_name` masked, `contributor_id` null).

77 tests pass across the details, extended-field and masking-policy suites; flake8 clean. Both release-note entries updated, since this changes 3093's shipped behaviour too — worth a look in case you'd rather that part rode its own ticket.

src/django/api/serializers/facility/utils.py
daniellekain: fixed!

daniellekain:
Retargeting this at **2.29** — it won't make today's 2.28 freeze, so the release note has moved to a new 2.29 section (752d66f).

One consequence worth flagging for QA rather than leaving implicit: OSDEV-3093 now reaches production in 2.28 **with its masked-contribution guard intact**, so I restored the sentence describing that to the 2.28 entry. It had been removed on the assumption both changes would ship together. So in 2.28, a masked claimant's SLC/list contributions stay labelled "Crowdsourced" while their claim-form data is labelled "Claimed" — the inconsistency this PR removes, now shipping for one release.

Nothing else changes: CI is green, and Vadim's review comment is addressed.

valencia-max (commented):
LGTM! 🤖

valencia-max (dismissed):

valencia-max (dismissed):

valencia-max (approved):

PR #1164, opened 2026-07-30
[OSDEV-992] Align sector search index with profile display

daniellekain:
## What
[OSDEV-992]
Fix the `/api/facilities/` **Sector** filter so it only returns facilities whose location profile displays the searched sector. Concretely, correct how the searchable `FacilityIndex.sector` column is populated.

## Why

Filtering by a sector such as **Toys** returned facilities that show no Toys anywhere on their profile (ticket repro: CN2021309BGQF7A; also the Electronics example in the comments).

Root cause: the old `index_sector()` aggregated the sectors of **every** MATCHED/CONFIRMED_MATCH list item a facility ever received. The profile, however, displays only **one list item per contributor** (`regroup_items_for_sector_field` in `api/serializers/facility/utils.py`) — preferring items from active sources, then the most recently updated — plus approved claim sectors. So when a contributor's newer upload changed a facility's sectors, the superseded values stayed searchable indefinitely. Same display-vs-search mismatch class as OSDEV-1034 (#1151).

## How

- **`src/django/sqls/0221_index_sector.sql`** — new `index_sector()` that selects, per contributor, the single list item the profile displays (`DISTINCT ON` with the display's exact ordering: active source first, then facility-match activity, then newest `updated_at`), plus the latest approved claim's sectors per contributor.
- **`0221_fix_sector_search_index.py`** — migration applying the corrected function (reverts to `0130` on rollback).
- **`facility_index_backfill/specs.py`** — new `sector` backfill field group; its filter targets only facilities where a single contributor has more than one matched list item (the only rows whose indexed value can change).
- **`post_deployment.py`** — the temporary 2.28.0 backfill now runs `processing_type,sector` (flagged for removal after the release rolls out everywhere).
- **Tests** — regression tests for superseded-sector exclusion, latest-sector inclusion, multi-contributor union, active-source preference, and claim sectors; backfill spec tests.
- **Release notes** — added to the 2.28.0 entry.

## Display quirks mirrored deliberately

- The display sort orders by `facilitymatch.is_active` **ascending** (`utils.py:235` — the `not` present on the source line is absent on the match line), i.e. it prefers items whose match is inactive. The new SQL mirrors this bug-for-bug so search always equals the profile as it behaves today. If that display behavior is ever deemed a bug and fixed, `index_sector()` must be updated in the same change.
[...]

PR #1172, opened 2026-08-03
[OSDEV-992] Add a parity guard between the sector search index and the profile

daniellekain:
## What

One test: assert the searchable `FacilityIndex.sector` column and the sector values the profile displays are the same set.

## Why

[OSDEV-992](https://opensupplyhub.atlassian.net/browse/OSDEV-992) was a divergence between two independent implementations of the same selection rules — `index_sector()` in SQL builds the search index, `get_sector()` in Python builds the profile view. #1164 realigned them, but nothing prevents them drifting apart again, and a change to the display path alone would reintroduce the exact reported symptom: facilities matching a "Sector" search that don't show that sector on their profile.

This test was written while working #1164 and never landed. Rebased onto current main, fixed (it read `data["sector"]` where the GeoJSON serializer puts it under `data["properties"]["sector"]`), and verified.

## How

Builds one facility that exercises every selection rule at once — a superseded item from the same source, a second contributor, an inactive source, and an approved claim with its own sector — then compares the two implementations' output as sets. Fails if either side changes without the other.

## Testing

`api.tests.test_sector_api` — 8 tests pass (7 existing, 1 new). Confirmed the new test genuinely exercises the path: it errored before the `properties` fix rather than passing vacuously.

Ticket: [OSDEV-992](https://opensupplyhub.atlassian.net/browse/OSDEV-992) (follow-up to #1164, no behavior change — test only)

[OSDEV-992]: https://opensupplyhub.atlassian.net/browse/OSDEV-992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OSDEV-992]: https://opensupplyhub.atlassian.net/browse/OSDEV-992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/django/api/tests/test_sector_api.py
+++ b/src/django/api/tests/test_sector_api.py
@@ -12,6 +12,9 @@ from api.models import (
     User,
 )
 from api.models.facility.facility_index import FacilityIndex
+from api.serializers.facility.facility_index_details_serializer import (
+    FacilityIndexDetailsSerializer,
+)
 from api.tests.facility_api_test_case_base import FacilityAPITestCaseBase
 
 from django.contrib.gis.geos import Point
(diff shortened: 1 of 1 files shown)


vlad-shapik (approved):
LGTM

PR #1176, opened 2026-08-04
[OSDEV-3176] Treat a blank claim form value and a NULL column as unchanged

daniellekain:
## What

Make the no-op detection in `PUT /api/facility-claims/{id}/claimed/` treat a NULL column and an empty string as the same value for text fields, so submitting a blank field that was already empty is correctly a no-op.

## Why

Found in QA of [OSDEV-3094](https://opensupplyhub.atlassian.net/browse/OSDEV-3094) (test case OSDEV-3148, preprod claim 5864): the first form save appeared to change something, and only the second identical save was a no-op.

The cause is a round trip the frontend does deliberately. `fetchClaimedFacilityDetails` (`src/react/src/actions/claimedFacilityDetails.js`) converts every null in the GET response to `''` when loading the form, and `submitClaimedFacilityDetailsUpdate` PUTs that same object back. So the form receives nulls and returns empty strings. OSDEV-3094 compared values exactly, so `'' != None` counted as a change: the claim saved, `updated_at` bumped, the claim reindex trigger fired, extended fields rebuilt, and list contributors were emailed — with nothing visible altered.

That matters because `updated_at` feeds the claimed section's "last updated" date (OSDEV-2679), which is precisely the signal OSDEV-3094 set out to make trustworthy. And it is not an edge case: **2,999 of 3,586 approved claims (84%) hold at least one NULL** across the 14 tracked text fields — 30,604 field-level NULLs, averaging 10.2 per affected claim — so nearly every claimant would have seen one unexplained bump on their first save after 2.28.

Fixing it server-side rather than in the form also covers API callers, and does not depend on the frontend keeping its current habits.

## How

- `get_tracked_claim_value` canonicalizes NULL to `''` after the existing `to_python` normalization, for `CharField`/`TextField` and their subclasses only.
- Keyed on **model field type, not a name list**, so fields added to the tracked groups are covered automatically.
- Only the comparison changes. The apply loops still write whatever the request contains, so a request that changes something else persists `''` exactly as today.
- The four `*_publicly_visible` booleans are untouched: they are `null=False, default=False` and NOT NULL in the database, so they can never hold NULL.

### Considered and rejected: a NULL → `''` backfill

Tidier data, but it would touch 2,999 rows and fire `facility_claim_post_update_insert_indexing_trigger` (`AFTER UPDATE ... FOR EACH ROW`, `sqls/0170_create_table_triggers.sql`) once per row — the same index work as 3,000 claim edits arriving at once — while leaving no simple-history record of the change. It also would not prevent the same class of bug for any nullable field added later, whereas normalizing in the comparison does.

## Testing

- New: a blank-over-NULL submit is a no-op — no save, `updated_at` unchanged, no email, and the columns keep their NULLs.
- New guard: clearing a field that *does* hold a value still saves, so the equivalence cannot grow into a swallowed edit.
- `api.tests.test_approved_facility_claim` (16 tests) and `api.tests.test_facility_claim_view_set` + `api.tests.test_facility_claim` (18 tests) pass; flake8 clean.

## Found alongside, filed separately

[OSDEV-3177](https://opensupplyhub.atlassian.net/browse/OSDEV-3177) — the same handler returns **500** when a PUT omits any of the four `*_publicly_visible` flags: an absent key becomes `None`, which the NOT NULL column rejects. Pre-existing (it predates OSDEV-3094), unreachable from the web form, and reachable by any API caller attempting a partial update. Not fixed here to keep this PR to one concern.

## Notes for review

- Targets **2.29** and creates the `## Release 2.29.0` section. PR #1170 creates the same section, so whichever merges second will take a small conflict there.

Ticket: [OSDEV-3176](https://opensupplyhub.atlassian.net/browse/OSDEV-3176)

[OSDEV-3094]: https://opensupplyhub.atlassian.net/browse/OSDEV-3094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OSDEV-3177]: https://opensupplyhub.atlassian.net/browse/OSDEV-3177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


madm2n (approved):

protsack-stephan (dismissed):

doc/release/RELEASE-NOTES.md
```diff
@@ -7,6 +7,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
 * Release date: *Provide release date*
```
vlad-shapik: Could you please format it properly? This line should be after `* Product name: Open Supply Hub`

vlad-shapik (changes requested):

doc/release/RELEASE-NOTES.md
daniellekain: Good catch, that was a bad auto-merge. Rebuilt the section in template order.

vlad-shapik (dismissed):
LGTM

protsack-stephan (approved):

vlad-shapik (approved):

PR #1180, opened 2026-08-06
[OSDEV-3185] Implement ContriBot process_list Lambda

protsack-stephan:
## Jira Ticket

[OSDEV-3185](https://opensupplyhub.atlassian.net/browse/OSDEV-3185)

---

## Summary of Changes

Implement the ContriBot `process_list` Step Functions Map task so each enqueued facility list is validated and its annotated report is uploaded for moderator review.

`fetch_lists` (OSDEV-2928, already on `main`) enqueues work for the Map state, but `process_list` was still a stub. Without this step there is no validation report or payload for the `notify` Lambda.

**`process_list/handler.py`** — For each Map item, loads the DynamoDB row, marks it `PROCESSING`, downloads the `.xlsx` from S3 using `file_name` as the object key, runs `ContriBot` with the bundled `lib/0000.error_codes.xlsx`, uploads the `.~PROCESSED.` workbook to Google Drive, stores `report_url` and summary stats on the row, and returns `{list_id, report_url, num_lines, num_errors, error_ratio}` for `notify`. Failures raise so Step Functions `Catch` can attach `$.error` and still invoke notify.

**Supporting libraries** — Add `lib/s3_storage.py` and `lib/google_drive.py` (Drive upload via service-account secret from Secrets Manager). Extend `ListsRepository` with `get_list`, `update_list`, and processing status constants.

**`fetch_lists` follow-up** — Filter the OS Hub API query to `status=PENDING` via a new `FacilityListStatus` enum so only lists ready for ContriBot are enqueued.

**Packaging & infra** — Update the `process_list` Makefile to cross-compile deps for Lambda Python 3.10 (avoids host-Python wheel mismatches), strip runtime-provided boto3 and non-Drive discovery docs (keep `drive.v3.json`), and bundle `lib/` plus the error-codes workbook. Scope ContriBot Lambda environment variables per function instead of sharing one block. Raise `process_list` memory to 1024 MB.

---

## Type of Change

- [ ] UI / frontend change

[OSDEV-3185]: https://opensupplyhub.atlassian.net/browse/OSDEV-3185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vlad-shapik (dismissed):
LGTM

vlad-shapik (dismissed):

src/contribot/process_list/handler.py
```diff
@@ -1,7 +1,105 @@
...
+    if not file_name.lower().endswith(".xlsx"):
+        raise ValueError(f"ContriBot requires an .xlsx workbook; got {file_name!r}")
+
+    work_dir = Path("/tmp") / "contribot" / list_id
+
+    if work_dir.exists():
+        shutil.rmtree(work_dir, ignore_errors=True)
+
+    work_dir.mkdir(parents=True, exist_ok=True)
+    source_path = work_dir / Path(file_name).name
```
daniellekain, line 48: can we include the list number here so we maintain the same naming conventions we had before  (reports have always been <list_id>.~PROCESSED.<original filename> — the moderation tooling reads the id from that)

daniellekain:
This looks great! Left a minor comment. 

Another learning from a real production list record (9588) that I was looking at today: 

(1) file_name is not the S3 object key -- it has spaces normalized + storage dedup suffix, so s3.download(file_name) will 404 on files like this. The key needs to come from the file field's URL. 
(2) That list is a CSV and the live pipeline processed it normally — the new handler's hard reject of non-.xlsx would fail real uploads that work today; we either need CSV conversion in process_list or to keep parity with whatever the current DAG does for CSVs (looks like  pd.read_csv(filepath) → renames to .xlsx → df.to_excel(filepath).).

daniellekain (approved):
lgtm! do we need to address the sonar findings?

src/contribot/process_list/handler.py
protsack-stephan: sure thing, I'll do that in a follow-up PR

protsack-stephan:
> lgtm! do we need to address the sonar findings?

Thanks, I've looked through the Sonar Cube stuff, it does not seem relevant for this specific workflow.

PR #1183, opened 2026-08-06
[OSDEV-3118] Build additional SLC form checks

jamesjohnson4395:
## Jira Ticket

[OSDEV-3118](https://opensupplyhub.atlassian.net/browse/OSDEV-3118)

---

## Summary of Changes

Add AI-powered data-quality warnings and a hard vertical-bar separator check to the SLC submission flow.

Flag likely data-quality issues in Single Location Contribution (SLC) submissions — an implausible name or address, an address that doesn't match the selected country, or multiple facilities bundled into one submission — before the moderation event is created, and reject outright any product/location/processing type value that embeds the `|` character. This reduces bad data reaching the moderation queue while keeping contributors in control: quality warnings are overridable, malformed type values are not.

Builds directly on OSDEV-2980, which established the soft-block-with-override pattern this PR generalizes. It is also the first LLM integration in the Django backend.

### Frontend: vertical-bar separator validation (blocking)

Each `productType`, `locationType`, and `processingType` value rejects an embedded vertical bar (`|`) as a hard inline validation error that disables Submit, with no "submit anyway" bypass. Only the pipe is blocked because it is the backend's own multi-value join character, so a value containing it would be re-split downstream; commas, slashes, backslashes, ampersands, and the word "and" appear in legitimate single values (e.g. "Cut & Sew") and are allowed. Implemented as a `no-embedded-separators` test in `SLC_ARRAY_FIELD_TESTS` (`src/react/src/util/util.js`), ordered before the latin-characters-only test (which also rejects `|`) so a pipe gets the specific "enter as separate values" message rather than the misleading Latin-characters one.

### Backend: AI submission-quality check

A new `SubmissionQualityProcessor` joins the moderation-event creation chain (after `DuplicateSubmissionProcessor`, before `GeocodingProcessor`). It makes **one** LLM call per submission — not one per check — returning verdicts for:

- **name quality** — does the name look like a plausible facility/business name
- **address quality** — does the address look real and sufficiently specific
- **address–country match** — does the address appear to match the selected country
- **multiple locations** — does the name or address appear to describe more than one facility

If any verdict fires, the moderation event is **not** created; the API returns HTTP 409 with a new generic `warnings: [{type, title, message}]` array, kept separate from the existing `errors`/`duplicate_of` contract. Warnings are computed per-request and returned in the response only — persistence on `ModerationEvent` is explicitly out of scope.

The processor **fails open**: any error building the Bedrock client or agent, any error or timeout (5s) during the call, and any output that fails schema validation is logged and skipped — a slow or broken LLM never blocks a legitimate submission. The whole check also sits behind a new `slc_submission_quality_check` waffle switch (created **active** by migration `0226`, so deploying changes nothing), letting it be turned off in the Django admin without a deploy; a missing switch row falls back to waffle's `False` default, disabling the check rather than blocking submissions.

**Model access** is via AWS Bedrock through **pydantic-ai** (`pydantic-ai-slim[bedrock]`), which is model- and vendor-independent: the output schema and instructions are defined against pydantic-ai's `Agent`, and only the model construction is Bedrock-specific (chosen because Bedrock authenticates with the ECS task's IAM role — no API key to manage). Claude Haiku 4.5 is only invocable through the cross-region inference profile `eu.anthropic.claude-haiku-4-5-20251001-v1:0` (Bedrock rejects the bare foundation-model ID); confirmed working end-to-end in this account. Swapping the model is an env-var + tfvars change. The system-level instructions are overridable per environment via `SUBMISSION_QUALITY_INSTRUCTIONS`.

- **IAM**: a new `bedrock_invoke_submission_quality_model` policy (`deployment/terraform/iam.tf`) attached to the existing Django ECS task role, scoped to the inference-profile ARN plus a region-wildcarded foundation-model ARN (required because the profile routes across regions).
- **Local dev**: a new optional `BEDROCK_AWS_PROFILE` env var (`.env.sample`) points this call at a developer's real named AWS profile, because the Django container hardcodes MinIO's dummy AWS credentials, which would otherwise shadow real ones; leave unset in production.

### Cost/volume monitoring instead of an in-app cap

Per-user volume is already bounded by the endpoint's `DataUploadThrottle` (30/minute), and organic SLC volume is ~40 submissions/week (~$0.01/week of Bedrock spend), so runaway usage is handled by monitoring rather than an in-app cap — an accepted-risk decision documented in `doc/ops/monitoring.md`:

- Each successful call logs its token usage (`Submission quality check tokens: input=… output=…`) to CloudWatch Logs.
- A CloudWatch alarm on `AWS/Bedrock` `Invocations` (hourly sum > 100 by default, no `ModelId` dimension, `notBreaching` on missing data) pages via the existing SNS → Chatbot → Slack path.
- An account-level monthly AWS Budget on Bedrock spend ($25 default) alerts at 80% actual / 100% forecasted through the same topic; created only by the account-owner envs (`manage_bedrock_cost_budget = true` in Test and Production tfvars, mirroring the Chatbot ownership pattern). `aws_sns_topic.global` gains an explicit topic policy (restating default owner access) with a `budgets.amazonaws.com` publish grant.
- Also restores a closing brace on `bedrock_submission_quality_foundation_models` in `deployment/terraform/variables.tf` that was lost in a merge from `main` and would have failed Terraform parsing.

### Frontend: combined warning dialogs and override accumulation

- All warnings within a tier are shown together in one dialog: client-detectable warnings (PO Box / short address) before any request is sent, server-returned warnings in a second dialog. `ContributionWarningDialog` was generalized to take a `warnings` array; the duplicate-submission dialog migrated to the same API with no visible change.
- **"Submit anyway"** resubmits with `?ignore_warnings=true` (mirroring `duplicate_override`), which skips the LLM call entirely on that resubmission.
- **Overrides accumulate** across dialogs within one submission attempt: each dialog's "Submit anyway" resends every override already granted (`duplicate_override` and/or `ignore_warnings` together), so confirming one warning never re-triggers an already-confirmed check — without this the duplicate and quality dialogs oscillated indefinitely. A fresh submission from the form resets both overrides.

### Fix: rejected submissions no longer trip the duplicate-request throttle

`DuplicateThrottle` records a hash of user + body + query params before the view runs and 429s any identical request within 15 minutes. A submission rejected by the quality or duplicate check left that entry behind, so "Go back and edit" followed by resubmitting unchanged data hit a raw 429 instead of re-running the checks. Fixed by adding `DuplicateThrottle.clear()`, called from `finalize_response` for every error response except 429 itself: a rejected request created nothing, so its throttle entry is deleted and an identical retry passes. Double-clicks stay blocked while a request is in flight; successful submissions keep the full 15-minute protection.

---

## Type of Change

- [x] UI / frontend change

---

## Testing

- **Django** (`docker compose exec django python manage.py test`): `test_submission_quality_service.py` (schema validation, fail-open on model errors/timeouts/invalid output/client construction failure, instructions override, token-usage logging), `test_submission_quality_processor.py` (flagging, multiple warnings, `ignore_warnings` bypass, waffle switch on/off, fail-open, non-SLC/UPDATE exclusions), and `test_production_locations_create.py` (endpoint contract incl. throttle clearing) — 40 tests passing; migration `0226` exercised by test-DB setup.
- **React** (`docker compose exec react yarn test`): separator validation (pipe blocked bare and spaced; comma/slash/backslash/ampersand/"and" allowed at both the`hasEmbeddedSeparator` and schema level), combined client/server warning dialogs, `ignore_warnings` resubmission, override accumulation in both orders — 188 utils + 21 `ProductionLocationInfo` tests passing.
- **Bedrock end-to-end**: the inference-profile call was confirmed by hand against Bedrock in this account.
- **Terraform**: `alarms.tf` / `variables.tf` parse-checked with `terraform fmt` (hashicorp/terraform:1.9); the new SNS topic policy and budget should be eyeballed in the `terraform plan` output on deploy.

---

## Known TODO Items

- Verify the AWS Budget's `Service = "Amazon Bedrock"` cost filter against Cost Explorer once real spend appears — some Bedrock-hosted models bill under marketplace-style service names (noted in a code comment).
- Screenshots/recording of the SLC warning dialogs to be attached for visual review.

---

### Implementation

### Testing & Validation

### Final Review

[OSDEV-3118]: https://opensupplyhub.atlassian.net/browse/OSDEV-3118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/serializers/v1/ignore_warnings_query_param_serializer.py
```diff
@@ -0,0 +1,22 @@
...
+        deliberate user action, a permissive parser risks an unintended
+        value (e.g. a stray "1" or "false") silently bypassing it.
+        '''
+        if value not in ('true', 'false'):
+            raise ValidationError(
+                'The ignore_warnings query parameter must be '
+                'exactly "true" or "false".'
+            )
+
+        return value == 'true'
```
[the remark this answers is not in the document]
jamesjohnson4395: This doesn't affect the functionality of this feature, and this feature's implementation matches the duplicate processor. We could just follow up on this in a future PR, where we can change this for all implementations. Not necessary right now.

src/react/src/__tests__/components/ProductionLocationInfo.test.js
```diff
@@ -622,3 +647,245 @@ describe("ProductionLocationInfo component, possible duplicate submission dialog
...
+            <Router>
+                <ProductionLocationInfo submitMethod="POST" />
+            </Router>,
+            { preloadedState },
+        );
+
+    test("carries the granted duplicate override when confirming the quality warnings dialog", async () => {
+        apiRequest.post.mockRejectedValue({
+            response: { status: 422, data: warningsRawData },
+        });
```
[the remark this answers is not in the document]
jamesjohnson4395: Addressed in recent push.

src/django/api/throttles.py
```diff
@@ -53,6 +53,19 @@ def _serialize_data(self, data):
...
+        pk = view.kwargs.get("pk")
+        pk_prefix = f":{pk}" if pk else ""
+        # Query params are included so that requests with the same body but
+        # different query params (e.g. a duplicate-check override retry via
+        # ?duplicate_override=true) aren't mistaken for the same request.
+        query_str = self._serialize_data(dict(request.query_params))
+        data_hash = hashlib.sha256(
+            f"{data_str}{query_str}".encode()
+        ).hexdigest()
+        return f"duplicate:{request.user.id}{pk_prefix}:{data_hash}"
```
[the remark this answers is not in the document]
jamesjohnson4395: Addressed in recent push.

src/django/api/services/submission_quality_service.py
```diff
@@ -0,0 +1,195 @@
...
+                multiple_locations=QualityVerdict(
+                    **data['multiple_locations']
+                ),
+            )
+        except (KeyError, TypeError):
+            logger.exception(
+                'Submission quality check returned an unexpected shape; '
+                'skipping (fail open).'
+            )
+            return None
```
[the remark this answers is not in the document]
jamesjohnson4395: Addressed in recent push.

deployment/terraform/iam.tf
```diff
@@ -93,6 +93,33 @@ resource "aws_iam_role_policy" "s3_read_write_files_bucket" {
...
+    effect = "Allow"
+
+    resources = [
+      "arn:aws:bedrock:${var.aws_region}:${data.aws_caller_identity.current.account_id}:inference-profile/eu.anthropic.claude-haiku-4-5-20251001-v1:0",
+      "arn:aws:bedrock:*::foundation-model/anthropic.claude-haiku-4-5-20251001-v1:0",
+    ]
+
+    actions = [
+      "bedrock:InvokeModel",
+    ]
```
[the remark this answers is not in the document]
jamesjohnson4395: Addressed in recent push.

src/django/api/moderation_event_actions/creation/location_contribution/processors/submission_quality_processor.py
```diff
@@ -0,0 +1,109 @@
...
+        for warning_type, title in _WARNING_TITLES.items():
+            verdict = getattr(verdicts, warning_type)
+            if verdict.flagged:
+                warnings.append({
+                    'type': warning_type,
+                    'title': title,
+                    'message': verdict.reason,
+                })
+
+        return warnings
```
protsack-stephan, line 124: Do I understand correctly that this will be triggered every time someone tries to submit a contribution through SLC or the API? Are we sure we want to trigger this kind of action here? Maybe it makes sense to have a separate API endpoint for this?

src/django/api/services/submission_quality_service.py
```diff
@@ -0,0 +1,214 @@
...
+    Fails open: any error, timeout, or unexpected response shape is logged
+    and returns None so the caller skips the warning rather than blocking
+    a legitimate submission.
+    '''
+
+    def __init__(self):
+        self._client = AnthropicBedrock(
+            aws_region=BEDROCK_AWS_REGION,
+            aws_profile=BEDROCK_AWS_PROFILE,
+        )
```
protsack-stephan: Are we sure we want to use Anthropic's SDK here? Bedrock allows us to use different models that are potentially cheaper than Anthropic's models. Also, we might want to experiment with different workflows. I would suggest we abstract this logic into a vendor-independent SDK, which would allow us to use models from different providers and swap them easily. For example, we could use https://pydantic.dev/docs/ai/overview/ or something like that.

deployment/terraform/iam.tf
```diff
@@ -93,6 +93,39 @@ resource "aws_iam_role_policy" "s3_read_write_files_bucket" {
...
+# actually invokes in), not var.aws_region. The eu.* profile is only
+# invocable from EU regions, so tying this ARN to var.aws_region would
+# silently break the check (which fails open) on any non-EU deployment.
+data "aws_iam_policy_document" "bedrock_invoke_submission_quality_model" {
+  statement {
+    effect = "Allow"
+
+    resources = [
+      "arn:aws:bedrock:eu-west-1:${data.aws_caller_identity.current.account_id}:inference-profile/eu.anthropic.claude-haiku-4-5-20251001-v1:0",
+      "arn:aws:bedrock:*::foundation-model/anthropic.claude-haiku-4-5-20251001-v1:0",
```
protsack-stephan: Are we sure we want to restrict this to a single model? Maybe it makes sense to keep it more open endned?

protsack-stephan (commented):
Great work! A couple of questions/suggestions from me. Overall, my main concern is whether this code will be triggered for API users—if it is, it might cause long wait times for API requests. I might be missing the context here, so please let me know if I'm missing something.

src/django/api/services/submission_quality_service.py
```diff
@@ -0,0 +1,214 @@
...
+        try:
+            response = self._client.messages.create(
+                model=SUBMISSION_QUALITY_MODEL_ID,
+                max_tokens=1024,
+                tools=[_TOOL_SCHEMA],
+                tool_choice={'type': 'tool', 'name': _TOOL_NAME},
+                messages=[{
+                    'role': 'user',
+                    'content': (
+                        'Evaluate this production location submission.\n'
```
valencia-max: Probably this prompt could be a configuration? I thinking about the posibility it needs to change in production/preprod/staging after the deployment

src/django/api/services/submission_quality_service.py
jamesjohnson4395: Thank you for the suggestion! The change has been revised to use Pydantic!

deployment/terraform/iam.tf
jamesjohnson4395: Changes have been pushed so models are now changeable via terraform variables & env variables changes!

src/django/api/services/submission_quality_service.py
jamesjohnson4395: Instructions for the model have now been added as an optional env variable. So, we can control task framing, strictness, etc outside of the deployment process. Definitions for each individual checks remain in the code, as they are mapped to the frontend warnings which would also need a deploy to change.

A note from Claude: for the env var to actually be settable in production, it needs to be plumbed into the ECS task definition (wherever BEDROCK_SUBMISSION_QUALITY_MODEL_ID/BEDROCK_AWS_REGION would be set — I didn't see those wired into the terraform container definitions on this branch either, so presumably that's handled in the private deployment config). Worth confirming that when the PR lands.

src/django/api/moderation_event_actions/creation/location_contribution/processors/submission_quality_processor.py
jamesjohnson4395: Nope - the check skips immediately if the moderation event is not an SLC request (see lines 52-53). Theoretically, the check could trigger if an API submission is tagged with source='SLC', but I don't think we necessarily need to account for this.

daniellekain:
Notes from a chat @jamesjohnson4395 and I had -- 

Before merging this, we should figure out how we can monitor the usage/cost of these checks and set up any necessary guard-rails. 

Relatedly we discussed that it may be helpful to put this behind a switch in case of unexpected cost or latency

src/react/src/util/util.js
```diff
@@ -1861,6 +1861,17 @@ export const isShortAddress = value => {
     return value.trim().length < SLC_FORM_CONSTRAINTS.MIN_ADDRESS_LENGTH;
 };
 
+const EMBEDDED_SEPARATOR_PATTERN = /[,\\/&]|\band\b/i;
+
+export const hasEmbeddedSeparator = fieldValue => {
```
daniellekain, line 1870: we chatted about this briefly -- but just putting in a reminder that we may want to check if this will over-filter on some allowed facility processing types and location type

deployment/terraform/variables.tf
```diff
@@ -1218,6 +1218,22 @@ variable "contribot_last_list_id" {
...
+  default     = "eu-west-1"
+}
+
+variable "bedrock_submission_quality_inference_profiles" {
+  type        = list(string)
+  description = "Bedrock inference profile IDs the SLC submission quality check may invoke. The active one is selected by the app's BEDROCK_SUBMISSION_QUALITY_MODEL_ID env var; keep the two in sync."
+  default     = ["eu.anthropic.claude-haiku-4-5-20251001-v1:0"]
+}
+
+variable "bedrock_submission_quality_foundation_models" {
```
daniellekain, line 1233: I think we're missing the closing brace for this variable

src/react/src/util/util.js
jamesjohnson4395: I decided to scope down to only blocking if the '|' separator is present for now, as this is the one that we typically use for imports & exports. If we see other submissions coming in erroneously, we can update this in the future.

jamesjohnson4395: Otherwise, switch and AWS alarms have been added!

daniellekain:
Awesome work, James! This is a really great PR and will serve as a really helpful reference for future LLM features in the backend. 

Earlier blockers have been addressed (pipe-only separator with the join-character rationale and tests; the variables.tf brace restored), and the waffle switch + monitoring package (alarm, budget, token logging, monitoring.md) is exactly the right shape, including the budgets.amazonaws.com SNS grant. 

One QA-note suggestion, non-blocking: make sure the first Test pass includes a garbage SLC that comes back 409 — that's the first time the task-role→inference-profile IAM chain is exercised (local dev bypasses it via BEDROCK_AWS_PROFILE), and since the check fails open, a broken grant would otherwise be invisible. 

Nice work!

daniellekain (approved):

deployment/terraform/variables.tf
jamesjohnson4395: fixed!

PR #1214, opened 2026-08-21
Add Polygon model with GeoJSON admin upload and boundary query

emilyoxford-osh:
> [!NOTE]
> **Reviewed together with #1215** — that draft PR (the India Labour Line helpline) is stacked on this branch. Reviewing the two together shows both the building block and how it gets used, but I recommend starting with this one.

## Jira Ticket

Related context: [OSDEV-3125](https://opensupplyhub.atlassian.net/browse/OSDEV-3125) (a manual data request for textile facilities in specific Indian regions for an India Labour Line hotline — the motivating real-world use case). Note this PR does not close that ticket: it's foundational infrastructure, not the requested data deliverable.

---

## Summary of Changes

Adds a `Polygon` model: a reusable, named geographic boundary (a unique variable-style `name`, an optional human-friendly `display_name`, a required `description`, geometry) that is not tied to country borders and supports arbitrary shapes, disjoint multi-part polygons, and polygons with holes.

This is meant as a foundation for future boundary-driven features — e.g. showcasing locations within a region, or auto-populating a partner field (like a labor hotline number) for locations within a specific area — rather than a user-facing feature by itself. Nothing yet uses `Polygon` in production; this PR only adds the building block for that future functionality, gated behind an off-by-default `polygons` waffle switch.

- `Polygon.get_production_locations()` returns the OS Hub production locations within a boundary, using the same `location__within` PostGIS lookup already used by the existing `boundary` facility-search query param, so it inherits that code path's performance characteristics.
- Boundaries are entered via the Django admin by pasting or uploading GeoJSON (`api/helpers/geojson_polygon.py`): accepts a bare `Polygon`/`MultiPolygon`, a `Feature`, or a `FeatureCollection` (features' geometries are merged), validates the result is a real, valid polygon, and surfaces a friendly form error otherwise instead of a 500.
- New migration `0224_create_polygon.py` was hand-written rather than autogenerated: `makemigrations` bundled in ~45 unrelated operations from pre-existing drift between models and migration history on `main` (e.g. altering unrelated fields, dropping `FacilityField`/`FacilityListItemField`). That drift is unrelated to this change and is not addressed here; flagging it separately for the team's awareness.

 Requirement 1: Staff can create a named boundary in the Django admin by pasting GeoJSON text.
- Polygon model (name, display_name, description, geometry) + database migration. → Boundaries have a permanent home.
- Admin form with a paste box, validated on save. → Staff add boundaries with no developer involved.
  - Code: [Polygon model](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon.py#L22) · [migration](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/migrations/0228_create_polygon.py) · [PolygonForm](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon_admin.py)

  Requirement 2: Staff can alternatively upload a GeoJSON file.
  - File-upload field on the same admin form (file wins if both provided; both optional when editing an existing polygon). → Big boundary files don't have to be pasted.
  - Code: [file-upload field](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon_admin.py) · [file/paste handling in clean()](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon_admin.py)

  Requirement 3: Real-world boundary shapes are accepted.
  - Geometry stored as MultiPolygon; parser accepts a bare Polygon/MultiPolygon, a Feature, or a FeatureCollection (features merged). → Disjoint regions (state + islands) and polygons with holes all work, and features that touch along shared borders (neighboring districts exported together) are dissolved into one piece instead of being wrongly rejected as invalid.
  - Code: [parse_polygon_geojson()](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/helpers/geojson_polygon.py#L105) (Feature/FeatureCollection unwrapping from [L82](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/helpers/geojson_polygon.py#L144))

  Requirement 4: Malformed or invalid GeoJSON produces a clear admin form error, never a crash. 
  - Parser validates real, non-self-intersecting polygon geometry; admin form surfaces failures as the standard red error box with the staff member's input preserved. → Garbage can't enter the database; staff see what to fix.
  - Code: [geometry validity check](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/helpers/geojson_polygon.py#L186) · [form error surfacing](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon_admin.py) · [tests](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/tests/test_geojson_polygon_helper.py)

  Requirement 5: GeoJSON declaring a non-WGS-84 coordinate system is rejected with re-export instructions. (WGS-84 is the OS Hub default already)
  - Add a check in the parser for a declared coordinate system (the legacy crs member); accept WGS 84 declarations and files with none (the spec default), reject anything else. → Red admin error like "declares a coordinate system other than WGS 84 (found: EPSG:32644) — please re-export in WGS 84 (EPSG:4326)" instead of a silently misplaced boundary. 
  - Tests: WGS-84 declarations accepted (both common spellings), foreign declaration rejected, malformed declaration rejected.
  - Code: [_check_crs()](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/helpers/geojson_polygon.py#L30)
  
  Requirement 6: Coordinates outside valid longitude/latitude ranges are rejected. 
  - Add a bounds sanity check after parsing: any coordinate beyond ±180 longitude / ±90 latitude → clear error explaining the file likely uses a projected coordinate system. → Catches the undeclared wrong-projection case (meters-based coordinates with no crs label).
  - Tests: projected-looking coordinates rejected; legitimate extreme coordinates accepted.
  - Code: [_check_bounds()](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/helpers/geojson_polygon.py#L78)
  
  Requirement 7: Given a polygon, the system returns the production locations inside it.
  - Polygon.get_production_locations() using PostGIS containment (same code path as the existing boundary search parameter), with correctness tests (holes excluded, all parts of multi-part polygons matched) and a performance benchmark. → The query the follow-up PR's behaviors stand on.
  - Code: [get_production_locations()](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon.py#L102) · [containment tests](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/tests/test_polygon_model.py)

  Requirement 8: Callers can narrow that query with filters on standard facility fields. 
  - Add an optional filters argument to get_production_locations(): a dict like {"country": ["IN"], "sector": ["Apparel"]}, validated against a allowlist of core fields (country, sector, facility type, processing type, product type, parent company — all confirmed to exist as filterable columns on the search index). Unknown keys raise a clear error instead of being ignored; country codes are case-normalized so "in" can't silently match nothing. → Cheap column checks and the polygon test run as one database query, with PostgreSQL automatically doing the cheap checks first — so no pre-filter/post-filter ordering decision needed. the follow-up PR's provider uses this for its "India only" guard.
  - Tests: country filter includes/excludes correctly, sector filter works, unknown key errors.
  - Code: [FILTERABLE_FIELDS allowlist](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon.py#L93) and the filter loop inside [get_production_locations()](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon.py#L102)

  Requirement 9: Polygon names are unique and machine-friendly. 
  - Add a uniqueness constraint on Polygon.name. Duplicates surface as a normal admin form error before hitting the database. → No duplicate "Maharashtra" rows to confuse staff, and the follow-up PR's by-name references become unambiguous.
  - Names must be variable-style identifiers (letters, digits, underscores; no leading digit), enforced by a model validator with a friendly admin message. This is to make sure that the polygons can be easily referenced in a machine-readable way. An optional display_name field carries the human-readable label in case boundaries are ever displayed on OS Hub.
  - Tests: duplicate name rejected (form level and database level), valid/invalid name styles, display_name optional and saved.
  - Code: [name validator](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon.py#L12) · [unique name field](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon.py#L41) · [display_name](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon.py#L51) · [tests](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/tests/test_polygon_admin_form.py)

  Requirement 9b: Staff can see and verify a saved boundary when editing it. 
  - The admin edit form pre-fills the paste box with the saved boundary's GeoJSON (size-capped), and a read-only boundary summary (parts, vertex count, lon/lat extent) appears on the change page and as a list column. → Staff can verify at a glance that a polygon is where it should be; previously the stored geometry was invisible after saving.
  - Tests: prefill round-trips to the identical shape; summary text; unsaved-polygon fallback.
  - Code: [GeoJSON prefill in PolygonForm.__init__](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon_admin.py) · [boundary_summary()](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon_admin.py)

  Requirement 10: ~~The whole feature sits behind a waffle switch.~~ Removed per review (protsack-stephan, daniellekain): the switch and its migration are gone. The admin section is simply available to admins, and nothing user-facing exists until something consumes a polygon.

---

## Changes 8/24/26 (in response to review)

- **Touching features now merge instead of being wrongly rejected.** A FeatureCollection of adjacent regions (e.g. neighboring districts sharing a border) used to fail with a false "self-intersection" error. Each feature is now validated on its own, then the pieces are dissolved together — touching pieces become one, separate pieces stay separate. → [the merge step (`_combine_polygons`)](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/helpers/geojson_polygon.py#L183) · [test with two edge-sharing squares](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/tests/test_geojson_polygon_helper.py#L198)
- **Waffle switch removed.** The `polygons` switch, its migration, the admin gating, and the switch tests are gone; the Polygons admin section is simply available to admins. → [removal commit](https://github.com/opensupplyhub/open-supply-hub/commit/a87c62be)
- **Polygon admin moved to its own file**, matching the `partner_field_admin.py` precedent, to keep `admin.py` from growing. → [`models/polygon_admin.py`](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon_admin.py#L117)
- **GeoJSON uploads capped at 10 MB** with a friendly error suggesting simplification. → [the cap](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon_admin.py#L48)
- **"whitelist" → "allowlist"** in code comments, docstrings, and this description. → [example](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon.py#L92)
- **By-name polygon reference warnings removed** — superseded by #1215's ForeignKey from the partner field to its polygon, which is rename-proof and lets the database itself block deleting a polygon in use, so the admin-side name-watching became unnecessary. → [removal commit](https://github.com/opensupplyhub/open-supply-hub/commit/99cb79e9)
- The description's "performance benchmark" line was reworded to make clear it was a one-off shell measurement, not code in this PR.

- **Enclosed gaps are provably preserved.** Merging touching pieces never fills space neither piece claimed — two pieces wrapping around an enclave (e.g. a country surrounded by another) merge into one piece with a hole where the enclave is, and containment queries exclude holes. Now pinned by a test. → [the enclave test](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/tests/test_geojson_polygon_helper.py#L218)
- **The boundary summary now shows a hole count** (parts, holes, vertices, extent), so a deliberate enclave or an accidental sliver-riddled export is visible at a glance in the admin. → [the summary](https://github.com/opensupplyhub/open-supply-hub/blob/c7a719d6/src/django/api/models/polygon_admin.py#L131)

## Type of Change

- [ ] UI / frontend change

## Testing

- 35 tests covering the GeoJSON parsing helper (holes, multipolygons, Feature/FeatureCollection unwrapping, invalid/self-intersecting rejection, WGS-84/foreign/malformed CRS declarations, out-of-range projected coordinates), the admin form (file upload, pasted text, validation errors, name rules, duplicate names, display_name, GeoJSON prefill round-trip, boundary summary, waffle switch gating), and `Polygon.get_production_locations()` against a real PostGIS-backed test database (holes excluded, all parts of a multipolygon matched, country/sector filters, unknown-filter errors). All pass locally via `docker compose exec django python manage.py test`.
- Manually tested in a local dev environment: added real boundaries through the Django admin by paste and by file upload (a Washington DC bounding box, the actual Delhi NCR administrative boundary, and a two-piece disjoint multipolygon), created fake `FacilityIndex` records inside and outside them via the Django shell, and confirmed `get_production_locations()` (with and without filters) returned exactly the expected matches, including any-part matching on the two-piece boundary. Also verified both projection rejections end-to-end by feeding the admin the same Delhi boundary genuinely reprojected to UTM 43N (EPSG:32643), once with the CRS declared and once without, and verified the waffle switch hides/reveals the admin section live.
- `flake8` (run with CI's exact flags) and `manage.py check` are clean.
- A one-off shell benchmark (not included in this PR) ran the underlying PostGIS `location__within` lookup against a 2,000-vertex polygon over ~500k points; query time stayed in the single-digit-to-low-double-digit millisecond range, which informed the decision to build on PostGIS.

## Known TODO Items

- No management UI/API yet for calling `get_production_locations()` outside the Django shell — this PR is intentionally just the building block.

### Implementation

### Testing & Validation

- [ ] For UI changes: I have included screenshots or a recording, and verified responsive/cross-browser behavior. (N/A — no UI change)
- [ ] For architectural changes: I have documented the design decision and notified affected teams. (N/A — not an architectural change to existing systems)

- [ ] I have updated the Jira ticket status and added any relevant notes. (No ticket directly covers this work; see Jira Ticket section above)

### Final Review

[OSDEV-3125]: https://opensupplyhub.atlassian.net/browse/OSDEV-3125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/migrations/0228_add_polygons_switch.py
```diff
@@ -0,0 +1,33 @@
...
+    Migration to introduce a switch gating the polygon admin.
+    """
+
+    dependencies = [
+        ('api', '0227_create_polygon'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_switch, delete_switch),
+    ]
```
protsack-stephan: Are we sure we need a switch for this? I'd suggest omitting it if possible.

protsack-stephan (commented):
Good work! One small comment.

src/django/api/migrations/0228_add_polygons_switch.py
daniellekain: Ah the switch was my suggestion. I suggested a switch to help de-risk this change, though I'm not sure if a switch on the admin panel is the right spot. This switch wouldn't undo any changes and would make it harder to debug a potential issue. 

I started a thread on Slack to discuss further

src/django/api/helpers/geojson_polygon.py
```diff
@@ -0,0 +1,242 @@
...
+    can paste or upload whatever their tool produced, without hand-
+    editing files first. Shapes with holes, and boundaries made of
+    several separate pieces, are supported for the same reason: real
+    boundaries look like that.
+
+    Bad input is rejected with a message written for the person using
+    the admin form. That covers: broken JSON, geometry that isn't a
+    polygon, shapes that cross over themselves, empty shapes, a
+    declared non-WGS-84 coordinate system, and coordinates outside the
+    possible longitude/latitude ranges.
```
daniellekain, line 265: a potential bug spot from Claude -- 

The docstring promises bad input is rejected "with a message written for the person using the admin form," covering "shapes that cross over themselves." But a FeatureCollection of adjacent features — districts sharing borders, which is what a real regional export looks like, including the OSDEV-3125 India case — trips that same rejection through no fault of the input: a MultiPolygon's parts may not touch along an edge, so concatenating the features here manufactures the invalidity, and the person is told their shapes cross over themselves when nothing in their file does. Dissolving instead of concatenating (unary_union, re-wrapped as MultiPolygon) makes touching features merge into one valid polygon and keeps the docstring's promise. Worth a test with two squares sharing an edge — test_feature_collection_is_merged's squares are disjoint, so it can't catch this.

src/django/api/models/polygon.py
```diff
@@ -0,0 +1,143 @@
...
+    # that implements each one. Two kinds of lookup appear here because
+    # the underlying columns differ:
+    #   - `__in` fields hold ONE value per location (a location has
+    #     exactly one country), so the check is "is that value in the
+    #     requested list".
+    #   - `__overlap` fields are ARRAYS (a location can have several
+    #     sectors), so the check is "does the location's list share at
+    #     least one value with the requested list".
+    # To make a new field filterable, add an entry here — anything not
+    # in this whitelist is rejected with an error rather than silently
```
daniellekain: I'd prefer that we use allowlist/denylist (here and a couple other times throughout the PR)

daniellekain (commented):
Amazing work!! 

I left a few comments throughout; high-level they are
1) agree with Stephan, let's remove the switch in this PR. Let's chat through if we want another one somewhere -- I started a convo on Slack.
2) potential bug with checking for invalid shapes? commented inline
3) replace "whitelist" with "allowlist" throughout
4) the description mentions a "performance benchmark" -- but I'm not sure if I saw one -- if it's not there we can remove that from the description. Apologies if I missed it!
5) it may be a good idea to put an upper bound on geojson_file.read(), but not strictly necessary

src/django/api/admin.py
```diff
@@ -309,6 +316,231 @@ class USCountyTigerlineAdmin(admin.ModelAdmin):
...
+        boundary is kept; if neither is provided on a brand-new
+        polygon, that is an error (a polygon must have a boundary).
+        """
+        cleaned_data = super().clean()
+        geojson_file = cleaned_data.get('geojson_file')
+        geojson_text = cleaned_data.get('geojson_text')
+
+        if geojson_file:
+            try:
+                raw = geojson_file.read().decode('utf-8')
```
daniellekain: nit: we may want to cap this? 

something like: 
```
# Cap uploads: bigger than this almost certainly means an
# undissolved/unsimplified export, and the geometry would bloat
# every containment query even if it parsed.
GEOJSON_FILE_MAX_BYTES = 10 * 1024 * 1024  # 10 MB

if geojson_file:
    if geojson_file.size > self.GEOJSON_FILE_MAX_BYTES:
        raise forms.ValidationError(
            f'GeoJSON file is too large '
            f'({filesizeformat(geojson_file.size)}; limit '
            f'{filesizeformat(self.GEOJSON_FILE_MAX_BYTES)}). '
            'Simplify the boundary (e.g. with mapshaper.org) and '
            'try again.'
        )
    raw = geojson_file.read().decode('utf-8')
```
(filesizeformat from django.template.defaultfilters renders "10.0 MB" instead of a byte count.)

src/django/api/models/polygon.py
```diff
@@ -0,0 +1,144 @@
...
+            'is ever displayed on OS Hub.'
+        )
+    )
+    description = models.TextField(
+        blank=False,
+        help_text=(
+            'Details about what this boundary represents and where it '
+            'came from.'
+        )
+    )
```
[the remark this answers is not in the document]
emilyoxford-osh: Intentional: descriptions are required by design (every boundary must say what it represents) — the PR summary's "optional" referred only to display_name, and has been reworded to say so. The requirement is now also pinned by a form test.

src/django/api/helpers/geojson_polygon.py
```diff
@@ -0,0 +1,286 @@
...
+    try:
+        data = json.loads(raw)
+    except (TypeError, ValueError) as exc:
+        raise InvalidPolygonGeoJSON(f'Not valid JSON: {exc}') from exc
+
+    if isinstance(data, dict):
+        _check_crs(data)
+
+    geometries = _extract_geometries(data)
+    polygons = _parse_polygons(geometries)
```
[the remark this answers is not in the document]
emilyoxford-osh: Good catch — fixed in c7a719d6: coordinate-system labels on individual features and on geometry objects are now checked the same way as top-level ones, with a regression test covering both nesting spots.

doc/release/RELEASE-NOTES.md
```diff
@@ -15,32 +15,44 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
... joining, unique variable-style `name` — letters, digits, and underscores only — an optional human-friendly `display_name`, `description`, and a `geom` `MultiPolygonField` in EPSG:4326) that is not tied to country borders. Polygons may be any shape, including multiple disjoint parts and polygons with holes. `Polygon.get_production_locations()` returns the OS Hub production locations within a given boundary via the existing `location__within` PostGIS lookup, optionally narrowed by a whitelist of core-field filters (e.g. `{'country': ['IN'], 'sector': ['Apparel']}`) applied in the same query. Boundaries are added through the Django admin by pasting or uploading GeoJSON (a bare `Polygon`/`MultiPolygon`, a `Feature`, or a `FeatureCollection`, which are parsed, merged, and validated before saving); GeoJSON that declares a non-WGS-84 coordinate system, or whose coordinates fall outside valid longitude/latitude ranges, is rejected with a clear form error. This is foundational infrastructure for future boundary-driven features (e.g. per-region partner fields) rather than a user-facing change on its own; related context in [OSDEV-3125](https://opensupplyhub.atlassian.net/browse/OSDEV-3125).
```
daniellekain: super minor nit -- one last "whitelist"

daniellekain (dismissed):
looks amazing! Thank you for making all these changes!

one one-word request in-line

daniellekain (approved):
amazing!!!

PR #1215, opened 2026-08-21
[OSDEV-3125] Add India Labour Line helpline partner field driven by polygon boundaries

emilyoxford-osh:
> [!NOTE]
> The polygon foundation this builds on has merged (#1214), so the diff below is now exactly this PR's own changes.

## Jira Ticket

[OSDEV-3125](https://opensupplyhub.atlassian.net/browse/OSDEV-3125) — the India Labour Line hotline request that motivated the polygon work. This PR implements the hotline partner field itself.

---

## Summary of Changes

This PR adds the **India Labour Line helpline** to Open Supply Hub. The idea in one sentence: production locations that sit inside a defined coverage area in India (and in the correct sectors for the labour hotline) automatically get the helpline's phone number shown on their page.

It works the same way the MIT Living Wage feature already works, except the covered area is a map boundary (a polygon, which can be made from the functionality set up in PR #1214) instead of US counties.

What it does, requirement by requirement, with a link to the code that does it:

**1. A location inside the covered area, working in a covered sector, shows the helpline number on its page.**
When someone opens a location's page, the system checks three things: is this location in India, does its point on the map fall inside the coverage boundary, and does it work in one of the covered sectors (Apparel, Apparel Accessories, Footwear, Home Textiles, Leather, Textiles — the underlying request was a textile facility list)? If yes to all three, the helpline appears as a partner field, credited to the partner organization. The same sector rule applies to the profile page (requirement 3) and the search filter (requirement 4).
→ Code: [the covered-sector list](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/partner_fields/india_labour_line_provider.py#L34) This works on both the current and previous location page designs, and in the API responses that feed them.
→ Code: [the provider — the piece that answers "does this location get the helpline?"](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/partner_fields/india_labour_line_provider.py#L86)

**2. The phone number can be changed by staff, with no code release.**
The number is not written into the code. It lives in the partner field's "display text" box in the Django admin, and the system reads it fresh each time. Change it there, and every page shows the new number immediately. (Per review, the field itself is now protected like other system fields — but the display text box stays editable, so this still works.)
→ Code: [where the number is read](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/partner_fields/india_labour_line_provider.py#L236)

**3. The partner's profile page lists every covered location.**
The partner organization's profile shows all the covered locations as its spotlight locations. "Covered" is defined in exactly one place in the code - [a shared covered-locations query](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/partner_fields/india_labour_line_provider.py#L46), built on the polygon model's general-purpose lookup from #1214 - and both this page and the search filter (requirement 4) use it, so the two can never disagree.
→ Code: [the profile page branch](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/views/user/user_profile_facilities.py#L114)

**4. Picking the partner in the search filter shows only covered locations.**
Choosing this partner under "Spotlight Partners" in search narrows the results and the map to locations inside the boundary, and for the specified sectors for this spotlight partner only, via the same shared covered-locations definition as the profile page. 
→ Code: [the search filter branch](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/models/facility/partner_contributor_filter.py#L140)

**5. ~~The whole feature has an on/off switch.~~** Removed per review (protsack-stephan, daniellekain): the switch and its migration are gone. The feature is controlled by configuration — it shows nothing until the boundary polygon exists and the partner field is assigned, and removing either turns it off again.

**6. If something is set up wrong, the feature turns itself off — loudly — instead of breaking anything.**
If the coverage boundary or the phone number is missing, pages still load fine; the helpline simply doesn't appear, a warning lands in the logs saying exactly what's missing, and the search filter shows nothing rather than accidentally showing everything.
→ Code: [the missing-boundary warning](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/partner_fields/india_labour_line_provider.py#L32) · [the show-nothing-not-everything guard](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/models/facility/partner_contributor_filter.py#L140)

**7. The feature is connected to its boundary by a database link that can't break by accident.**
Per review (daniellekain's ForeignKey suggestion), the partner field now points at its coverage polygon through a database link set in the Django admin, instead of the code knowing the polygon's name. Renaming the polygon can't disconnect anything, and the database refuses to delete a polygon a field still points at. (The earlier name-watching warnings became unnecessary and were removed.)
→ Code: [the link on the partner field](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/models/partner_field.py#L115) · [how the provider follows it](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/partner_fields/india_labour_line_provider.py#L26)

**7b. The field itself is set up by code, like MIT Living Wage.**
Per review (protsack-stephan), a migration now creates the partner field — or, since production already has a hand-made one, adopts and normalizes it without deleting it (so its contributor assignment survives), marking it protected so it can't be renamed or deleted by accident. If the coverage polygon already exists when the migration runs, it links it automatically.
→ Code: [the migration](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/migrations/0230_setup_india_labour_line_partner_field.py)

**8. All of the above is tested.**
→ [Tests for the provider](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/tests/test_india_labour_line_provider.py) · [for the profile page](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/tests/test_user_profile_facilities.py) · [for the search filter](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/tests/test_partner_contributor_filter_india.py)

---

## Changes 8/24/26 (in response to review)

- **ForeignKey instead of name lookup.** The partner field now points at its coverage polygon through a database link set in the Django admin, replacing the polygon-name list that used to live in code. Renaming the polygon can no longer disconnect the feature, and the database blocks deleting a polygon a field points at. → [the link on PartnerField](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/models/partner_field.py#L115) · [the migration adding the column](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/migrations/0229_add_partner_field_polygon_link.py) · [how the provider follows the link](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/partner_fields/india_labour_line_provider.py#L26)
- **The field is now code-managed, like MIT Living Wage.** A migration creates the partner field — or adopts and normalizes the hand-made production row without deleting it (so its contributor assignment survives) — and marks it a protected system field, so it can't be renamed or deleted by accident. It also links the coverage polygon automatically if one with the expected name already exists. → [the migration](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/migrations/0230_setup_india_labour_line_partner_field.py)
- **The helpline number moved to `display_text`.** Protection locks the field's name, type, and JSON schema — so the number now lives in the display-text box, which stays editable on protected fields. Staff can still change the number in the admin with no code release. → [where the number is read](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/partner_fields/india_labour_line_provider.py#L236)
- **Waffle switch removed.** The `india_labour_line_helpline` switch, its migration, and every switch check are gone. The feature is controlled by configuration: it shows nothing until a coverage polygon is linked, and unlinking it turns the feature off again (with a logged warning, never an over-match). → [removal commit](https://github.com/opensupplyhub/open-supply-hub/commit/7ed6572c)
- **One flat query for the spotlight and search filter.** The shared covered-locations rule is now applied directly to each surface's own queryset instead of going through an intermediate list of IDs — same single definition, simpler and faster database plan. → [the shared rule](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/partner_fields/india_labour_line_provider.py#L62) · [the spotlight using it](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/views/user/user_profile_facilities.py#L117) · [the search filter using it](https://github.com/opensupplyhub/open-supply-hub/blob/e66d2aa6/src/django/api/models/facility/partner_contributor_filter.py#L140) 
- **Comment accuracy fix** in the spotlight view (it stated the right behavior for a wrong reason). *(review: daniellekain)*
- The `india_labour_line_helpline_ares` → `_areas` typo in this description's preconditions was fixed.

## Type of Change

- [ ] UI / frontend change

## Testing

- 17 new tests: the provider (in/out of boundary, non-India skip, switch off, missing polygon warns, phone number read from dict *and* string schema, missing phone default warns), the profile spotlight endpoint (in-boundary only; empty with switch off), and the search filter (in-boundary only; match-nothing when switch off or polygons missing). All pass locally via `docker compose exec django python manage.py test`.
- Regression: the full Django test suite passes locally on this branch (1,093 tests, 1 standard skip, 0 failures).
- Manually tested end-to-end in a local dev environment mirroring production config: created the `india_labour_line_helpline` partner field with the production JSON schema, the `india_labour_line_helpline_areas` polygon (Delhi-area stand-in), and flipped the switch; `GET /api/facilities/{id}/` returns the helpline (`"phone_number": "1-800-833-9020"`, correct contributor attribution and label) for an in-boundary facility and nothing for an out-of-boundary one.
- `flake8` (run with CI's exact flags) is clean.

## Known TODO Items

- **Stacked on #1214**: rebase onto main after that PR merges (squash-merge means dropping the duplicated commits — one `git rebase --onto`).
- **Production preconditions before the feature shows anything** (admin actions, deliberately not migrations — the field already exists in production): assign the partner field to the partner **contributor**, assign it to a **PartnerFieldGroup** (both the search filter and the filter UI only consider grouped fields), and create the `india_labour_line_helpline_areas` polygon via the #1214 admin.
- Polygon name `india_labour_line_helpline_areas` is per current plan — a rename is a one-line change to `INDIA_LABOUR_LINE_POLYGON_NAMES` if the final polygon is named differently.
- CSV/Excel download columns and any v1-API-specific work are intentionally out of scope (the field rides along in the detail endpoints via the provider registry; nothing more).

### Implementation

### Testing & Validation

- [x] All existing tests pass locally. (Full suite: 1,093 tests, OK.)
- [ ] For UI changes: I have included screenshots or a recording, and verified responsive/cross-browser behavior. (N/A — no UI change)
- [ ] For architectural changes: I have documented the design decision and notified affected teams. (N/A)

### Final Review

- [x] I have performed a self-review on my PR and I am presenting my best work for my peers to review. (Draft — pending author self-review.)

[OSDEV-3125]: https://opensupplyhub.atlassian.net/browse/OSDEV-3125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/migrations/0228_add_polygons_switch.py
```diff
@@ -0,0 +1,33 @@
...
+    Migration to introduce a switch gating the polygon admin.
+    """
+
+    dependencies = [
+        ('api', '0227_create_polygon'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_switch, delete_switch),
+    ]
```
protsack-stephan: Are we sure we really need this switch? I'd  suggest omitting it if possible.

src/django/api/migrations/0229_add_india_labour_line_switch.py
```diff
@@ -0,0 +1,34 @@
...
+    helpline partner field (provider, spotlight, and search filter).
+    """
+
+    dependencies = [
+        ('api', '0228_add_polygons_switch'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_switch, delete_switch),
+    ]
```
protsack-stephan: Are we sure we really need this switch? I'd suggest omitting it if possible. We can control this through partner field assignment.

src/django/api/partner_fields/india_labour_line_provider.py
```diff
@@ -0,0 +1,296 @@
...
+# Matching is exact-text against the production sector taxonomy, so
+# these must be spelled exactly as they appear in the sector dropdown.
+INDIA_LABOUR_LINE_SECTORS = [
+    'Apparel',
+    'Apparel Accessories',
+    'Footwear',
+    'Home Textiles',
+    'Leather',
+    'Textiles',
+]
```
protsack-stephan, line 23: Is there a need to dynamically control these? It might be worth exposing them through the admin panel.

src/django/api/partner_fields/india_labour_line_provider.py
```diff
@@ -0,0 +1,296 @@
...
+        sectors = getattr(production_location, 'sector', None)
+        if sectors is None:
+            from api.models.facility.facility_index import FacilityIndex
+            sectors = (
+                FacilityIndex.objects
+                .filter(id=production_location.id)
+                .values_list('sector', flat=True)
+                .first()
+            ) or []
+        return bool(set(sectors) & set(INDIA_LABOUR_LINE_SECTORS))
```
protsack-stephan, line 236: Are we sure this assumption is correct? I think we should be able to get the sector values in the new API endpoint either from OpenSearch or from the database, so we don't have to do the query here. But I might be wrong, so it's worth double-checking.

src/django/api/views/user/user_profile_facilities.py
```diff
@@ -105,6 +112,22 @@ def get_queryset(self):
...
+        # matches nothing for a system field, since system fields have
+        # no ExtendedField rows).
+        if (
+            IndiaLabourLineProvider.FIELD_NAME in partner_fields
+            and switch_is_active(INDIA_LABOUR_LINE_SWITCH)
+        ):
+            covered = get_covered_production_locations()
+            return self.__base_queryset().filter(
+                id__in=covered.values('id')
+            )
```
protsack-stephan: This query might be quite expensive, especially if we are filtering through the IDs. It might be worth exploring better or faster ways to get the data and using a different filter. Just a suggestion—happy to brainstorm together.

src/django/api/admin.py
```diff
@@ -309,6 +316,231 @@ class USCountyTigerlineAdmin(admin.ModelAdmin):
...
+        return (
+            switch_is_active(POLYGONS_SWITCH)
+            and super().has_change_permission(request, obj)
+        )
+
+    def has_delete_permission(self, request, obj=None):
+        return (
+            switch_is_active(POLYGONS_SWITCH)
+            and super().has_delete_permission(request, obj)
+        )
```
protsack-stephan: If possible, I would suggest we move this to a different file, something like `polygon_admin.py` in the models section, to keep this specific file smaller, as it's already quite large.

src/django/api/partner_fields/india_labour_line_provider.py
```diff
@@ -0,0 +1,296 @@
...
+                name__in=INDIA_LABOUR_LINE_POLYGON_NAMES
+            ).exists():
+                logger.warning(
+                    'No India Labour Line polygons exist in the '
+                    'database (expected names: '
+                    f'{sorted(INDIA_LABOUR_LINE_POLYGON_NAMES)}). The '
+                    'field will not appear anywhere until they are '
+                    'created.'
+                )
+            return None
```
protsack-stephan, line 168: Are we sure this is useful? If we want to clearly distinguish between those failures, maybe we can query for the geometry first and then check if the location is in the polygon directly in code versus in the database. That would simplify the logic a little bit. Just a suggestion, feel free to ignore it.

protsack-stephan (commented):
Good work! One thing I would suggest is creating this new field with a migration and removing the manually added one in production. That way, we can follow the same pattern where all of the 'system fields' are created through migrations and can't be easily deleted, just disabled. Also, because we tie the code to a specific field, that would give us a better guarantee that a misconfiguration does not break the functionality.

src/django/api/views/user/user_profile_facilities.py
daniellekain: perhaps `get_covered_production_locations()` accept a base queryset and apply `location__within` +`sector__overlap` to it directly? One flat query, no IN, and the shared single-source-of-truth definition survives.

src/django/api/views/user/user_profile_facilities.py
```diff
@@ -105,6 +112,22 @@ def get_queryset(self):
                 country_code__in=["US"],
             )
 
+        # The India Labour Line spotlight shows the locations the
+        # helpline covers, using the shared covered-locations
+        # definition so this page can never disagree with the search
+        # filter. Gated by the feature's waffle switch; if the switch
+        # is off, fall through to the generic branch below (which
```
daniellekain: Claude nit: this comment says the generic branch matches nothing because "system fields have no ExtendedField rows," but the field is created with system_field=False (tests and prod). The behavior is right — provider-driven fields have no EF rows regardless of the flag — it's the stated reason that's off. One-line comment fix (and moot in the other direction if the field becomes migration-created with system_field=True).

doc/release/RELEASE-NOTES.md
```diff
@@ -15,11 +15,15 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
 
 #### Schema changes
 * [OSDEV-3142](https://opensupplyhub.atlassian.net/browse/OSDEV-3142) - Added the `is_anonymized` column to `api_source`.
 
 ### Code/API changes
+* Added the India Labour Line helpline as a system partner field (`india_labour_line_helpline`), mirroring the MIT Living Wage pattern but keyed on `Polygon` boundaries instead of US counties: production locations in India whose coordinates fall inside the `india_labour_line_helpline_areas` polygon, and that work in a covered sector (Apparel, Apparel Accessories, Footwear, Home Textiles, Leather, Textiles), get the helpline phone number (read at request time from the partner field's JSON schema `properties.phone_number.default`, so the number is editable in the Django admin without a deploy). The partner's contributor-profile spotlight and the Spotlight Data Partners search filter both resolve via PostGIS polygon containment. Everything is gated behind the `india_labour_line_helpline` waffle switch; missing configuration (polygon or schema) disables the feature with a logged warning rather than erroring or over-matching. Requires no front-end changes.
```
daniellekain: One typo to fix before we deploy -- the code and release notes name the polygon india_labour_line_helpline_areas, but the PR body's production preconditions say india_labour_line_helpline_ares. I'm guessing that whoever is following the runbook would figure this out, but still worth fixing

src/django/api/partner_fields/india_labour_line_provider.py
daniellekain: I think it's a trade-off: keeping the list in code means coverage changes get review, and it's what lets the admin's rename/delete warning discover which polygons code depends on. 

Maybe the cleanest option is probably a ForeignKey to Polygon?

src/django/api/partner_fields/india_labour_line_provider.py
daniellekain: I'd drafted a reply here on first review but it never got posted -- apologies!

My concern with checking containment in code is that the boundary can be megabytes of geometry, and _fetch_raw_data runs on every page render of a covered Indian facility — so each render now fetches and parses the full boundary to answer a yes/no about one point.

(The polygon link still allows doing the check in the database in one query: filter partner fields on polygon__geom__contains=production_location.location and use .exists() — the boundary never has to leave the database.)

@protsack-stephan lmk your thoughts!

src/django/api/migrations/0230_setup_india_labour_line_partner_field.py
```diff
@@ -0,0 +1,84 @@
...
+    polygon_model = apps.get_model('api', 'Polygon')
+
+    field, _ = partner_field_model.objects.get_or_create(
+        name=FIELD_NAME,
+        defaults={'type': 'object'},
+    )
+    field.type = 'object'
+    field.label = 'India Labour Line Helpline'
+    field.json_schema = JSON_SCHEMA
+    field.display_text = HELPLINE_NUMBER
```
daniellekain, line 47: it looks like this over-write is safe (same value here as in prod). One optional hardening for the future would be to only set display_text when it's blank (the same courtesy the migration already extends to the polygon link), so if someone changes it on the admin page we don't over-write it here on the next deploy

src/django/api/partner_fields/india_labour_line_provider.py
```diff
@@ -0,0 +1,266 @@
...
+        case). Callers must treat None as "match nothing", never as
+        "match everything".
+    """
+    polygon = get_india_labour_line_polygon()
+    if polygon is None:
+        return None
+    return (
+        Q(location__within=polygon.geom)
+        & Q(sector__overlap=INDIA_LABOUR_LINE_SECTORS)
+    )
```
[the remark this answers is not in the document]
emilyoxford-osh: Fixed in 5e883502 — the country check is now part of the shared covered-locations rule, so the search filter, spotlight, and location page all apply the same three conditions (in India, inside the boundary, covered sector), with a regression test for an in-boundary non-Indian location.

daniellekain (approved):
looks good to me

src/django/api/partner_fields/india_labour_line_provider.py
emilyoxford-osh: I think it's okay to leave as-is for the time being given the turnaround and the fact that it appears to be working, however it should be considered as a future change. I do think that having it in the code is one way to make sure this isn't changed unnecessarily by accident, but there's more than one way to do that.

src/django/api/partner_fields/india_labour_line_provider.py
emilyoxford-osh: I'm inclined to agree with Danielle here - but we can talk about as a future possible change!

src/django/api/partner_fields/india_labour_line_provider.py
emilyoxford-osh: @protsack-stephan apologies I did not get to this, I'll admit I'm trying to get this in before the code freeze and I don't think the tests will finish running before it kicks in! however I can add this as a small improvement for the next release - clearly I need some practice doing PRs in a timely fashion anyway 😅

PR #1189, opened 2026-08-10
[Do not merge][OSDEV-3189] Add hierarchical taxonomy search with ISIC filtering and counts

VadimKovalenkoSNF:
## Jira Ticket

[OSDEV-3189](https://opensupplyhub.atlassian.net/browse/OSDEV-3189)

---

## Summary of Changes

Add hierarchical ISIC filtering and indexed processing-type suggestions to facility search.

The previous facility/processing taxonomy browser could only suggest the small static taxonomy even though facilities contain thousands of contributor-supplied processing values. This change keeps Facility Type as its own dropdown and adds a processing-type typeahead backed by the values already present in `FacilityIndex`. Suggestions are ranked by match quality and location count, show parent facility types and counts, place standard taxonomy terms before contributor values, and dim—but do not hide—values outside the selected facility types. Common placeholder values are excluded from both indexing and suggestions.

A trigger-maintained `api_facility_processing_value` table avoids aggregating the 12 GB facility index on every request. `GET /api/processing-type-suggestions/` queries that table, while migration `0229` and `recompute_facility_processing_value_index` create, backfill, and repair it. Facility and processing filters also support accent-insensitive whole-word and word-prefix free-text matching, with exact matches ranked first and stable secondary sorting.

The main search page also gains an ISIC Rev 4 hierarchical typeahead with indexed section, division, group, and class filters and per-node counts. Users can require facility/processing criteria and ISIC criteria to both match. Embedded maps retain their existing dropdowns.

Staff can upload, validate, preview, publish, enable, and disable ISIC taxonomies from Django admin. Published JSON remains in the private files bucket and is served through Django; the React app reads runtime configuration from the API and hides ISIC controls when no taxonomy is enabled.

---

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Backend change (API, database, business logic)
- [x] UI / frontend change
- [x] Architectural change (affects structure, contracts, or shared modules)
- [ ] Refactor (no functional change)
- [ ] Breaking change (existing functionality may not work as before)
- [ ] Documentation / tooling only

---

## Testing

**Latest processing-type regression run**
- `docker exec open-supply-hub-django-1 sh -c "python3 manage.py test api.tests.test_processing_type_suggestions_api api.tests.test_facility_and_processing_type_api"` — 39 tests passed; Django system checks reported no issues.
[...]

PR #1217, opened 2026-08-21
[OSDEV-3189] Add Facility Type and Processing Type taxonomy search

VadimKovalenkoSNF:
## Jira Ticket

[OSDEV-3189](https://opensupplyhub.atlassian.net/browse/OSDEV-3189)

---

## Summary of Changes

Add Facility Type and Processing Type taxonomy search to the facility map.

The existing filters exposed only static taxonomy options and could not suggest the thousands of contributor-submitted Processing Type values already indexed on facilities. This change adds a ranked autocomplete endpoint and sidebar control that search taxonomy and contributed values, show distinct-location counts and parent Facility Types, support keyboard and multi-value selection, and persist exact selections in URLs.

A trigger-maintained aggregate avoids scanning the 12 GB facility index for every typeahead request. Facility list, download, and tile paths share the same filtering behavior. Case-only variants such as `CAPS`, `Caps`, and `caps` are combined under a deterministic label and count each facility once; exact selections match whole array values case-insensitively while preserving punctuation, accent, and whitespace distinctions.

Facility filtering is index-assisted rather than expanding arrays for every one of the 2.5M locations. Exact selection improved from 2,679 ms to 2 ms, free-text search from 3,913 ms to 250 ms, and taxonomy filtering to under 1 ms. The typeahead now uses a composite normalized trigram index (a typical query improved from 3.5 ms to 0.7 ms), and relevance scoring evaluates both tiers in one pass.

The migrations create indexes concurrently and release the trigger-creation lock before rebuilding aggregates. The aggregate rebuild unnests `api_facilityindex` once instead of scanning its 12 GB table twice. `index_facility_type()` and `index_processing_type()` also scan visible ExtendedFields once and use positional array lookup instead of a cartesian expansion; the heaviest measured location improved from 20.2 ms to 8.4 ms. Their output now intentionally includes unmatched raw values, retains only profile-visible values from approved claims or active matches and sources, and excludes placeholder and legacy junk values.

Deployment now runs the scoped `facility_type,processing_type` facility-index backfill after migrations so historical unmatched values are available to search. Its runtime and database load will be validated during the Preproduction release before Production rollout, and the one-time step will be removed after all environments complete it.

After merging `main`, migration-number conflicts were resolved by renumbering the processing-search chain and its SQL assets to `0231`–`0233`, with an explicit dependency on the merged `0230` migration. Duplicate Schema changes and What’s new sections introduced while resolving the release-note conflict were also consolidated.

Follow-up refinements label standard results as `OS Hub Taxonomy`, retain multi-parent Facility Type breadcrumbs without a redundant parent-count badge, and implement complete combobox/listbox semantics for assistive technology. Suggestion requests now carry an identity derived from both the query and selected Facility Types so stale responses cannot replace results for a newer filter combination. Review follow-ups preserve numeric-looking Processing Type URL values as text, commit pending homepage free text on Find, Apply, and drawer close, and balance positive-count empty-query suggestions between taxonomy and contributor values. The response cache uses a stable semantic namespace.

This implementation is based on the Facility Type / Processing Type work from [PR #1189](https://github.com/opensupplyhub/open-supply-hub/pull/1189), narrowed to this ticket's current scope and updated for casing-standardization review feedback. The ISIC and taxonomy-administration work from that PR is not included.

### Flow

1. A contributor submits a new Processing Type.
2. Depending on the submission path:
   - A facility claim stores it until the claim is approved.
   - `POST /api/v1/production-locations/` stores it in a pending `ModerationEvent`.
   - `POST /api/facilities/` creates a facility-list item and sends it through matching.
3. When accepted, it is stored as an `ExtendedField`.
4. Taxonomy matches use the canonical taxonomy value; **unmatched values retain the trimmed submitted value**.
5. Once the contribution is profile-visible and facility indexing runs, the existing `api_facilityindex.processing_type` array is updated.
6. Database triggers update the suggestion aggregates, making the value available in autocomplete with a distinct-location count.

No additional `api_facilityindex` row is created.

Pending, inactive, blank, placeholder, and excluded legacy values are not indexed.

---

## Type of Change

- [x] UI / frontend change
- [x] Refactor (no functional change)

---

## Testing

- Verified the post-merge migration graph has one API leaf (`0230_index_facility_processing_search`) and a linear dependency chain from the merged `0227` migration through `0228`–`0230`.
- Updated the local database by applying the merged `0227` migration and safely recording the renamed processing migrations that had already run under their old names; `python3 manage.py migrate --plan` reports no pending operations.
- Compared the previous and optimized indexing functions value-for-value across 2,400 locations, including the 400 with the most ExtendedFields: zero facility-type or processing-type mismatches.
- `docker exec open-supply-hub-django-1 sh -c "python3 manage.py test api.tests.test_facility_and_processing_type api.tests.test_facility_and_processing_type_api api.tests.test_index_facilities api.tests.test_processing_type_suggestions_api api.tests.test_facility_processing_query"` — 77 tests passed.
- Contribution, claim, download, and facility-search regression modules — 72 tests passed. Combined focused backend coverage: 149 tests passed.
- `docker compose exec react yarn test --watchAll=false src/__tests__/components/ProcessingTypeSearch.test.js src/__tests__/utils.tests.js` — 211 tests passed.
- `docker compose exec react yarn test --watchAll=false ProcessingTypeSearch.test.js` — 23 tests passed, including combobox/listbox semantics and keyboard selection.
- `docker compose exec react yarn test --watchAll=false filterOptions.processingTypeSuggestions.test.js` — 6 tests passed, including stale completion and failure responses for changed Facility Types.
- Focused review-regression coverage — 32 Django tests and 217 React tests passed, covering numeric URL hydration, homepage pending-text submission, balanced empty suggestions, zero-count handling, and stable response caching.
- Targeted ESLint and backend lint checks passed; `git diff --check` reported no errors.
- Live local API/UI verification confirmed populated HTTP 200 suggestion responses. Real `Packing`, `packing`, and `PACKING` rows collapse into one `Packing` suggestion with the correct 6,132 distinct-location count, and case variants return the same exact-filter results.
- PostgreSQL `EXPLAIN ANALYZE` confirmed exact, taxonomy, free-text, and typeahead queries use the intended GIN/trigram indexes.
- The `post_deployment` backfill was not run against a production-scale dataset locally; runtime and database load will be monitored during the Preproduction release.
- The complete Django and React test suites, responsive/cross-browser testing, and a UI recording were not run.

---

## Known TODO Items

- Approval of a newly created v1 production location may not perform the final Processing Type reindex because of trigger ordering. That path requires a fix and integration test.

---

### Implementation

### Testing & Validation

### Final Review

[OSDEV-3189]: https://opensupplyhub.atlassian.net/browse/OSDEV-3189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/react/src/reducers/FilterOptionsReducer.js
```diff
@@ -274,6 +283,44 @@ export default createReducer(
...
+            }
+
+            return update(state, {
+                processingTypeSuggestions: {
+                    fetching: { $set: false },
+                    error: { $set: null },
+                    data: { $set: data },
+                },
+            });
+        },
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Check fixes.

src/react/src/components/Filters/HierarchicalTaxonomySearch/TaxonomySearchControl.jsx
```diff
@@ -0,0 +1,110 @@
...
+                    type="text"
+                    className={classes.searchInput}
+                    value={query}
+                    placeholder={showPlaceholder ? placeholder : ''}
+                    disabled={disabled}
+                    aria-controls={showResultsPanel ? listboxId : undefined}
+                    aria-autocomplete="list"
+                    aria-expanded={showResultsPanel && resultsCount > 0}
+                    aria-invalid={!!validationMessage}
+                    role="combobox"
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Check fixes.

src/django/api/management/commands/post_deployment.py
```diff
@@ -11,3 +11,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         call_command('migrate')
+        call_command(
+            'backfill_facility_index',
+            fields='facility_type,processing_type',
+        )
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: We need this command for upcoming release.

src/django/api/processing_type_search.py
```diff
@@ -0,0 +1,368 @@
...
+        api_facility_processing_value
+    WHERE
+        kind = %s
+        AND identity IN (
+            SELECT lower(taxonomy_value)
+            FROM unnest(%s::text[]) AS taxonomy(taxonomy_value)
+        )
+"""
+
+TOP_CONTRIBUTOR_VALUES_SQL = """
```
valencia-max: This value is the same as `TAXONOMY_VALUES_SQL` just adding:
```
ORDER BY
        facility_count DESC,
        value COLLATE "C" ASC
    LIMIT %s
```

If the query is not expected to change, can we reuse this value?

valencia-max:
Amazing job! This changes will add a lot of value to OS Hub!! 🚀 

Using Claude I found a couple of issues that we should review:

1. Searching for `http://localhost:6543/facilities?processing_type=300` and reloading the page, the app is crashing:

```
option.value.toLowerCase is not a function ::::::: TypeError: option.value.toLowerCase is not a function
    at http://localhost:6543/src/util/util.js:286:47
    at Array.reduce (<anonymous>)
    at createFiltersFromQueryString (http://localhost:6543/src/util/util.js:285:80)
    at http://localhost:6543/src/actions/filters.js:39:19
    at http://localhost:6543/node_modules/.vite/deps/redux-thunk.js?v=f3cee626:7:46
    at hydrateFiltersFromQueryString (http://localhost:6543/src/util/withQueryStringSync.jsx:108:5)
    at componentWithWrapper.componentDidMount (http://localhost:6543/src/util/withQueryStringSync.jsx:26:5)
    at commitLifeCycles (http://localhost:6543/node_modules/.vite/deps/react-dom-BTe0TO1U.js?v=f3cee626:12843:16)
    at commitLayoutEffects (http://localhost:6543/node_modules/.vite/deps/react-dom-BTe0TO1U.js?v=f3cee626:14295:6)
    at HTMLUnknownElement.callCallback (http://localhost:6543/node_modules/.vite/deps/react-dom-BTe0TO1U.js?v=f3cee626:969:11)
```
**Suggested fix**
String(option.value).toLowerCase() at util.js:557, at :1018 (restoreExactProcessingTypeLabels has the same assumption, currently shielded by the isExact early-out), and in caseIdentity at ProcessingTypeSearch.jsx:13. Better still, stop mapParamToReactSelectOption number-coercing free-text filter values at all — that coercion is a footgun for every text filter, not just this one.

2. The rewritten index functions narrow what gets indexed, and the release note says they don't
The previous definitions (0130_index_facility_type.sql, 0220_index_processing_type.sql) had no visibility predicate and no exclusion list — they indexed matched_values[2] from every api_extendedfield row of the location. The new ones add both: a profile-visibility predicate (approved claim, or an active match on an active source) and excluded_values = {null, none, n/a, na, unknown, other, -, denim services, boarding}.
That cannot hold for any location whose facility- or processing-type ExtendedFields come from a deactivated source, an inactive match, or a non-approved claim. Those values were indexed and will be deleted from api_facilityindex as soon as the backfill touches the row. The 2,400-location sample presumably just didn't contain one.
A location whose only processing_type came from a list whose source was later deactivated currently matches ?processing_type=Knitting. After the backfill it silently drops out of that filter, out of downloads and out of tiles. Nobody goes looking, because the release note says nothing changed.

**Suggested fix**
Measure it before Preproduction, then rewrite the note as a deliberate scope narrowing plus placeholder filtering, with the affected row count stated. If the number is large this needs a product decision, not just a note.

3. post_deployment runs backfill_facility_index with the triggers enabled and never calls the recompute. The result is correct — the deltas keep counts accurate — but every backfilled row pays roughly ten extra statements: four set-difference SELECT … INTOs, up to six DML statements against the aggregate tables, and the representative-label subquery, all funnelled through the hot aggregate rows, single-threaded. The release instructions omit the recompute entirely, so the new management command is currently unreachable from any automated path.

**Suggested fix**
Pick one story and make code and docs agree. Either drop the trigger-disable advice and accept the per-row cost (and say so in the file), or have the runbook disable the three facility_index_processing_value_* triggers, run the backfill, then run the recompute.

And also 2 things could be a follow up:

1. Free text typed on the homepage is silently discarded: Homepage → expand "Find facilities" → type cement mixing into Processing Type (the panel even prompts "Press Search to match 'cement mixing' as free text") → click Apply Filters or Find Facilities. The text vanishes and the search runs with no processing-type filter. The three-character validation is lost on this path too. Pressing Enter does work, which is exactly what will make this look intermittent in QA.

**Suggested fix**
Forward a ref from HomepageSidebarSearch the same way FilterSidebarSearchTab does, and call commitPendingQuery() from both buttons and on drawer close.

2. The user clicks into Processing Type without typing and gets 40 taxonomy values, several with count: 0. A contributor value carried by 50,000 locations is never shown, at any allowed limit, and the "Contributor values" heading renders an empty group. Reproduced with a stub cursor: at limit=40 the last row is Garment Dyeing; at limit=100 it's Shredding. Zero contributor rows either way. TOP_CONTRIBUTOR_VALUES_SQL runs on every one of these requests and its rows are discarded.

**Suggested fix**
Quota the empty-query path per category instead of sort-then-truncate — say top 20 taxonomy by count plus top 20 contributor by count — or return the two groups as separate lists and let the client render both. Worth also suppressing count: 0 taxonomy values from the empty state: a suggestion that leads to zero results is worse than no suggestion.

src/django/api/processing_type_search.py
VadimKovalenkoSNF: These queries also differ by **IN** versus **NOT IN**, so the top-contributor query isn’t just the taxonomy query with ordering and a limit appended. I’d prefer to keep each SQL statement explicit and independently readable.

VadimKovalenkoSNF:
Hi @valencia-max ! Thank you for the review. I've addressed issues mentioned in https://github.com/opensupplyhub/open-supply-hub/pull/1217#issuecomment-5400766273

src/django/api/facility_processing_query.py
```diff
@@ -0,0 +1,45 @@
...
+            FacilitiesQueryParams.PROCESSING_TYPE_EXACT,
+        )
+        if value.lower() in processing_type_identities
+    ]
+
+    return FacilityProcessingQuery(
+        facility_types,
+        processing_types,
+        exact_processing_types,
+    )
```
protsack-stephan: I would move this to a separate file, and make it a class that is easier to test and interpret, create a directory inside the project instead of putting it into the root and consider the following implementation:

```python
from typing import List, NamedTuple, Protocol

from api.constants import FacilitiesQueryParams


class ParamsProtocol(Protocol):
    """Protocol matching Django/Flask QueryDict interfaces."""
    def getlist(self, key: str) -> List[str]:
        ...


class FacilityProcessingResult(NamedTuple):
    facility_types: list[str]
    processing_types: list[str]
    exact_processing_types: list[str]


class FacilityProcessingQuery:
    def __init__(self, params: ParamsProtocol) -> None:
        self.params = params

    def _nonblank_values(self, key: str) -> list[str]:
        return [
            value
            for value in self.params.getlist(key)
            if value is not None and value.strip()
        ]

    def parse(self) -> FacilityProcessingResult:
        facility_types = self._nonblank_values(
            FacilitiesQueryParams.FACILITY_TYPE
        )
        processing_types = self._nonblank_values(
            FacilitiesQueryParams.PROCESSING_TYPE
        )
        processing_type_identities = {
            value.lower() for value in processing_types
        }
        exact_processing_types = [
            value
            for value in self._nonblank_values(
                FacilitiesQueryParams.PROCESSING_TYPE_EXACT
            )
            if value.lower() in processing_type_identities
        ]

        return FacilityProcessingResult(
            facility_types=facility_types,
            processing_types=processing_types,
            exact_processing_types=exact_processing_types,
        )
```

src/django/api/processing_type_search.py
```diff
@@ -0,0 +1,408 @@
...
+
+
+def get_processing_type_parents(value):
+    """Return the facility types a processing type belongs to."""
+    if not value:
+        return ()
+
+    canonical = _TAXONOMY_BY_KEY.get(_grouping_key(value))
+
+    return canonical[1] if canonical else ()
```
protsack-stephan: It looks like this function is not being used anywhere, so we should be able to delete it.

src/django/api/processing_type_search.py
```diff
@@ -0,0 +1,408 @@
...
+_TAXONOMY_VALUES = sorted(PROCESSING_TYPE_PARENTS)
+
+_FACILITY_TYPE_BY_KEY = {
+    clean(facility_type): facility_type
+    for facility_type in ALL_FACILITY_TYPES.values()
+}
+
+
+def _grouping_key(value):
+    return value.lower()
```
protsack-stephan: It looks like this function is not being used anywhere, so we should be able to delete it.

src/django/api/processing_type_search.py
```diff
@@ -0,0 +1,408 @@
...
+        rows.sort(key=lambda row: (
+            -row['score'],
+            -row['count'],
+            row['value'],
+        ))
+
+    return [
+        {field: value for field, value in row.items() if field != 'score'}
+        for row in rows[:limit]
+    ]
```
protsack-stephan: This ranking logic is a self-contained suggestions builder, but it currently lives as a bag of module-level functions in `api/processing_type_search.py`. That package root is already crowded with models, views, and one-off helpers; this belongs with the other request-scoped services under `api/services/`.

I suggest moving `api/services/processing_type_search.py`. Keep scores, regexes, and the import-time taxonomy maps at module level, fold the pipeline into a class whose constructor holds the query and whose `build_suggestions()` method is what the view calls:

```python
pt_search = ProcessingTypeSearch(
    query=request.query_params.get(QUERY_PARAM, ''),
    facility_types=request.query_params.getlist(FACILITY_TYPE_PARAM),
    limit=self.__parse_limit(...),
)
suggestions = pt_search.build_suggestions()
return Response(suggestions)
```

The raw SQL should move to the ORM as well. `api_facility_processing_value` is created by migration 0232, so add an unmanaged model (Django 5.2 composite PK fits the real key) rather than querying through `connection.cursor()`:

```python
# api/models/facility/facility_processing_value.py
class FacilityProcessingValue(models.Model):
    pk = models.CompositePrimaryKey('kind', 'identity')
    kind = models.TextField()
    identity = models.TextField()
    value = models.TextField()
    facility_count = models.IntegerField()

    class Meta:
        managed = False
        db_table = 'api_facility_processing_value'
```

The typed-query predicate can stay on `immutable_unaccent(lower(value))`. That is the expression the GIN index from 0233 is built on; `value__icontains=` would seq-scan the table. Wrap both sides in a `Func` so the generated SQL still matches the index, and keep `COLLATE "C"` on the empty-query contributor sort via `Collate`.

Sketch of the service:

```python
from collections.abc import Iterable
from typing import TypedDict

from django.db.models import BooleanField, F, Func, TextField, Value
from django.db.models.functions import Collate, Lower

from api.models.facility.facility_processing_value import (
    FacilityProcessingValue,
)


class ImmutableUnaccent(Func):
    function = 'immutable_unaccent'
    output_field = TextField()


class Like(Func):
    arity = 2
    arg_joiner = ' LIKE '
    template = '%(expressions)s'
    output_field = BooleanField()


class CandidateGroup(TypedDict):
    value: str
    count: int
    score: int


class ScoredSuggestion(TypedDict):
    value: str
    label: str
    count: int
    in_taxonomy: bool
    facility_types: list[str]
    dim: bool
    score: int


class ProcessingTypeSuggestion(TypedDict):
    value: str
    label: str
    count: int
    in_taxonomy: bool
    facility_types: list[str]
    dim: bool


CandidateRow = tuple[str, str, int]  # identity, value, facility_count


class ProcessingTypeSearch:
    """Rank processing type suggestions for the filter typeahead."""

    def __init__(
        self,
        query: str = '',
        facility_types: Iterable[str] | None = None,
        limit: int = DEFAULT_SUGGESTION_LIMIT,
    ) -> None:
        self._limit = min(max(int(limit), 0), MAX_SUGGESTION_LIMIT)
        self._normalized_query = _normalize(query) if query else ''
        self._selected_facility_types = self._resolve_facility_types(
            facility_types,
        )
        self._is_empty_query = not self._normalized_query

    def build_suggestions(self) -> list[ProcessingTypeSuggestion]:
        """
        Return ranked suggestions for the current query.

        An empty query balances positive-count taxonomy and contributor
        values. Selected facility types boost matching rows and dim the
        rest; nothing is filtered out for a typed query.
        """
        if self._limit == 0:
            return []

        if self._is_empty_query:
            candidates = self._fetch_top_values_by_count()
        else:
            candidates = self._fetch_matching_values()

        grouped = self._group_candidates(
            [*candidates, *self._taxonomy_candidates()],
        )
        rows = [
            self._build_row(key, group)
            for key, group in grouped.items()
        ]

        if self._is_empty_query:
            rows.sort(key=lambda row: (
                -row['score'],
                not row['in_taxonomy'],
                -row['count'],
                row['value'],
            ))
            rows = self._select_empty_query_rows(rows)
        else:
            rows.sort(key=lambda row: (
                -row['score'],
                -row['count'],
                row['value'],
            ))

        return [
            {
                field: value
                for field, value in row.items()
                if field != 'score'
            }
            for row in rows[:self._limit]
        ]

    def _fetch_matching_values(self) -> list[CandidateRow]:
        """Fetch values whose unaccented form contains the query."""
        return list(
            FacilityProcessingValue.objects.filter(
                kind=PROCESSING_TYPE_KIND,
            )
            .alias(
                matches=Like(
                    ImmutableUnaccent(Lower(F('value'))),
                    ImmutableUnaccent(
                        Lower(Value(_like_pattern(self._normalized_query))),
                    ),
                ),
            )
            .filter(matches=True)
            .values_list('identity', 'value', 'facility_count')
        )

    def _fetch_top_values_by_count(self) -> list[CandidateRow]:
        """
        Fetch candidates for an empty query.

        Every taxonomy value is fetched, not just the most common ones: a
        selected facility type can promote any of its children.
        """
        taxonomy_identities = [value.lower() for value in _TAXONOMY_VALUES]
        queryset = FacilityProcessingValue.objects.filter(
            kind=PROCESSING_TYPE_KIND,
        )
        taxonomy_rows = queryset.filter(
            identity__in=taxonomy_identities,
        ).values_list('identity', 'value', 'facility_count')
        contributor_rows = (
            queryset.exclude(identity__in=taxonomy_identities)
            .order_by('-facility_count', Collate('value', 'C'))
            [:self._limit]
            .values_list('identity', 'value', 'facility_count')
        )
        return [*taxonomy_rows, *contributor_rows]

    def _taxonomy_candidates(self) -> list[CandidateRow]:
        """
        Supply taxonomy labels even when no locations are indexed under
        them yet. Empty-query ranking drops those zero-count rows later.
        """
        return [
            (processing_type.lower(), processing_type, 0)
            for processing_type in PROCESSING_TYPE_PARENTS
        ]

    def _group_candidates(
        self,
        candidates: Iterable[CandidateRow],
    ) -> dict[str, CandidateGroup]:
        """Dedupe candidates by their PostgreSQL lower-case identity."""
        grouped: dict[str, CandidateGroup] = {}
        normalized_query = (
            None if self._is_empty_query else self._normalized_query
        )

        for identity, value, count in candidates:
            if not value or not identity:
                continue

            score = (
                0 if normalized_query is None
                else _match_score(_normalize(value), normalized_query)
            )
            if score is None:
                continue

            group = grouped.get(identity)
            if group is None:
                grouped[identity] = {
                    'value': value,
                    'count': count,
                    'score': score,
                }
                continue

            group['count'] = max(group['count'], count)
            group['score'] = max(group['score'], score)

        return grouped

    def _build_row(
        self,
        key: str,
        group: CandidateGroup,
    ) -> ScoredSuggestion:
        """Attach taxonomy metadata, facility-type boost, and dim flag."""
        canonical = _TAXONOMY_BY_KEY.get(key)
        facility_types = canonical[1] if canonical else ()
        value = canonical[0] if canonical else group['value']
        matches_selection = any(
            facility_type in self._selected_facility_types
            for facility_type in facility_types
        )
        score = group['score']
        if matches_selection:
            score += FACILITY_TYPE_MATCH_BOOST

        return {
            'value': value,
            'label': value,
            'count': group['count'],
            'in_taxonomy': canonical is not None,
            'facility_types': list(facility_types),
            'dim': (
                bool(self._selected_facility_types) and not matches_selection
            ),
            'score': score,
        }

    def _select_empty_query_rows(
        self,
        rows: list[ScoredSuggestion],
    ) -> list[ScoredSuggestion]:
        """
        Apply the empty-query taxonomy/contributor quota.

        Taxonomy receives half the slots rounded up and contributor values
        receive half rounded down. Unused slots are backfilled from the
        other group.
        """
        positive_count_rows = [row for row in rows if row['count'] > 0]
        taxonomy_rows = [
            row for row in positive_count_rows if row['in_taxonomy']
        ]
        contributor_rows = [
            row for row in positive_count_rows if not row['in_taxonomy']
        ]

        taxonomy_quota = (self._limit + 1) // 2
        contributor_quota = self._limit // 2
        selected = [
            *taxonomy_rows[:taxonomy_quota],
            *contributor_rows[:contributor_quota],
        ]

        if len(selected) < self._limit:
            selected.extend([
                *taxonomy_rows[taxonomy_quota:],
                *contributor_rows[contributor_quota:],
            ][:self._limit - len(selected)])

        selected_row_ids = {id(row) for row in selected}
        return [
            row for row in positive_count_rows
            if id(row) in selected_row_ids
        ][:self._limit]

    def _resolve_facility_types(
        self,
        facility_types: Iterable[str] | None,
    ) -> set[str]:
        """Map requested facility types to their taxonomy labels."""
        resolved: set[str] = set()
        for facility_type in facility_types or []:
            if not facility_type:
                continue
            label = _FACILITY_TYPE_BY_KEY.get(clean(facility_type))
            if label is not None:
                resolved.add(label)
        return resolved
```

We can leav `_normalize`, `_like_pattern`, and `_match_score` as module-level helpers.

src/django/api/views/facility/processing_type_suggestions.py
```diff
@@ -0,0 +1,127 @@
...
+                    "dim": true
+                }
+            ]
+
+        """
+        return Response(search_processing_types(
+            query=request.query_params.get(QUERY_PARAM, ''),
+            facility_types=request.query_params.getlist(FACILITY_TYPE_PARAM),
+            limit=self.__parse_limit(request.query_params.get(LIMIT_PARAM)),
+        ))
```
protsack-stephan: Can we add a serializer here? We should really protect this endpoint against someone setting a large limit, as it is open to the outside world without any authentication.

protsack-stephan (commented):
Good work, couple of comments from me.

src/django/api/facility_processing_query.py
VadimKovalenkoSNF: I moved the parser to `api/services/facility_processing_query.py`, converted it into a typed `FacilityProcessingQuery` class, and updated its callers and tests.

src/django/api/processing_type_search.py
VadimKovalenkoSNF: Confirmed that this function had no callers and removed.

src/django/api/processing_type_search.py
VadimKovalenkoSNF: Thanks, I removed it.

src/django/api/views/facility/processing_type_suggestions.py
VadimKovalenkoSNF: Updated. The endpoint now validates its query parameters through a DRF serializer. Invalid and negative limits return 400, oversized limits remain capped at `100` as documented, and blank limits retain the existing default behavior.

src/django/api/processing_type_search.py
VadimKovalenkoSNF: I moved the ranking pipeline to `api/services/processing_type_search.py` as a request-scoped `ProcessingTypeSearch` class and replaced the raw SQL with an unmanaged `FacilityProcessingValue` model using Django’s composite primary key support. Migration 0232 now tracks the model state without duplicating database creation.

src/django/api/models/facility/facility_manager_index_new.py
```diff
@@ -11,6 +14,332 @@
...
+
+    return queryset.annotate(
+        **{
+            annotation_name: RawSQL(
+                sql,
+                params,
+                output_field=IntegerField(),
+            )
+        }
+    ), True
```
protsack-stephan: **Nit / suggestion:** These helpers (classify → match SQL → relevance SQL → annotate) look like one object, not six functions sharing the same three lists. Classification also runs twice, and `exact_processing_types` is intersected here even though `FacilityProcessingQuery.parse()` already does that.

Please extract something like `FacilityProcessingFilter` that takes the parsed lists (or `FacilityProcessingResult`), classifies once, and exposes match/relevance/annotate. Leave parsing on `FacilityProcessingQuery`. Don’t fold this into `FacilityIndexNewManager` — both `facility_manager.py` and `facilities_view_set.py` need the same annotations.

Example:

```python
class FacilityProcessingFilter:
    MIN_FREE_TEXT_LENGTH = 3

    def __init__(self, facility_types, processing_types, exact_processing_types=None):
        # classify once; split exact vs legacy processing types once
        ...

    @classmethod
    def from_result(cls, parsed):
        return cls(
            parsed.facility_types,
            parsed.processing_types,
            parsed.exact_processing_types,
        )

    def match_sql(self): ...
    def relevance_sql(self): ...
    def annotate_match(self, queryset, name="_fp_match"): ...
    def annotate_relevance(self, queryset, name="_fp_relevance"): ...
```

Call sites:

```python
fp = FacilityProcessingFilter.from_result(
    FacilityProcessingQuery(params).parse()
)
facilities_qs, has_fp_filter = fp.annotate_match(facilities_qs)
if has_fp_filter:
    facilities_qs = facilities_qs.filter(_fp_match=True)
```

A bag of `@staticmethod`s isn’t worth it; the win is holding classified terms on the instance.

**On Django ORM:** only the taxonomy overlap maps cleanly to `__overlap`. The rest has to keep hitting the expression indexes from migration 0233 (`facility_processing_search_text(...)` trigram, `lower_varchar_array(processing_type) && …`). Vanilla lookups would change matching and/or seq-scan `api_facilityindex`.

If you want an ORM-shaped API, wrap those **exact** fragments in `Func` / a custom `Lookup` — don’t rewrite the predicates:

```python
class FacilityProcessingSearchText(Func):
    function = "facility_processing_search_text"
    output_field = TextField()

    def __init__(self):
        super().__init__(F("facility_type"), F("processing_type"))


class LowerVarcharArray(Func):
    function = "lower_varchar_array"
    output_field = ArrayField(TextField())


# taxonomy
qs.filter(facility_type__overlap=taxonomy_values)

# exact processing (must use the expression, not processing_type__overlap)
qs.filter(LowerVarcharArray(F("processing_type")).overlap(lowered_values))

# free-text prefilter: same LIKE as today, so the trigram index still matches
qs.annotate(
    _fp_search=FacilityProcessingSearchText(),
).filter(_fp_search__contains=like_term)  # only if this still emits
# facility_processing_search_text(...) LIKE '%' || immutable_unaccent(lower(%s)) || '%'
```

`__contains` on a `TextField` will **not** wrap `immutable_unaccent(lower(...))`, so a custom lookup (or `RawSQL` inside `annotate_match`) is still required for the free-text path. Keep the tests that assert the generated SQL (`LIKE` + `immutable_unaccent`, `lower_varchar_array(...) &&`, LIKE escaping).

protsack-stephan (dismissed):
Great work, one additional comment. Would be good to implement, but not critical.

src/django/api/models/facility/facility_manager_index_new.py
VadimKovalenkoSNF: Addressed the review comment:

- Extracted FacilityProcessingFilter into the services layer.
- Classification and exact/legacy splitting now happen once.
- Shared the same instance between match and relevance annotation.
- Preserved indexed RawSQL predicates exactly.
- Updated both managers, the viewset, and SQL regression tests.
- Added a test proving classification is not repeated.

protsack-stephan (approved):

PR #1232, opened 2026-08-27
[Quickfix][OSDEV-3189] Restore default search sort after homepage navigation

VadimKovalenkoSNF:
## Jira Ticket

[OSDEV-3189](https://opensupplyhub.atlassian.net/browse/OSDEV-3189)

---

## Summary of Changes

Restore the default facility-search sort after homepage navigation.

The taxonomy release changed homepage search to rebuild the `/facilities` URL from Redux so a pending processing-type query can be committed before navigation. That URL omits `sort_by` because homepage Redux starts with an empty sort. The destination then hydrates filters through `getAlgorithm`, which fell back to `optionsForSortingResults[0]` (`name_asc` / A to Z) instead of the intended `# Contributors` descending sort.

This change makes `getAlgorithm` use `DEFAULT_SORT_OPTION_INDEX` (`contributors_desc`), matching the Sort By dropdown and the API default when `sort_by` is omitted. An explicit `sort_by=name_asc` is still honored. Embed mode still forces A to Z.

---

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Backend change (API, database, business logic)
- [x] UI / frontend change
- [ ] Architectural change (affects structure, contracts, or shared modules)
- [ ] Refactor (no functional change)
- [ ] Breaking change (existing functionality may not work as before)
- [ ] Documentation / tooling only

---

## Testing

- Added unit coverage that a missing or unknown `sort_by` hydrates as `contributors_desc`, while an explicit `sort_by=name_asc` stays `name_asc`.
- Updated `createFiltersFromQueryString` fixtures that assumed the old implicit A to Z default.
- Ran the related frontend suites: `utils.tests.js`, `filters.test.js`, `HomepageSidebarSearch.test.js`, and `embedModeNavigation.test.jsx` — 4 suites, 203 tests passed.
- Locally opened `/facilities?contributors=1` (no `sort_by`); the URL became `sort_by=contributors_desc`. Opening the same URL with `sort_by=name_asc` kept A to Z.
- No screenshot or cross-browser pass was recorded.

---

## Known TODO Items

None

---

### Implementation

- [x] My code follows the project's style guidelines and naming conventions.
- [x] I have removed debug logs, commented-out code, and unused imports.
- [ ] Any AI-assisted code (e.g., Claude) has been reviewed, understood, and adapted to fit our codebase.
- [x] I have considered backward compatibility and migrations where applicable.
[...]

PR #1240, opened 2026-08-28
[Backmerge][OSDEV-3189] Restore default search sort after homepage navigation

VadimKovalenkoSNF:
Backmerge of https://github.com/opensupplyhub/open-supply-hub/pull/1232

**Summary of Changes**

Restore the default facility-search sort after homepage navigation.

The taxonomy release changed homepage search to rebuild the /facilities URL from Redux so a pending processing-type query can be committed before navigation. That URL omits sort_by because homepage Redux starts with an empty sort. The destination then hydrates filters through getAlgorithm, which fell back to optionsForSortingResults[0] (name_asc / A to Z) instead of the intended # Contributors descending sort.

This change makes getAlgorithm use DEFAULT_SORT_OPTION_INDEX (contributors_desc), matching the Sort By dropdown and the API default when sort_by is omitted. An explicit sort_by=name_asc is still honored. Embed mode still forces A to Z.


vlad-shapik (dismissed):

protsack-stephan (approved):

VadimKovalenkoSNF:
@coderabbit resolve

PR #1257, opened 2026-08-31
OSDEV-3370: Platform API changes

valencia-max:
## Jira Ticket

[OSDEV-2278](https://opensupplyhub.atlassian.net/browse/OSDEV-2278) — subtasks [OSDEV-3370](https://opensupplyhub.atlassian.net/browse/OSDEV-3370) (Platform API) and [OSDEV-3371](https://opensupplyhub.atlassian.net/browse/OSDEV-3371) (UX + notification)

---

## Summary of Changes

Lets claimants view and edit their own PENDING claims from My Account → My Facilities — replacing the email-based document exchange that raised the PII/SOC 2 concern behind OSDEV-2278.

- `GET/PATCH /api/facility-claims/{id}/pending/` — claimant view/edit of their own PENDING claim. Guard: owning contributor + PENDING; everything else 404s (no claim-id existence leak). `EditPendingClaimSerializer` derives from `FacilityCreateClaimSerializer` so create and edit share one set of validation rules. Write paths lock the claim row (`select_for_update`) so concurrent attachment uploads cannot race past the lifetime file cap.
- **Root cause fix:** `FacilityClaimAttachments.claim` was `on_delete=PROTECT` while facility deletion calls `claim.delete()` — any claim with attachments raised `ProtectedError` and aborted the whole deletion. Migration 0234 moves it to CASCADE with a `post_delete` signal removing the stored file.
- Terraform: signing role + assume-role wiring, and an `s3:signatureAge` ≤ 15 min deny **scoped to the `claim_attachments/` prefix** (caps a leaked URL's usable life at the policy layer). Other presigned consumers — facility-list downloads, partner-field-group icons embedded in CloudFront-cached responses — are deliberately untouched.

---

## Type of Change

- [x] UI / frontend change

**Breaking change detail:** claim API payloads no longer carry `claim_attachment` (raw storage URL). The `automated-claims` Lambda consumes exactly that field via `requests.get(presigned_url)`, so it must move to direct S3 reads via IAM **before or together with** this deploy, or attachment processing breaks. The signature-age cap affects only `claim_attachments/*` objects; presigned URLs elsewhere in the bucket behave as before.

---

## Testing

- `api/tests/test_pending_claim_edit.py` — 28 tests, all passing:
  - Partial-update semantics; URL validation; lifetime 20-file cap (19+2 rejected, 19+1 accepted); magic-byte rejection (HTML-as-pdf); extension allowlist; opaque storage keys (original filename never in the key).
  - Download endpoint: 302 for owner and superuser, 404 stranger, 401 anonymous; moderator claim-details payload contains no storage URLs.
  - OSDEV-3371: PATCH stamps `claimant_updated_at` + sends the notification (subject/body/dashboard link asserted, zero attachments); no-op saves send nothing and stamp nothing; attachment add stamps + notifies; `?statuses=` default/param/invalid-400 behavior; `claimant_updated_at` present in the list payload.
- `makemigrations --check`: migrations 0234/0235 fully cover the model changes.

---

## Known TODO Items

- **`automated-claims` (blocking dependency, in OSDEV-2278 backlog):** read attachments directly via IAM instead of API presigned URLs; treat `updated_at > processed_at` as "reprocess" in the DynamoDB idempotency check (an edited claim is currently skipped — or was already auto-processed on stale evidence); clean up its S3/GCS copies when the platform disposes of documents.
- Legacy attachments remain at the bucket root, outside the signature-age cap (their 60-second URLs are the only guard); after migrating them under `claim_attachments/`, tighten the signer role's `s3:GetObject` to the prefix (noted in `iam.tf`).
- The first render of the v1 claim-flow step components outside the stepper; visual QA notes welcome.

---

### Implementation

### Testing & Validation

### Final Review

### DIFF
--- a/src/django/api/views/facility/facility_claim_view_set.py
+++ b/src/django/api/views/facility/facility_claim_view_set.py
@@ -14,6 +14,7 @@ from rest_framework.viewsets import ModelViewSet
 from django.contrib.gis.geos import GEOSGeometry
 from django.core.exceptions import FieldDoesNotExist
 from django.db import models, transaction
+from django.http import HttpResponseRedirect
 from django.utils import timezone
 from waffle import switch_is_active
 
@@ -32,9 +33,18 @@ from ...mail import (
     send_claim_update_notice_to_list_contributors,
     send_message_to_claimant_email,
 )
+from ...helpers.attachment_download import generate_attachment_download_url
+from ...helpers.claim_attachments import (
+    create_claim_attachment,
+    delete_claim_attachment,
+    validate_attachment_files,
+)
 from ...models.contributor.contributor import Contributor
 from ...models.extended_field import ExtendedField
 from ...models.facility.facility_claim import FacilityClaim
+from ...models.facility.facility_claim_attachments import (
+    FacilityClaimAttachments
+)
 from ...models.facility.facility_claim_review_note import (
     FacilityClaimReviewNote
 )
@@ -48,6 +58,10 @@ from ...serializers import (
     FacilityClaimSerializer,
     FacilityClaimDetailsSerializer,
     FacilityClaimListQueryParamsSerializer
+)
+from ...serializers.facility.edit_pending_claim_serializer import (
+    EditPendingClaimSerializer,
+    PendingClaimSerializer,
 )
 from ..make_report import _report_facility_claim_email_error_to_rollbar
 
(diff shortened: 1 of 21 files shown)


src/django/api/views/facility/facility_claim_view_set.py
```diff
@@ -638,3 +652,156 @@ def geocode_claim_address(self, request, pk=None):
...
+        lifetime total (submission plus edits), not per request.
+        """
+        if not switch_is_active('claim_a_facility'):
+            raise NotFound()
+
+        files = request.FILES.getlist('files')
+        if not files:
+            raise BadRequestException('No files submitted.')
+
+        claim = self.__get_owned_pending_claim(request, pk)
```
daniellekain: do we want to select_for_update the claim here (inside __get_owned_pending_claim)? atomic doesn't serialize the count-then-insert, so two concurrent posts can both read the same existing_count and land over the 20 cap. one row lock here would cover the PATCH path too

src/django/oar/settings.py
```diff
@@ -637,6 +637,21 @@
 
 AWS_S3_FILE_OVERWRITE = False
 
+# Presigned URLs generated through django-storages (e.g. facility list
+# file links in the moderation dashboard) stay valid for 5 minutes
+# instead of the 1-hour default. Claim attachment downloads do not use
+# this path: they go through the download endpoint on
+# FacilityClaimViewSet, which mints 60-second single-object URLs.
+AWS_QUERYSTRING_EXPIRE = int(os.getenv('AWS_QUERYSTRING_EXPIRE', '300'))
```
daniellekain: this would be a global change -- is there anywhere that we send links that we wouldn't want to expire? email or export?

daniellekain (commented):
looks great! A couple of quick questions

src/django/oar/settings.py
valencia-max: Good catch — the facility-list download and the partner-group icons (which ride inside a CloudFront-cached API response) both rely on longer-lived presigned URLs, so the global expiry override is removed and the signatureAge policy cap is scoped to `claim_attachments/*`. Claim docs are unaffected (60-second dedicated signer). No emails or exports carry storage URLs — verified.

src/django/api/views/facility/facility_claim_view_set.py
valencia-max: Good catch! Done!

src/django/oar/settings.py
daniellekain: great -- thanks for investigating!

daniellekain (approved):
looks good to me barring the code quality failures

PR #1260, opened 2026-08-31
OSDEV-3371: Platform UX Changes and Notification

valencia-max:
## Jira Ticket

[OSDEV-2278](https://opensupplyhub.atlassian.net/browse/OSDEV-2278) — subtasks [OSDEV-3370](https://opensupplyhub.atlassian.net/browse/OSDEV-3370) (Platform API) and [OSDEV-3371](https://opensupplyhub.atlassian.net/browse/OSDEV-3371) (UX + notification)

---

## Summary of Changes

Lets claimants view and edit their own PENDING claims from My Account → My Facilities — replacing the email-based document exchange that raised the PII/SOC 2 concern behind OSDEV-2278. Two commits, one per subtask.

- `GET/PATCH /api/facility-claims/{id}/pending/` — claimant view/edit of their own PENDING claim. Guard: owning contributor + PENDING; everything else 404s (no claim-id existence leak). `EditPendingClaimSerializer` derives from `FacilityCreateClaimSerializer` so create and edit share one set of validation rules.
- **Root cause fix:** `FacilityClaimAttachments.claim` was `on_delete=PROTECT` while facility deletion calls `claim.delete()` — any claim with attachments raised `ProtectedError` and aborted the whole deletion. Migration 0234 moves it to CASCADE with a `post_delete` signal removing the stored file.
- Terraform: signing role + assume-role wiring, `s3:signatureAge` ≤ 15 min deny on the files bucket (caps any presigned URL's usable life at the policy layer), `AWS_QUERYSTRING_EXPIRE=300` for remaining django-storages URLs.

**UX + notification (OSDEV-3371).**

- My Facilities (`/claimed`) now requests `?statuses=PENDING,APPROVED` (endpoint default stays approved-only for other consumers), shows a status chip per row, and forks the row click: pending → new `/claimed/pending/{id}` edit view; approved → existing profile editor, unchanged.
- `PendingClaimEdit` reuses the v1 claim flow's step components (`ContactInfoStep`, `BusinessStep`, `ProfileStep`) on a single page — fields, Yup validation, and the OSDEV-2280 PII warning are the create flow's own components, not copies. Own Redux slice + Formik; the API↔form field mapping lives in one file (`PendingClaimEdit/utils.js`). Uploaded documents render in a manager section (download via the new endpoint, delete with confirmation); newly picked files upload on Save, then field edits go as a JSON PATCH.
- The OSDEV-2280 warning copy is lifted into `CLAIM_PII_WARNING_TEXT` and shared by both claim-flow steps and the edit view (it was duplicated inline).
- `claim_updated_by_claimant` email to `NOTIFICATION_EMAIL_TO` on every claimant save: changed field names, document add/remove by filename, and a dashboard link — never a document or a URL to one.
- New `FacilityClaim.claimant_updated_at` (migration 0235), stamped **only** by claimant edits — moderator actions bump `updated_at` but not this — feeding a sortable "Updated by Claimant" column in the claims queue and its XLSX export, so claims edited after review starts don't hide in the "already reviewed" pile.

---

## Type of Change

- [x] UI / frontend change

**Breaking change detail:** claim API payloads no longer carry `claim_attachment` (raw storage URL). The `automated-claims` Lambda consumes exactly that field via `requests.get(presigned_url)`, so it must move to direct S3 reads via IAM **before or together with** this deploy, or attachment processing breaks. Additionally, any presigned URL older than 15 minutes now dies at the bucket policy — stale dashboard-tab links need a page refresh.

---

## Testing

- `api/tests/test_pending_claim_edit.py` — 28 tests, all passing:
  - Partial-update semantics; URL validation; lifetime 20-file cap (19+2 rejected, 19+1 accepted); magic-byte rejection (HTML-as-pdf); extension allowlist; opaque storage keys (original filename never in the key).
  - Download endpoint: 302 for owner and superuser, 404 stranger, 401 anonymous; moderator claim-details payload contains no storage URLs.
  - OSDEV-3371: PATCH stamps `claimant_updated_at` + sends the notification (subject/body/dashboard link asserted, zero attachments); no-op saves send nothing and stamp nothing; attachment add stamps + notifies; `?statuses=` default/param/invalid-400 behavior; `claimant_updated_at` present in the list payload.
- `makemigrations --check`: migrations 0234/0235 fully cover the model changes.

---

## Known TODO Items

- **`automated-claims` (blocking dependency, in OSDEV-2278 backlog):** read attachments directly via IAM instead of API presigned URLs; treat `updated_at > processed_at` as "reprocess" in the DynamoDB idempotency check (an edited claim is currently skipped — or was already auto-processed on stale evidence); clean up its S3/GCS copies when the platform disposes of documents.
- Legacy attachments remain at the bucket root; after migrating them under `claim_attachments/`, tighten the signer role's `s3:GetObject` to the prefix (noted in `iam.tf`).
- The first render of the v1 claim-flow step components outside the stepper; visual QA notes welcome.

---

### Implementation

### Testing & Validation

### Final Review

### DIFF
--- a/src/django/api/views/facility/facility_claim_view_set.py
+++ b/src/django/api/views/facility/facility_claim_view_set.py
@@ -670,12 +670,20 @@ class FacilityClaimViewSet(ModelViewSet):
     def __notify_claims_team_of_update(request, claim, changes):
         """
         AC #5: the Claims team is notified on every claimant save. Sent
-        best-effort — a mail failure must not roll back the edit.
+        best-effort — a mail failure must not roll back the edit — and
+        deferred with transaction.on_commit, so a save that ends up
+        rolled back sends nothing, and the claim row lock is not held
+        open across the SMTP round-trip.
         """
-        try:
-            send_claim_updated_by_claimant_notice(request, claim, changes)
-        except Exception:
-            _report_facility_claim_email_error_to_rollbar(claim)
+        def send():
+            try:
+                send_claim_updated_by_claimant_notice(
+                    request, claim, changes
+                )
+            except Exception:
+                _report_facility_claim_email_error_to_rollbar(claim)
+
+        transaction.on_commit(send)
 
     def __get_owned_pending_claim(self, request, pk, for_update=False):
         """
--- a/src/django/api/views/facility/facility_claim_view_set.py
+++ b/src/django/api/views/facility/facility_claim_view_set.py
@@ -31,6 +31,7 @@ from ...mail import (
     send_claim_facility_denial_email,
     send_claim_facility_revocation_email,
     send_claim_update_notice_to_list_contributors,
+    send_claim_updated_by_claimant_notice,
     send_message_to_claimant_email,
 )
 from ...helpers.attachment_download import generate_attachment_download_url
@@ -654,6 +655,28 @@ class FacilityClaimViewSet(ModelViewSet):
         geocode_result = geocode_address(address, country_code)
         return Response(geocode_result)
 
+    @staticmethod
+    def __stamp_claimant_update(claim):
+        """
+        Record that the claimant changed something on this claim.
+        claimant_updated_at is deliberately separate from updated_at,
+        which moderator actions also bump — the claims queue sorts on
+        this field to surface claims edited since review started.
+        """
+        claim.claimant_updated_at = timezone.now()
+        claim.save(update_fields=['claimant_updated_at', 'updated_at'])
+
+    @staticmethod
+    def __notify_claims_team_of_update(request, claim, changes):
+        """
+        AC #5: the Claims team is notified on every claimant save. Sent
+        best-effort — a mail failure must not roll back the edit.
+        """
+        try:
+            send_claim_updated_by_claimant_notice(request, claim, changes)
+        except Exception:
+            _report_facility_claim_email_error_to_rollbar(claim)
+
     def __get_owned_pending_claim(self, request, pk, for_update=False):
         """
         Fetch a claim the requesting user may edit: they are its
@@ -714,10 +737,17 @@ class FacilityClaimViewSet(ModelViewSet):
 
         changed_fields = serializer.apply_to_claim(claim)
         if changed_fields:
+            claim.claimant_updated_at = timezone.now()
             # simple-history records this save with history_user set to
             # the claimant via HistoryRequestMiddleware, which is what
             # lets moderators see exactly what the claimant changed.
             claim.save()
+            self.__notify_claims_team_of_update(
+                request,
+                claim,
+                ['Updated {}'.format(field)
+                 for field in sorted(changed_fields)],
+            )
 
         return Response(PendingClaimSerializer(claim).data)
 
@@ -746,8 +776,15 @@ class FacilityClaimViewSet(ModelViewSet):
         ).count()
         validate_attachment_files(files, existing_count=existing_count)
 
-        for file in files:
-            create_claim_attachment(file, claim)
+        created = [create_claim_attachment(file, claim) for file in files]
+
+        self.__stamp_claimant_update(claim)
+        self.__notify_claims_team_of_update(
+            request,
+            claim,
+            ['Added document: {}'.format(attachment.file_name)
+             for attachment in created],
+        )
 
         return Response(PendingClaimSerializer(claim).data)
 
@@ -775,7 +812,15 @@ class FacilityClaimViewSet(ModelViewSet):
         except FacilityClaimAttachments.DoesNotExist as exc:
             raise NotFound() from exc
 
+        deleted_file_name = attachment.file_name
         delete_claim_attachment(attachment)
+
+        self.__stamp_claimant_update(claim)
+        self.__notify_claims_team_of_update(
+            request,
+            claim,
+            ['Removed document: {}'.format(deleted_file_name)],
+        )
 
         return Response(PendingClaimSerializer(claim).data)
 
--- a/src/django/sqls/0232_facility_processing_value_index.sql
+++ b/src/django/sqls/0232_facility_processing_value_index.sql
@@ -18,8 +18,7 @@ the recompute_facility_processing_value_index management command after them
 instead, and disable the triggers below for the duration if the run is large
 enough for the counter updates to matter.
 */
-DROP TABLE IF EXISTS api_facility_processing_value;
-DROP TABLE IF EXISTS api_facility_processing_value_variant;
+DROP MATERIALIZED VIEW IF EXISTS api_facility_processing_value;
 
 CREATE TABLE api_facility_processing_value (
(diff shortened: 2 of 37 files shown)


daniellekain (commented):
Hi Max! This looks great. I can leave more in depth comments tomorrow, but two quick things that Claude found -- 

1. Failed save retries duplicate the uploaded documents (actions/pendingClaimEdit.js, savePendingClaim). Files upload first, then fields PATCH. If the PATCH fails validation, the files are already stored — but the pickers only clear on full success. User fixes the field, clicks Save again → the same files POST a second time → duplicate attachments on the claim, plus a second "documents added" email. Fix: clear the picker fields (or fold the upload response into state) as soon as the upload succeeds, before the PATCH.
2. Cleared values silently reappear (PendingClaimEdit/utils.js). Emptied numeric/date fields are omitted from the PATCH payload, and disabled emissions fields are dropped — but in a partial PATCH, omitted means "unchanged," so the old value survives. The emissions checkboxes then re-derive from stored values, so a claimant who unchecks coal and saves watches the checkbox turn itself back on. Either send null for cleared fields (needs allow_null on the #1257 serializer side) or disable clearing in the UI and say so.

src/django/sqls/0232_facility_processing_value_index.sql
```diff
@@ -18,7 +18,8 @@ the recompute_facility_processing_value_index management command after them
 instead, and disable the triggers below for the duration if the run is large
 enough for the counter updates to matter.
 */
-DROP MATERIALIZED VIEW IF EXISTS api_facility_processing_value;
+DROP TABLE IF EXISTS api_facility_processing_value;
```
daniellekain: should we mention this change in the description or split it into it's own PR?

src/django/api/views/facility/facility_claim_view_set.py
```diff
@@ -654,6 +654,28 @@ def geocode_claim_address(self, request, pk=None):
...
+        claim.save(update_fields=['claimant_updated_at', 'updated_at'])
+
+    @staticmethod
+    def __notify_claims_team_of_update(request, claim, changes):
+        """
+        AC #5: the Claims team is notified on every claimant save. Sent
+        best-effort — a mail failure must not roll back the edit.
+        """
+        try:
+            send_claim_updated_by_claimant_notice(request, claim, changes)
```
daniellekain: should we wrap this in transaction.on_commit()? two benefits: no email for a save that ends up rolled back, and the transaction isn't held open across the SMTP round-trip

src/django/api/views/facility/facility_claim_view_set.py
```diff
@@ -700,10 +722,17 @@ def pending(self, request, pk=None):
...
         if changed_fields:
+            claim.claimant_updated_at = timezone.now()
             # simple-history records this save with history_user set to
             # the claimant via HistoryRequestMiddleware, which is what
             # lets moderators see exactly what the claimant changed.
             claim.save()
+            self.__notify_claims_team_of_update(
+                request,
+                claim,
+                ['Updated {}'.format(field)
```
daniellekain, line 756: nit: we're exposing the internal field name here

src/react/src/components/PendingClaimEdit/PendingClaimAttachments.jsx
```diff
@@ -0,0 +1,143 @@
...
+                                >
+                                    {attachment.file_name}
+                                </a>
+                                <span className={classes.uploadedAt}>
+                                    {moment(attachment.uploaded_at).format(
+                                        'LL',
+                                    )}
+                                </span>
+                            </span>
+                            {deleting ? (
```
daniellekain, line 93: nit: I think there's a chance that we'll show the deleting spinner on all attachments, not just the one being deleted

src/django/sqls/0232_facility_processing_value_index.sql
valencia-max: My bad! I did for workaround, these changes are not of this PR.

src/django/api/views/facility/facility_claim_view_set.py
valencia-max: This is only for internal purposes, the moderators are the ones that receive this email. I'd be happy to modify it, but I think there's no problem.

src/react/src/components/PendingClaimEdit/PendingClaimAttachments.jsx
valencia-max: Actually it's working like this, If you try to remove an attachment, the spinner is displayed in all the rows:

https://github.com/user-attachments/assets/f2b340c7-502c-4c07-8fcb-bddc247ea02a

src/django/api/views/facility/facility_claim_view_set.py
valencia-max: Done!

valencia-max:
> Hi Max! This looks great. I can leave more in depth comments tomorrow, but two quick things that Claude found --
> 
> 1. Failed save retries duplicate the uploaded documents (actions/pendingClaimEdit.js, savePendingClaim). Files upload first, then fields PATCH. If the PATCH fails validation, the files are already stored — but the pickers only clear on full success. User fixes the field, clicks Save again → the same files POST a second time → duplicate attachments on the claim, plus a second "documents added" email. Fix: clear the picker fields (or fold the upload response into state) as soon as the upload succeeds, before the PATCH.
> 2. Cleared values silently reappear (PendingClaimEdit/utils.js). Emptied numeric/date fields are omitted from the PATCH payload, and disabled emissions fields are dropped — but in a partial PATCH, omitted means "unchanged," so the old value survives. The emissions checkboxes then re-derive from stored values, so a claimant who unchecks coal and saves watches the checkbox turn itself back on. Either send null for cleared fields (needs allow_null on the #1257 serializer side) or disable clearing in the UI and say so.

1. "Fixed by reordering instead: fields PATCH runs first, upload second. A failed PATCH uploads nothing, and a retried PATCH after an upload failure is a server-side no-op — so retries can't duplicate attachments and the user never has to re-pick files."
2. "Fixed with the null approach: the payload sends null for cleared fields and unchecked emissions sections, and the edit serializer allows null on that clearable set only."

src/django/api/views/facility/facility_claim_view_set.py
daniellekain: got it -- I missed this was internal only -- thanks!

src/react/src/components/PendingClaimEdit/PendingClaimAttachments.jsx
daniellekain: yep -- that was my read too. I wasn't sure if this was intended, or if it was meant to only spin the one that was being removed

daniellekain (approved):
Awesome work, Max! I'm so excited to see these changes in the product.

PR #1264, opened 2026-09-01
[OSDEV-3216] Fix three check crashes carried over into the ContriBot rebuild

daniellekain:
## Jira Ticket

[OSDEV-3216](https://opensupplyhub.atlassian.net/browse/OSDEV-3216)

---

## Summary of Changes

Three contributor file shapes were mishandled by the Airflow ContriBot. Because that DAG had no per-list isolation, each failure killed the whole hourly batch and the `MAX(list_id)` cursor then skipped every list in it permanently — eleven uploads were lost that way in early August.

The Step Functions rebuild fixed the *isolation*: failures are caught per list, recorded as `FAILED` in DynamoDB, surfaced in Slack, and retried. But the faults themselves were ported into `src/contribot/lib/contribot.py` unchanged. This fixes them.

**Verified against `main` before fixing.** Each regression test was run against unmodified `main` first, so they pin real behaviour rather than the fixed behaviour:

| Input | Behaviour on `main` today |
| --- | --- |
| File above 10,000 rows | Crashes: `AttributeError: 'ContriBot' object has no attribute 'df_config'` |
| File with no `address` column | Crashes: `AttributeError: 'DataFrame' object has no attribute 'address'` — kills `process()` |
| File with an unrecognised column | No crash; the finding's column name is blank |

### 1. Oversized files are rejected, not trimmed

The row-limit path crashed, and fixing the crash surfaced a worse problem underneath, so the behaviour changed rather than the crash simply being patched.

Previously the frame was cut to 10,000 rows and processing continued. The report's own `num_lines` and `error_ratio` are computed from `len(self.df)` *after* the slice, so a 12,000-row upload produced a report stating `num_lines: 10000` — indistinguishable from a complete review. A moderator approving on that basis ingests the whole file, including the rows nothing examined. The only record of the truncation was `T0016`, which has no row in the bundled `0000.error_codes.xlsx` and therefore rendered through the unknown-code path as *"Tool error: No entry for field 'error' and error code T0016"*.

It now raises. `process_list` already catches, marks the list `FAILED` and re-raises so Step Functions routes to `notify`, so the file is visibly rejected and the contributor can be asked to split it — the remedy the original report described.

The limit itself stays, now as the named constant `MAX_ROWS` with the reason recorded: `check_name_address_duplicates` compares every row pair, so cost grows with the square of the row count.

### 2. A missing `address` column no longer aborts the run

`check_name_address_duplicates` and `check_for_country_name_in_address` dereferenced `self.df.address` unguarded, so a file without that column crashed *after* the missing column had already been correctly recorded as `T0003`. Both now check the column is present.

Every finding inside the guarded block (`C0003`–`C0006`, `C0025`) reads the address column, so skipping it when the column is absent loses no signal. The duplicate check also guards `name`, which it dereferences the same way — the twin of the reported fault.

### 3. `T0008` names the column it is reporting

The call passed `column=` where the signature takes `column_name=`, so the record's name field stayed blank. Both are now passed: `column_name` populates the record, `column` feeds the `{{ column }}` message template. The original report suggested passing `column_name=` alone; that would have blanked the message instead, so both are needed.

Not a crash here because `T0008` is a table code, and the cell-reference path that raised the `openpyxl` `ValueError` in Airflow is only reached for column codes.

**Not in scope:** the Airflow implementation. It is retired at 2.30 and a fix there would not carry forward.

---

## Type of Change

---

## Testing

`89 passed` — 83 existing plus 6 new, via `python -m pytest lib/tests/test_contribot.py` under `src/contribot`. All six new tests fail against unmodified `main`.

- `TestOversizedFile` — a file of `MAX_ROWS + 1` is rejected with a message naming both counts and telling the reader to split it; a file of exactly `MAX_ROWS` still processes, pinning the boundary as inclusive.
- `TestMissingAddressColumn` — the duplicate check, the country-in-address check and a full `process()` all complete, and `T0003` is still recorded.
- `TestUnrecognisedColumnFinding` — the `T0008` record carries the column name *and* renders it in the message.

No new flake8 findings from these lines; the five under `src/contribot/lib/contribot.py` are pre-existing.

---

## Known TODO Items

* `T0016` is now unused in code. Its row can be retired from the error-codes workbook separately — the deployed copy comes from Drive, so it is not in this diff.
* Whether oversized files should be processable at all, rather than rejected, is a separate question. It would mean replacing the pairwise duplicate scan with blocking or bucketing.

---

### Implementation

### Testing & Validation

### Final Review

[OSDEV-3216]: https://opensupplyhub.atlassian.net/browse/OSDEV-3216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


valencia-max (approved):
LGTM!

PR #1266, opened 2026-09-01
[OSDEV-3242] Add candidate columns to Facility (model + migration)

tana-bee:
## Jira Ticket

[OSDEV-3242](https://opensupplyhub.atlassian.net/browse/OSDEV-3242)

---

## Summary of Changes

Adds five candidate-distinguishing columns to `Facility` for the Earth Genome satellite-detection integration: `is_candidate` (default `false`, indexed), `polygon` (nullable `geometry(Polygon,4326)` detection footprint, GiST-indexed), `confidence` (nullable float), and `external_id` + `source` (NULL/empty for normal facilities), with a unique constraint on `(source, external_id)` for ingest idempotency.

Purely additive — nothing reads the fields yet. This is the dependency root for the Phase 1 Earth Genome stories; the default-manager exclusion that hides candidates from existing surfaces ships separately in OSDEV-3380.

Because `api_facility` and `api_historicalfacility` hold millions of rows and take writes continuously, the migration is non-atomic and hand-written via `SeparateDatabaseAndState`: columns are added with `ADD COLUMN ... DEFAULT` (metadata-only on Postgres 11+), every index is built `CREATE INDEX CONCURRENTLY` with an invalid-index drop guard so an interrupted build is retryable, and the unique constraint is attached to the pre-built index (`ADD CONSTRAINT ... UNIQUE USING INDEX`, instant lock). State operations mirror the model exactly (`makemigrations --check` is clean).

The DB defaults on `is_candidate`/`source` are **kept** (unlike Django's `AddField`, which drops them): dedupe-hub writes `api_facility` through its own SQLAlchemy model that predates these columns, so its INSERTs must keep succeeding — as non-candidates — until OSDEV-3243 teaches it the new schema. Dropping the defaults broke the CI integration test (not-null violation on `is_candidate` from dedupe-hub's insert); same keep-the-default pattern as migration 0234.

---

## Type of Change

- [ ] UI / frontend change

---

## Testing

On a fresh seeded local environment (`./scripts/start_local_dev`):

- Migration applies cleanly (`Applying api.0235_add_facility_candidate_fields... OK`); `\d api_facility` confirms all five columns, the btree index on `is_candidate`, the GiST index on `polygon`, and the unique constraint.
- Reverse verified: `migrate api 0234` removes all columns/indexes/constraint; re-applying restores them.
- Constraint verified: inserting two rows with the same `(source, external_id)` pair is rejected with `duplicate key value violates unique constraint "api_facility_source_external_id_uniq"` (tested in a rolled-back transaction). Existing rows (`external_id` NULL) are unaffected — Postgres treats NULLs as distinct.
- Full test suite: 1265 tests, 1264 pass, 1 skipped, 1 failure — `test_successfully_geocoded_item_has_correct_results`, which calls the live Google geocoding API; Google now returns zero results for the test's hardcoded address. **Reproduced identically on unmodified `origin/main`** (same single failure running that test module at `a6f89337`), confirming it pre-dates this branch.
- API contract: every public facility serializer targets `FacilityIndex` (or claim/list models), none serialize `Facility` directly, and `api_facility_index` gains no columns — response shapes are unchanged by construction.
- ORM edge cases (in a rolled-back transaction): candidate saves with empty name/address and gets a normal OS ID; the `simple_history` row mirrors all five new fields; a duplicate `(source, external_id)` raises `IntegrityError` through the ORM; two candidates with NULL `external_id` coexist (Postgres NULLs-distinct semantics, matching Django's `UniqueConstraint`).

---

## Known TODO Items

- **Migration number collision:** open PR #1260 (OSDEV-3371) also contains a migration numbered 0235 — whichever PR merges second must renumber and re-point its dependency.
- No unit test for the constraint yet (verified at SQL level); a model-level test can ride along with OSDEV-3380, which adds the manager test suite.
- The flaky live-API geocoding test above predates this PR and needs its own fix/ticket.
- The five new fields appear in the Facility Django admin edit form (`FacilityHistoryAdmin` declares no `fields`/`exclude`); admin treatment of candidates is owned by OSDEV-3379.
- Follow-on tickets: OSDEV-3380 (default manager), OSDEV-3376/3377/3378/3379 (blast-radius audits).

---

### Implementation

### Testing & Validation

### Final Review

[OSDEV-3242]: https://opensupplyhub.atlassian.net/browse/OSDEV-3242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/django/api/migrations/0237_add_facility_candidate_fields.py
```diff
@@ -0,0 +1,418 @@
...
+        'CREATE INDEX CONCURRENTLY IF NOT EXISTS '
+        'api_historicalfacility_is_candidate_idx '
+        'ON api_historicalfacility (is_candidate)'
+    ),
+    'api_historicalfacility_polygon_gist': (
+        'CREATE INDEX CONCURRENTLY IF NOT EXISTS '
+        'api_historicalfacility_polygon_gist '
+        'ON api_historicalfacility USING gist (polygon)'
+    ),
+}
```
valencia-max, line 59: Do we really need an index for every new column? Every index we are building is using disk space in the DB, this will increase our DB size.

valencia-max: For example, on `api_facility` we already have these indexes with these sizes:

Image

valencia-max:
Before merging this PR to main, can we please deploy this to Development environment, insert some facility candidates and see how the platform works?

These new candidates will appear in the platform as "facilities" since we are syncing everything to opensearch. I would like to ensure the syncronization is correct and there's no error on the matching process.

src/django/api/migrations/0237_add_facility_candidate_fields.py
tana-bee: good point! especially since the candidate rows are going to only be a fraction of the database, i agree it makes more sense to go with a partial index.

valencia-max:
I was thinking another posibility.
Deploying this to an environment will set `is_candidate` to false by default, this means that will update the failicities. If this operation also updates the `updated_at`, basically will cause a full reindex of all facilities table.

Can you check this in `Development `during and after the deployment? You can check how much time the workflow take to finish and fetching the `updated_at` field in the `api_facility` table.

valencia-max (approved):

PR #1274, opened 2026-09-04
OSDEV-2278: [Claims Process Update] Allow claimants to view and edit pending claims

valencia-max:
## Jira Ticket

[OSDEV-2278](https://opensupplyhub.atlassian.net/browse/OSDEV-2278) — subtasks [OSDEV-3370](https://opensupplyhub.atlassian.net/browse/OSDEV-3370) (Platform API), [OSDEV-3371](https://opensupplyhub.atlassian.net/browse/OSDEV-3371) (UX + notification), and the platform side of [OSDEV-3374](https://opensupplyhub.atlassian.net/browse/OSDEV-3374) (automated-claims pipeline support)

---

## Summary of Changes

Lets claimants view and edit their own PENDING claims from My Account → My Facilities — replacing the email-based document exchange that raised the PII/SOC 2 concern behind OSDEV-2278.

- `GET/PATCH /api/facility-claims/{id}/pending/` — claimant view/edit of their own PENDING claim. Guard: owning contributor + PENDING; everything else 404s (no claim-id existence leak). `EditPendingClaimSerializer` derives from `FacilityCreateClaimSerializer` so create and edit share one set of validation rules; it additionally allows `null` on the clearable numeric/date fields (all `null=True` on the model), because in a partial PATCH omitted means "unchanged" — null is how a claimant actually clears a value or unchecks an emissions section. Write paths lock the claim row (`select_for_update`) so concurrent requests cannot race past validation.
- **Root cause fix:** `FacilityClaimAttachments.claim` was `on_delete=PROTECT` while facility deletion calls `claim.delete()` — any claim with attachments raised `ProtectedError` and aborted the whole deletion. Now CASCADE, with a `post_delete` signal removing the stored file.
- Terraform: signing role + assume-role wiring, and an `s3:signatureAge` ≤ 15 min deny **scoped to the `claim_attachments/` prefix** (caps a leaked URL's usable life at the policy layer). Other presigned consumers — facility-list downloads, partner-field-group icons embedded in CloudFront-cached responses — are deliberately untouched.

**UX + notification (OSDEV-3371).**

- My Facilities (`/claimed`) now requests `?statuses=PENDING,APPROVED` (endpoint default stays approved-only for other consumers), shows a status chip per row, and forks the row click: pending → new `/claimed/pending/{id}` edit view; approved → existing profile editor, unchanged.
- `PendingClaimEdit` reuses the v1 claim flow's step components (`ContactInfoStep`, `BusinessStep`, `ProfileStep`) on a single page — fields, Yup validation, and the OSDEV-2280 PII warning are the create flow's own components, not copies. Own Redux slice + Formik; the API↔form field mapping lives in one file (`PendingClaimEdit/utils.js`). Uploaded documents render in a manager section (download via the new endpoint, delete with confirmation). On Save, the field PATCH runs **before** the file upload, so a failed save can be retried without duplicating attachments in either direction (a retried no-op PATCH sends no email and no stamp).
- New generic `ClaimOutcomeDialog` (title/body/actions via props), also adopted by `ClaimForm` and `ClaimFacilityStepper` in place of their duplicated inline dialogs.
- The OSDEV-2280 warning copy is lifted into `CLAIM_PII_WARNING_TEXT` and shared by both claim-flow steps and the edit view (it was duplicated inline).
- `claim_updated_by_claimant` email to `NOTIFICATION_EMAIL_TO` on every claimant save, deferred with `transaction.on_commit` (no email for a rolled-back save; the SMTP round-trip doesn't hold the row lock): changed field names, document add/remove by filename, and a dashboard link — never a document or a URL to one.
- The message-claimant email templates gain a "How to update your claim" section pointing claimants at the platform edit flow (new `claimed_url` context), with replying kept as a fallback only.
- New `FacilityClaim.claimant_updated_at`, stamped **only** by claimant edits — moderator actions bump `updated_at` but not this — feeding a sortable "Updated by Claimant" column in the claims queue and its XLSX export, so claims edited after review starts don't hide in the "already reviewed" pile.

**Pipeline support (OSDEV-3374, platform side).** The automated-claims Lambda consumed exactly the raw presigned URLs this PR removes, so it moves to direct S3 reads via IAM (pipeline PR in the automated-claims repo):

- The staff-facing `FacilityClaimDetailsSerializer` exposes each attachment's opaque `storage_key` and the claim's `claimant_updated_at` (for the pipeline's edited-since-processed reprocessing). The claimant-facing pending payload exposes neither.
- New `claim_attachments_reader_role_arns` Terraform variable (default empty): listed role ARNs get a cross-account `s3:GetObject` bucket-policy grant scoped to `claim_attachments/*`; with the default the statement is not emitted.

`doc/release/RELEASE-NOTES.md`: all of the above added to the 2.30.0 entry.

---

## Type of Change

- [x] UI / frontend change

**Breaking change detail:** claim API payloads no longer carry `claim_attachment` (raw storage URL). The `automated-claims` Lambda consumed that field; its updated version (OSDEV-3374, automated-claims repo) handles all three payload generations — `storage_key` direct-S3, legacy presigned URLs, and the download endpoint — and should deploy before or together with this. The signature-age cap affects only `claim_attachments/*` objects; presigned URLs elsewhere in the bucket behave as before.

---

## Testing

- `api/tests/test_pending_claim_edit.py` — 31 tests, all passing:
  - Partial-update semantics; null clears the clearable fields (model + response) while null on required fields still 400s; URL validation; lifetime 20-file cap (19+2 rejected, 19+1 accepted); magic-byte rejection (HTML-as-pdf); extension allowlist; opaque storage keys (original filename never in the key).
  - Download endpoint: 302 for owner and superuser, 404 stranger, 401 anonymous; moderator claim-details payload carries `storage_key` + `claimant_updated_at` but no storage URLs; claimant payload carries no `storage_key`.
  - OSDEV-3371: PATCH stamps `claimant_updated_at` + sends the notification via captured `on_commit` callbacks (subject/body/dashboard link asserted, zero attachments); no-op saves send nothing and stamp nothing; attachment add stamps + notifies; `?statuses=` default/param/invalid-400 behavior.
- `api/tests/test_facility_claim_view_set.py` — new test asserting both message-claimant email bodies carry the platform-edit instructions, the `/claimed` link, and the reply fallback.
- `api/tests/test_facility_download.py` — attachment fixtures updated with real JPEG/PNG/PDF magic bytes (content validation made the old fake bodies fail claim creation).

---

## Known TODO Items

- **`automated-claims` (companion PR, OSDEV-3374):** direct S3 reads + reprocessing of edited claims; deploy order is platform first (this PR + the `claim_attachments_reader_role_arns` tfvar), pipeline second.
- Legacy attachments remain at the bucket root, outside the signature-age cap (their 60-second URLs are the only guard); after migrating them under `claim_attachments/`, tighten the signer role's `s3:GetObject` to the prefix (noted in `iam.tf`).

---

### Implementation

### Testing & Validation

### Final Review

[OSDEV-3374]: https://opensupplyhub.atlassian.net/browse/OSDEV-3374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployment/terraform/variables.tf
```diff
@@ -1565,3 +1565,18 @@ variable "contribot_google_drive_service_key_secret_name" {
...
+# attachment objects directly from the files bucket. Granted s3:GetObject
+# on the claim_attachments/ prefix only, via the bucket policy
+# (storage.tf). Set per environment in uncommitted tfvars; empty by
+# default so no cross-account access exists unless explicitly configured.
+# To get this value from the AWS console:
+# Lambda → auto-claims-moderate-claim → Configuration → Permissions → Execution role.
+variable "claim_attachments_reader_role_arns" {
+  type    = list(string)
+  default = [
+    "arn:aws:iam::343975343274:role/auto-claims-ModerateClaimRole-zZNYv0M0at1w"
```
daniellekain: +1 to coderabbit; please fix; comment above also says that this should be empty so maybe this was left in from testing?

doc/release/RELEASE-NOTES.md
```diff
@@ -16,15 +16,22 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
+* `0238_claim_attachment_cascade_and_upload_prefix.py` - Changes `FacilityClaimAttachments.claim` from `on_delete=PROTECT` to `CASCADE` (a claim with attachments previously raised `ProtectedError` and aborted the whole facility deletion) and moves the `claim_attachment` FileField's `upload_to` to the `claim_attachments/` prefix. See OSDEV-3370.
```
daniellekain: nit: 0238 is add_claimaint_updated_at and 0239 is claim_attachment_cascade_and_upload_prefix

src/django/api/helpers/claim_attachments.py
```diff
@@ -0,0 +1,119 @@
...
+    file.name = f'{uuid.uuid4().hex}{file_extension.lower()}'
+
+    return FacilityClaimAttachments.objects.create(
+        claim=facility_claim,
+        file_name=display_name[:200],
+        claim_attachment=file,
+    )
+
+
+def delete_claim_attachment(attachment):
```
daniellekain, line 111: Could we log the storage key in the post_delete failure handler? It currently reports attachment_id to Rollbar, but the row is already gone by the time that fires, so an orphaned S3 object has nothing pointing at it — recovering one later would mean diffing the whole bucket against the table. delete_claim_attachment() would be a natural home for that, which would also give the helper something to do beyond passing straight through to attachment.delete().

deployment/terraform/variables.tf
valencia-max: No, it should be the ARN code which will read objects to S3

daniellekain: ah okay should we update the comment on 1573/1574?

valencia-max: Done!

daniellekain (dismissed):

src/django/api/helpers/claim_attachments.py
valencia-max: Done!

doc/release/RELEASE-NOTES.md
valencia-max: Done!

daniellekain (approved):

PR #1280, opened 2026-09-09
[OSDEV-3438] Restore RBA promotions reverted by the OS Hub sync

daniellekain:
## Jira Ticket

[OSDEV-3438](https://opensupplyhub.atlassian.net/browse/OSDEV-3438)

Root cause was found in the [OSDEV-3391](https://opensupplyhub.atlassian.net/browse/OSDEV-3391) spike, which is now closed.

---

## Summary of Changes

Promotions made on the RBA private instance do not survive. The daily one-way sync from OS Hub (`sync_databases`) overwrites **every** synced field of a shared facility — including `created_from` — so a promotion is silently reverted whenever the public record next changes. There is no schedule to it: it happens on the next public edit, merge, claim change or admin action, which is why the decay went unnoticed. The overwhelming majority of the instance's OS IDs are shared with the public hub, so this affects nearly the whole dataset; only facilities minted on the instance itself are immune, because the sync never reads rows that do not exist publicly.

This adds `reassert_rba_promotions`, a management command that restores them after each sync. It is the tactical option (P0) from the [ownership-aware sync RFC](https://opensupplyhub.atlassian.net/wiki/spaces/SD/pages/1487994903) — it changes nothing about the sync itself, which is what makes it reversible.

Two design points worth reviewing:

**No journal, but promotion evidence is required.** The contribution survives the sync: its source, list item and match are rows created on the RBA instance, and the sync only upserts rows that also exist publicly. So the values are already durable in `api_facilitylistitem`, and the reverted set is derivable by query. Storing the values separately would only create a second copy that can disagree with the row `promote` actually reads.

The derivation must also require *evidence that a promotion happened* — a `PROMOTE_MATCH` processing result on the list item. Without it the query cannot distinguish a reverted promotion from the ordinary case of a contribution that dedupe-hub matched to a facility and nobody promoted, which is the majority of instance-origin matches; selecting those would rewrite the canonical fields of facilities nobody ever promoted, including RBA-minted ones the sync never touches. The marker is written by the promote endpoint onto an instance-local row the sync does not upsert, so it outlives the overwrite — and it is the same predicate `api/services/production_locations_lookup.py` and `reindex_promoted_locations` already use.

**Newest-first, then exclude.** The query picks the newest *promoted* RBA match per facility *before* filtering out already-promoted ones. Doing it the other way round would select an older contribution whenever the newest was already in place, so every run would flip the facility backwards. Restricting to promoted matches also means the command cannot overrule a moderator who deliberately promoted an earlier contribution.

Restoration mirrors the `promote` endpoint (canonical fields, `created_from`, the `PROMOTE_MATCH` processing result). The change reason keeps the `Promoted … over …` wording that `facility_history.py` matches on to render a promotion, with a `(re-asserted after sync)` suffix so the audit trail distinguishes this from a moderator action.

---

## Type of Change

---

## Testing

18 unit tests (`api.tests.test_reassert_rba_promotions`), all passing, covering:

- a reverted promotion is restored (canonical fields and `created_from`)
- the `PROMOTE_MATCH` processing result records the previous `created_from`
- the change reason still contains `Promoted`, so facility history renders it
- a second run is a no-op — nothing is selected once the promotion is in place
- the newest RBA contribution wins, and an older one is never re-applied
- `os_hub`-origin matches, inactive matches, unconfirmed matches and non-matched items are all ignored
- `--dry-run` reports without writing
- `--limit` caps the run
- the RBA list item and match survive a sync overwrite intact — status, `is_active` and `origin_source` are untouched, which is the assumption the whole no-journal design rests on
- a promotion reverted by a sync overwrite is detected and restored from the surviving list item
- three consecutive overwrite/re-assert cycles leave the facility correct, with exactly one processing result recorded per cycle

`flake8` clean on all three files. `makemigrations --check` drift on this branch is pre-existing and unrelated (claims, partnerfield, tigerline models).

Not yet exercised against real RBA data — see below.

---

## Known TODO Items

- **Nothing schedules this yet.** Deliberate: it is run manually until a few runs have been watched. Wiring it to the Batch job-completion event, plus a divergence report so a silent failure does not recreate the decay, is a follow-up.
- ~~**Provenance verification outstanding.**~~ **Resolved.** The query relies on `origin_source`, and the concern was that a non-ORM write outside a request could leave it null. Counted against the Rba database on 2026-09-09: **zero** null `origin_source` values on `api_facilitymatch`, every row stamped either `os_hub` or `rba`. No contributor-based fallback is needed.
- **Items with no geocoded point are skipped.** `FacilityListItem.geocoded_point` is nullable while `Facility.location` is not, so selecting such a row raised `IntegrityError` on save, was swallowed by the generic handler, and would have failed identically on every later run while making each run exit non-zero.
- **The command's true target set on the RBA instance is currently empty.** Promotions there are rare and the existing state is already correct, so the query selects nothing on that instance today. This change is preventative: it protects the promotions the address-update automation will create, and it needs to be in place before that automation runs — otherwise those promotions start being reverted and there is real damage to repair. A consequence for review: the command cannot be exercised against real reverted data yet. The unit tests cover the mechanism; the first genuine run follows the address batch and should be watched rather than assumed correct.
- **Does not address merges or deletions** (W4 in the RFC): the sync *resurrects* merged-away records rather than merely reverting them, so those cannot be re-asserted the way a promotion can. [OSDEV-3435](https://opensupplyhub.atlassian.net/browse/OSDEV-3435) forbids them instead.
- The durable fix (field-level ownership or a rebase model) is the RFC's open decision and would need a PCR.

---

### Implementation

### Testing & Validation

- [x] For architectural changes: I have documented the design decision and notified affected teams.

- [x] I have updated relevant documentation (release notes; design recorded in the RFC).

### Final Review

[OSDEV-3391]: https://opensupplyhub.atlassian.net/browse/OSDEV-3391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OSDEV-3435]: https://opensupplyhub.atlassian.net/browse/OSDEV-3435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OSDEV-3438]: https://opensupplyhub.atlassian.net/browse/OSDEV-3438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


valencia-max (dismissed):
LGTM!! Amazing work 🚀

daniellekain:
@valencia-max need one more approval after addressing code rabbit

valencia-max (approved):

PR #1281, opened 2026-09-09
[OSDEV-3435] Refuse merges of publicly-synced records on the RBA instance

daniellekain:
## Jira Ticket

[OSDEV-3435](https://opensupplyhub.atlassian.net/browse/OSDEV-3435)

---

## Summary of Changes

Merges performed on the RBA private instance against a record that also exists on public OS Hub are silently only half-applied. The one-way nightly sync upserts by `uuid` and never deletes, so the merged-away facility is recreated on the next run because it still exists publicly — while the aliases and matches the merge re-pointed stay with the target. The sync cannot repair that state, so this refuses the merge rather than allowing it to be produced.

The guard sits in the merge endpoint before any writes, returns 400, and names the OS IDs that cannot be merged so the admin knows which record is the problem.

Three things worth reviewing:

**There is only one merge-performing path.** The ticket specifies enforcement on two paths, but that is wrong and I have corrected it here. `merge()` in `moderation_events_view_set.py` is a **GET returning merge history** and performs no merge; grepping `MERGE_FACILITY` / `FacilityMatch.MERGED` writes confirms `POST /api/facilities/merge/` in `facilities_view_set.py` is the only path that merges. The dashboard calls that same endpoint (`src/react/src/util/util.js`), so one server-side check covers both the API and the UI, satisfying the ticket's "not in the UI alone" requirement. The ticket has been updated.

**Scoped by `INSTANCE_SOURCE` rather than a new flag.** The ticket left activation TBD between an RBA tfvars flag and deriving from instance identity. `INSTANCE_SOURCE` already exists (`oar/settings.py`, default `os_hub`) and is plumbed through every ECS task and Batch job definition, so deriving from it needs no new configuration to keep in sync and no deploy-time coordination. It is also the same check `reassert_rba_promotions` uses on the OSDEV-3438 branch.

**The check fails closed.** `Facility.origin_source` is nullable, and the database trigger's fallback for a write arriving without the session setting is `os_hub`. Anything not explicitly stamped `rba` is therefore treated as publicly synced and refused. Rejecting an unstamped local record is the safe error; permitting a merge that the sync then half-undoes is not.

---

## Type of Change

---

## Testing

7 new tests in `api.tests.test_rba_merge_guard`, plus the 10 existing `api.tests.test_facility_merge` tests — **17 passing**, run together to confirm no regression on the shared endpoint:

- two records created on the instance merge exactly as before
- the merged record being publicly synced is refused, and the response names it
- the target record being publicly synced is refused, and the response names it
- both records unmergeable — the response names both
- an unstamped record (`origin_source` null) is refused
- a refused merge writes nothing: facility count unchanged, and the merge facility's match keeps its facility and `AUTOMATIC` status
- public OS Hub is unaffected — with `INSTANCE_SOURCE=os_hub` the merge succeeds even for records this instance did not create

`flake8` clean on all three changed/added Python files.

Not exercised against the RBA instance itself; the guard is inert on every other environment, so there is nothing to verify outside the tests until an Rba deploy.

---

## Known TODO Items

- **`is_rba_instance` will exist twice until OSDEV-3438 merges.** `api/reassert_rba_promotions.py` on that branch defines the same three-line check. Deliberately not refactored across branches to avoid a conflict — once both land, that module should import from `api/helpers/rba_instance.py` and the duplicate should go.
- **Dashboard error surfacing is server-message-only.** The admin sees the 400 message returned by the endpoint. A pre-flight disable in the merge UI is the optional polish the ticket describes and is not included.
- **Does not address deletions.** The same resurrection behaviour applies to privately-deleted rows; this PR covers merges only.
- The guard is per-instance by `INSTANCE_SOURCE`, so a future private instance would inherit it automatically once its `origin_source` value is added to `OriginSource.CHOICES`. No such instance exists today.

---

### Implementation

### Testing & Validation

- [ ] For architectural changes: I have documented the design decision and notified affected teams. — design decision recorded on the ticket and above; affected teams **not** yet notified, since the guard changes what an RBA admin can do on their own instance.

- [x] I have updated relevant documentation (release notes; module docstring in `api/helpers/rba_instance.py` documents the rule).

### Final Review

[OSDEV-3435]: https://opensupplyhub.atlassian.net/browse/OSDEV-3435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


valencia-max (dismissed):
LGTM! Just address CodeRabbit comments!

valencia-max (dismissed):

valencia-max (approved):

PR #1283, opened 2026-09-09
[OSDEV-3424] Return the created item and match ids from approval endpoints

daniellekain:
## Jira Ticket

[OSDEV-3424](https://opensupplyhub.atlassian.net/browse/OSDEV-3424)

---

## Summary of Changes

A client that needs to act on the contribution an approval just created — promoting it to the production location's primary values, for example — had no way to identify it. The response carried only `os_id`, so the client had to call `GET /api/facilities/{os_id}/split/` afterwards and work out which match was the new one by taking its own most recent.

That is an extra call to reconstruct a fact the server already held, and it is only reliable while nothing else is writing to that location. That assumption was tolerable when we ran such batches ourselves; it is not one we can make for a partner running [OSDEV-3440](https://opensupplyhub.atlassian.net/browse/OSDEV-3440) unsupervised, which is what raised this from a nice-to-have.

Both approval flows now return `item_id` and `match_id` alongside `os_id`:

- `POST /api/v1/moderation-events/{id}/production-locations/` (create)
- `PATCH /api/v1/moderation-events/{id}/production-locations/{os_id}/` (update)

**The change is additive.** `os_id` is unchanged and existing consumers are unaffected.

Three implementation notes:

**The template discarded the match it created.** `EventApprovalTemplate.__create_facility_match` returned `None`; it now returns the record, surfaced through a `created_facility_match` property. The `FacilityMatchTemp` path shares the same private creator, so that now returns its record too — unused, but the alternative was two near-identical creators.

**The response body moved into `ModerationEventsService.build_approval_response`** so the create and update flows cannot drift apart. They previously constructed the same literal in two places, which is exactly how one of them ends up missing a field later.

**`match_id` is tolerated as null rather than raising.** It should never be null on a successful approval, but a response-shape problem must not be able to turn a completed approval — which has already written the source, item and match — into a 500 the caller reads as failure.

---

## Type of Change

---

## Testing

**49 tests passing** across `api.tests.test_moderation_events_add_production_location` and `api.tests.test_moderation_events_update_production_location`. `flake8` clean on all four changed files.

Rather than add isolated tests, I extended the shared `assert_success_response` helper in `base_moderation_events_production_location_test.py` to assert that both ids are present and non-null, that `item_id` resolves to a `FacilityListItem` whose `facility_id` matches the returned `os_id`, and that `match_id` resolves to a `FacilityMatch` pointing at that same item and location. Both approval flows call that helper, so create and update are both covered and any future flow that uses it inherits the check.

Verified by construction that this is a real assertion rather than a tautology: the helper fails if either id is missing, null, or inconsistent with the others.

---

## Known TODO Items

- **API documentation is not updated here.** The v1 docs live in a separate repository and there is no inline `swagger_auto_schema` on these two actions to update, so there is nothing to change in this diff. The response-schema update belongs with the other pending api-docs work.
- **`FacilityMatchTemp`'s returned record is unused.** It comes along because both match types share one private creator. Splitting them to avoid an unused return value seemed worse than the unused return value.
- No consumer in this repository reads the new fields yet. [OSDEV-3440](https://opensupplyhub.atlassian.net/browse/OSDEV-3440) reads `match_id` opportunistically and needs no change once this deploys.

---

### Implementation

- [x] I have considered backward compatibility and migrations where applicable — the response change is additive and there are no schema changes.

### Testing & Validation

- [ ] I have updated relevant documentation (README, ADRs, inline comments, API docs). — release notes and inline docstrings updated; the **v1 API docs are not**, see Known TODO.

### Final Review

[OSDEV-3424]: https://opensupplyhub.atlassian.net/browse/OSDEV-3424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OSDEV-3440]: https://opensupplyhub.atlassian.net/browse/OSDEV-3440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OSDEV-3440]: https://opensupplyhub.atlassian.net/browse/OSDEV-3440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


valencia-max (dismissed):
LGTM!

daniellekain:
@valencia-max re-requesting after rebase

valencia-max (dismissed):

valencia-max (dismissed):

valencia-max (approved):

